PR #1, opened 2020-05-19
[13.0][MIG] dms

etobella:
Creating DMS from MuK original module.

Still WIP but comments are welcome

@pedrobaeza @geluboros @AntoniRomera @dalonsod @jarroyomorales


pedrobaeza:
cc @keshrath

dms/models/directory.py
```diff
@@ -0,0 +1,549 @@
...
+        not_starred_records = self.env["dms.directory"].sudo()
+        for record in self:
+            if not record.starred and self.env.user in record.user_star_ids:
+                starred_records |= record
+            elif record.starred and self.env.user not in record.user_star_ids:
+                not_starred_records |= record
+        not_starred_records.write({"user_star_ids": [(4, self.env.uid)]})
+        starred_records.write({"user_star_ids": [(3, self.env.uid)]})
+
+    @api.returns("self", lambda value: value.id)
```
emagdalenaC2i: Delete @api.returns

dms/models/dms_file.py
```diff
@@ -0,0 +1,621 @@
...
+        for record in self:
+            values = self._get_content_inital_vals()
+            binary = base64.b64decode(record.content or "")
+            values = record._update_content_vals(values, binary)
+            updates[tools.frozendict(values)].add(record.id)
+        with self.env.norecompute():
+            for vals, ids in updates.items():
+                self.browse(ids).write(dict(vals))
+
+    @api.returns("self", lambda value: value.id)
```
emagdalenaC2i: delete @api.returns

dms/readme/USAGE.rst
```diff
@@ -0,0 +1,2 @@
+The best way to manage the documents is to switch to the Documents view. Existing
+documents can be managed there and new documents can be created.
```
emagdalenaC2i:
```suggestion
The best way to manage the documents is to switch to the Documents view.
Existing documents can be managed there and new documents can be created.
```

etobella:
@emagdalenaC2i Comments attended :wink:

dms/models/directory.py
pedrobaeza: Although in migration guide, it puts that it should be removed, I'm not sure now about the removal as Odoo code still contains this decorator.

dalonsod:
Tested locally, got this error when creating an storage (I don't know if error was already present in MuK implementation):

dms_err_1

etobella:
@dalonsod In runbot works as expected... This is weird.

dalonsod:
OK, I'll debug it

dalonsod:
> @dalonsod In runbot works as expected... This is weird.

Forget it, it was surely an old Odoo v13 bug, updated it and working. Sorry :(

I'll continue testing the module

dms/views/storage.xml
```diff
@@ -0,0 +1,246 @@
...
+        <field name="view_mode">kanban,tree,form</field>
+        <field name="domain">
+            [
+            ("storage_id", "=", active_id),
+            ("is_hidden", "=", False),
+            ]
+        </field>
+        <field name="context">
+            {
+            'default_root_storage': active_id,
```
dalonsod: typo in `context`

```suggestion
            'default_root_storage_id': active_id,
```

dms/views/storage.xml
```diff
@@ -0,0 +1,246 @@
...
+                        <group>
+                            <field name="company_id" />
+                        </group>
+                    </group>
+                    <notebook>
+                        <page name="page_roots" string="Root Directories">
+                            <field
+                                name="root_directory_ids"
+                                groups="dms.group_dms_manager"
+                                context="{'default_root_storage': active_id, 'default_is_root_directory': True}"
```
dalonsod: Typo in `context`

```suggestion
                                context="{'default_root_storage_id': active_id, 'default_is_root_directory': True}"
```

dms/views/directory.xml
```diff
@@ -0,0 +1,697 @@
...
+        <field name="name">dms_directory.form</field>
+        <field name="model">dms.directory</field>
+        <field name="inherit_id" ref="view_dms_directory_form" />
+        <field name="groups_id" eval="[(6, 0, [ref('dms.group_dms_manager')])]" />
+        <field name="arch" type="xml">
+            <field name="is_root_directory" position="attributes">
+                <attribute name="invisible">0</attribute>
+            </field>
+            <field name="root_storage_id" position="attributes">
+                <attribute name="readonly">0</attribute>
```
dalonsod, line 606: Maybe `default_root_storage_id` context value can be used here in order to prevent creating a directory with a different storage from a certain storage form

dalonsod (changes requested):
Some typos in context variables

dms/views/directory.xml
etobella: Added

etobella:
Fixed the issues with context, and proposed a solution simplifying the view :wink:

rousseldenis:
Web drop is not working as missing directory_id.

image

primes2h:
Tested on runbot:

Documents→Files→Create

immagine

rousseldenis:
One question. Was the ir.attachment too tough to be used as base for files?

pedrobaeza:
For me it's clear: they are different things. Things that you want to put to files (like versions, permissions, etc), don't have to charge the `ir.attachment` model. Even more, one file can have several attachments (due to versions as said).

keshrath:
We had also considered using the attachment model to store the data, but finally decided against it because an own model offers more freedom. But there are also valid arguments in favour of the attachment model.

etobella:
@rousseldenis @primes2h Comments have been attended. I hope they fill your expectatives

etobella:
I think there is too much rules in attachment in order to use them.

rousseldenis:
> For me it's clear: they are different things. Things that you want to put to files (like versions, permissions, etc), don't have to charge the `ir.attachment` model. Even more, one file can have several attachments (due to versions as said).

Yes, I just wanted to know if they were thinking about it (maybe, to avoid a new model), but the answer is done. I've a little bit tested the interface and it is very promising !

One more question: isn't it possible to split a little bit into more modules as it is actually huge? For example the base dms application with base security, base menus, then a module for files/category, ...

pedrobaeza:
You want all the things together, not to need to install a lot of separated modules for the same feature. I don't consider this one as very huge personally...

rousseldenis:
> You want all the things together, not to need to install a lot of separated modules for the same feature. I don't consider this one as very huge personally...

It has pros and cons but it was a minor question. So, let's go forward !

etobella:
Originally they were splitted, but we discussed a bit about this and our decision was that it had no sense too much modules. Maybe new modules could be necessary if we want to do some kind of dms or something like that.

LoisRForgeFlow (commented):
This looks great! Thanks to MuK for sharing and to contributors here for migrating it.

I noticed that the preview does not work neither using web_responsive or web_enterprise. The following error appear for any file:

image

The same file is previewed correctly if attached and opened in any regular record chatter.

etobella:
Now a preview of the file is available. Just wondering what to do 

The error is found because odoo is giving an error with svg file. I will try to find what happens there.

etobella:
The problem is related to this:

https://github.com/odoo/odoo/blob/13.0/odoo/http.py#L1627

On odoo 13, svg is no longer set as safe image val. On 12, there was no constrain like this.

Adding `image/svg+xml` to the list solved the issue, but I don't know if Odoo will accept this change. @pedrobaeza What do you think?

etobella:
I have created a PR on odoo: odoo/odoo#51832 :crossed_fingers: 
The change of the behavior was implemented on odoo/odoo#31811

etobella:
The problem was related with the name of the field mimetype.Now it is fixed and should work as expected!!

etobella:
I think now it is already clean and functional. Any comments are welcome, a Preview has been added in order to simplify usage.

LoisRForgeFlow (approved):
All my wishes have been fulfilled! Thanks @etobella 

LGTM :+1:

Rad0van:
LGTM
One thing: on first install it screwed up the web because of missing lessc / node-less. Don't know if there is a way to prevent this...

etobella:
@Rad0van I could not reproduce your issues. It worked for me without any issues. How can I reproduce your issue?

Rad0van:
> @Rad0van I could not reproduce your issues. It worked for me without any issues. How can I reproduce your issue?

I run local instance of Odoo 13.0 on Kubuntu 19.10. When I installed dms module it briwfly showed some warning but in console this was logged:

`WARNING odoo13com odoo.addons.base.models.assetsbundle: Could not execute command 'lessc'
This error occured while compiling the bundle 'web.assets_backend' containing`
After the installation the CSS was broken so the GUI was showing unstyled things.

I had to install node-less package (which contains lessc command) and reinstall module to get this fixed.

pedrobaeza:
@Rad0van there's no less files in this module, so it's another thing of your deployment.

Rad0van:
> @Rad0van there's no less files in this module, so it's another thing of your deployment.

Could it be a thing of one of the dependencies? I can see the same error popping up in Google searches with Odoo. I was using the instance for many weeks and this happened every time I tried to install / upgrade dms module until I installed node-less package...

Rad0van:
> @Rad0van there's no less files in this module, so it's another thing of your deployment.

There is .less file in web_drop_target module that is dependency for dms. Could that be the culprit?

pedrobaeza:
Indeed, there's https://github.com/OCA/web/blob/13.0/web_drop_target/static/src/less/web_drop_target.less in the dependency `web_drop_target`. 

@etobella did you migrate that module? `less` files should be converted to `scss` for avoiding that extra burden, and I see the file is very simple.

etobella:
It was migrated by @NuriaMForgeFlow but I will make the fix :wink:

dms/README.rst
```diff
@@ -0,0 +1,110 @@
+===
+DMS
```
pedrobaeza: Document Management System

dms/README.rst
```diff
@@ -0,0 +1,110 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/server-tools-13-0/server-tools-13-0-dms
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/149/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+DMS is a module to create, manage and view files directly within Odoo.
```
pedrobaeza:
```suggestion
DMS is a module for creating, managing and viewing document files directly within Odoo.
```

dms/README.rst
```diff
@@ -0,0 +1,110 @@
...
+In addition to views for files and folders, it offers its own all in one view
```
pedrobaeza: `folders` or `directories`?
```suggestion
In addition to views for files and folders, it offers on its own an all in one view
```
I don't fully get any way the sense of this sentence

dms/README.rst
```diff
@@ -0,0 +1,110 @@
...
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to *Documents -> Configuration -> Storages*.
+#. Create a new document setting.
```
pedrobaeza: Explain the differences between both systems. I think anyway that I will comment on this later on code review as for me the ideal storage is through binary field with `attachment=True`.

dms/README.rst
```diff
@@ -0,0 +1,110 @@
...
+#. Afterwards go to *Documents -> Directories*.
+#. Create a new directory, mark it as root and select the previously created setting.
+
+Usage
+=====
+
+The best way to manage the documents is to switch to the Documents view. Existing
```
pedrobaeza: What do you mean by `Documents view`? Better to indicate the menu path. And also you should explain all the possible features, like lock, navigation, etc.

dms/README.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to *Documents -> Configuration -> Storages*.
+#. Create a new document setting.
+#. Afterwards go to *Documents -> Directories*.
+#. Create a new directory, mark it as root and select the previously created setting.
```
pedrobaeza, line 48: Explain also access management.

dms/README.rst
```diff
@@ -0,0 +1,110 @@
...
+
+* Mathias Markl <mathias.markl@mukit.at>
+* Enric Tobella <etobella@creublanca.es>
+* Antoni Romera
+* Gelu Boro
+
+Other credits
+~~~~~~~~~~~~~
+
+This module was originaly created by MuK
```
pedrobaeza: I don't think this is needed, as it's already put as author.

pedrobaeza (changes requested):
Thanks for the excellent work.

Let me put first a functional review and later I will perform a code review. These are the weird things I have found:

- When browsing the document hierarchy, full path is shown:
  Selección_015
  I suppose it's due to the name_get/display_name automatically fetched, but it's not practical and when you have more than 2 levels (or long names), it gets difficult to see the final part. Can you please try to put a context key there and modify name_get/display_name for showing only the `name`? Maybe that tree filter has any mechanism for indicating another field to be used as display.
- The tooltip for both buttons on directory kanban item put a number that is not useful nor confusing (maybe its ID):
  Selección_016
- If you click on the file button of the directory kanban item for showing files, it shows only the files for that directory, but the whole directory tree is shown:
  Selección_017
  Is there a way to show only the hierarchy from the selected directory? Or maybe as alternative to put the directory filter in the search bar, but allows to remove it.
- What this menu means?
  Selección_018

dms/README.rst
etobella: I think that all the sentence has no sense. I will remove it.

dms/README.rst
etobella: This comes from the original documentation from MuK with the current system, it has no sense. Removing it

etobella:
Thanks @pedrobaeza All comments attended. About your questions:

The number is the number of files or subdirectories of a folder. I have changed the title to be easier to understand (5 Files, 2 Subdirectories)

The migration is a tool to migrate storages from one Save Type to another. I have tried to explain it on the readme.

The browsing path should be solved now. Thanks for your comments

pedrobaeza:
In a quick test, my third point is still happening (about navigating to files from the directory).

etobella:
Sorry @pedrobaeza I passed the point 3 :disappointed:.

About your comment, the widget is a default from Odoo 13.0 and it is not recalculated on every filter. So I don't know how to modify it to delete the selection.

etobella:
@pedrobaeza The problem is that trees cannot be refreshed on webpanel. It is strange, but it is refreshed only once as you can check on
https://github.com/odoo/odoo/blob/5b9e5564d377bad88bef6c33ab5d0262fe561f89/addons/web/static/src/js/views/search_panel.js#L393

However, I think it is useful to see all the tree of folders

keshrath:
You could maybe adjust the SearchPanel to match the selected folder.
Current:
ezgif com-video-to-gif (45)
Expected:
image
Or alternatively like in version 12, where we reduce the tree to match the selected folder, but I agree with you that it might be better to see the whole tree.:
ezgif com-video-to-gif (46)

etobella:
Thanks @keshrath comments attended. I think now the tree works as expected :wink:

primes2h:
Files→Create
I suggest to hide **/False** until path has been given a value.

immagine

etobella:
Thanks @primes2h hidden if directory_id or name are not defined.

primes2h:
> Thanks @primes2h hidden if directory_id or name are not defined.

Great, thank you! :+1:

primes2h:
@etobella 
Configuration→ System → Migration
I noticed that clicking on Files→ "Create" does nothing. Any hints?

etobella:
Thanks!!! fixed :wink: 
Create button shouldn't be there, I have added `create="0" delete="0"` on the migration tree view and the issue is solved

agent-z28:
One general question, will it be a good idea to migrate this backwards to v11, or do you suggest to use the muk version, since we want to have a foward compatibility ?

etobella:
About the migration backward, I don't have any issue, but I think that @keshrath @Muk-it should say something about this.

Remember that some functionalities we have used will not be available on 11 and 12, as they are new. from 13

etobella:
Also, there is no forward compatibility or migration scripts

yajo:
> some functionalities we have used will not be available on 11 and 12, as they are new

Which ones concretely?

etobella:
The left bar with the selection of folders
imagen

agent-z28:
@etobella so if we provide an alternative, or backport for this bar, 
there are no other difficulties?

keshrath:
> @etobella so if we provide an alternative, or backport for this bar,
> there are no other difficulties?

https://github.com/muk-it/muk_web/tree/12.0/muk_web_searchpanel

etobella:
There shouldn't be any difficulties but it needs to be migrated too :wink: 

Also, we need the confirmation from MuK that there is no issue in porting back the modules

etobella:
Besides the porting to 12.0, Can we proceed with this module? @pedrobaeza

rousseldenis (approved):

etobella:
@pedrobaeza @dalonsod Can you review again? :wink:

dms/models/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+from . import dms_security_mixin
+
+from . import storage
+from . import directory
+from . import dms_file
+
+from . import category
+from . import tag
+
+# from . import ir_http
```
pedrobaeza: Remove this

dms/models/access_groups.py
```diff
@@ -0,0 +1,130 @@
+# Copyright 2017-2019 MuK IT GmbH
+# Copyright 2020 RGB Consulting
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models
+
+
+class DmsAccessGroups(models.Model):
+    _name = "dms.access_groups"
```
pedrobaeza: Avoid plurals in model names and optionally also `_`:
```suggestion
    _name = "dms.access.group"
```

dms/models/access_groups.py
```diff
@@ -0,0 +1,130 @@
...
+        compute="_compute_count_directories", string="Count Directories"
+    )
+
+    @api.depends("directory_ids")
+    def _compute_count_directories(self):
+        for record in self:
+            record.count_directories = len(record.directory_ids)
+
+    @api.model
+    def _add_magic_fields(self):
```
pedrobaeza, line 40: Why putting these fields this way? If needed, a good docstring should be added

dms/models/access_groups.py
```diff
@@ -0,0 +1,130 @@
...
+        super(DmsAccessGroups, self)._add_magic_fields()
+
+        def add(name, field):
+            if name not in self._fields:
+                self._add_field(name, field)
+
+        add(
+            "parent_group",
```
pedrobaeza: I don't want to force you to change too much, but all these fields don't follow the conventions: `parent_group_id`, `group_ids`, etc. Change it now will avoid later renamings.

dms/models/mixins_thumbnail.py
```diff
@@ -0,0 +1,102 @@
...
+from odoo import api, fields, models
+from odoo.modules.module import get_resource_path
+
+
+class Thumbnail(models.AbstractModel):
+
+    _name = "dms.mixins.thumbnail"
+    _description = "Thumbnail Mixin"
+
+    custom_thumbnail = fields.Image(
```
pedrobaeza, line 17: Does it make sense to apply here new v13 `image.mixin` model?

dms/models/res_config_settings.py
```diff
@@ -0,0 +1,44 @@
...
+class ResConfigSettings(models.TransientModel):
+
+    _inherit = "res.config.settings"
+
+    documents_binary_max_size = fields.Integer(
+        string="Size", help="Defines the maximum upload size in MB. Default (25MB)"
+    )
+
+    documents_forbidden_extensions = fields.Char(
+        string="Extensions",
```
pedrobaeza, line 19: I think we should provide here a sane default.

dms/static/description/index.html
```diff
@@ -0,0 +1,176 @@
+<section class="oe_container">
```
pedrobaeza: This HTML file should be removed and regenerate it from the README fragments. Banner and other images should go out as well: marketing and old screenshots.

dms/static/src/img/thumbnails/file_ai.svg
```diff
@@ -0,0 +1 @@
... 141.697 109.143 L 151.571 109.143 L 151.571 115.429 L 170.429 115.429 L 170.429 96.571 L 151.571 96.571 L 151.571 102.857 L 142.36 102.857 C 143.551 93.369 150.041 85.527 158.772 82.372 C 160.632 87.001 165.142 90.286 170.429 90.286 C 177.362 90.286 183 84.647 183 77.714 L 183 74.571 L 157.857 74.571 Z  M 101.286 109.143 L 95 109.143 L 95 102.857 L 101.286 102.857 L 101.286 109.143 Z  M 91.857 140.571 C 88.391 140.571 85.571 137.752 85.571 134.286 C 85.571 130.819 88.391 128 91.857 128 C 95.324 128 98.143 130.819 98.143 134.286 C 98.143 137.752 95.324 140.571 91.857 140.571 Z  M 157.857 102.857 L 164.143 102.857 L 164.143 109.143 L 157.857 109.143 L 157.857 102.857 Z  M 129.571 112.286 C 126.105 112.286 123.286 109.467 123.286 106 C 123.286 102.533 126.105 99.714 129.571 99.714 C 133.038 99.714 135.857 102.533 135.857 106 C 135.857 109.467 133.038 112.286 129.571 112.286 Z  M 170.429 84 C 168.156 84 166.176 82.777 165.073 80.97 C 165.802 80.904 166.538 80.857 167.286 80.857 L 175.869 80.857 C 174.785 82.733 172.751 84 170.429 84 Z " fill="rgb(200,189,184)"/><path d=" M 157.857 40.475 L 157.857 77.714 L 195.097 77.714 L 157.857 40.475 Z " fill="rgb(217,215,202)"/></g></g></g></svg>
```
pedrobaeza: Credits for these files should be placed on the README.

dms/tools/file.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+import re
+import shutil
+import tempfile
+import unicodedata
+
+from odoo.tools.mimetypes import guess_mimetype
+
+
+def slugify(value, lower=True):
```
pedrobaeza: This tool already exists in Odoo. It should use that one IMO.

dms/models/storage.py
```diff
@@ -0,0 +1,113 @@
...
+
+    count_storage_files = fields.Integer(
+        compute="_compute_count_storage_files", string="Count Files"
+    )
+
+    # ----------------------------------------------------------
+    # Actions
+    # ----------------------------------------------------------
+
+    def action_storage_migrate(self):
```
pedrobaeza, line 87: Isn't make sense to auto-execute this when `save_type` is changed and remove the manual button?

pedrobaeza (changes requested):
Some code comments (it's barely impossible to review this at once, but time will serve for refining), but at least some things that catch my eye.

Any way, it's a very good work from both MuK and porters.

keshrath:
> There shouldn't be any difficulties but it needs to be migrated too 😉
> 
> Also, we need the confirmation from MuK that there is no issue in porting back the modules

Instead of a backport I would prefer migration scripts to v13. This would also make upgrading easier.

pedrobaeza:
@keshrath indeed that can be done, but understand that OCA contributors feel more comfortable having all the possible modules hosted in OCA, and that's why they insist in backport this to v12. Let's make one thing: the contributors that have asked for the backport do it, and I compromise to make migration scripts from both OCA and MuK. Deal?

etobella:
Also, I was thinking of how to make the migration script, So don't worry, we will work on it somehow :wink:

dms/models/__init__.py
etobella: Sure!

dms/models/access_groups.py
etobella: Yes, it implies some changes, but it has sense

dms/models/access_groups.py
etobella: This is done this way because relation is computed. I am not sure how to solve this directly

dms/static/src/img/thumbnails/file_ai.svg
etobella: This images are not generated on by MuK or us, it is just a repository image. In order to maintain all, I am moving it to `static/lib/img`

dms/tools/file.py
etobella: Removed

dms/models/access_groups.py
etobella: Changed

dms/models/mixins_thumbnail.py
etobella: Because there is two images, the original one and the computed

dms/models/storage.py
etobella: Yes and No, I mean, if you have a small set of elements, it could be done this way, otherwise, you might need to do it on small batches. :thinking:

dms/static/description/index.html
etobella: Done

etobella:
There was a small bug when clicking on the path. Now it is fixed :wink:

etobella:
@keshrath @pedrobaeza I have made a first approach of a migration script for `muk_dms`, it was necessary to add `muk_dms` on the `apriory.py` file from OpenUpgrade

pedrobaeza:
@etobella please leave out of this initial PR the migration scripts, as it will delay the merge of the working 13.0 module. You can do a PR immediately and I'll check it. Right now I'm about merging with final review!

etobella:
Perfect!! I have created a branch for the migration script. Once this is merged, we can publish the PR in order to review it :wink:

etobella:
If anyone wants to continue with the migration he can take the branch: 

https://github.com/tegin/dms/tree/13.0-dms-migration

etobella:
Also, a backport to 12.0 of the module can be found on our repo:

https://github.com/tegin/dms/tree/12.0-dms

pedrobaeza (approved):

pedrobaeza:
Let's proceed with the merge. Thanks to all the involved.

/ocabot merge nobump

agent-z28:
WOW!!!!!

PR #5, opened 2020-07-08
[IMP] dms: add migration script for muk_dms

etobella:
Migration script used to migrate muk_dms to dms.

Tested with OpenUpgrade adding a rename from muk_dms to dms on `apriori.py`

### DIFF
--- a/dms/migrations/13.0.1/post-migration.py
+++ b/dms/migrations/13.0.1/post-migration.py
@@ -1,0 +1,9 @@
+# Copyright 2020 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env["dms.directory"].search([])._compute_complete_name()
--- a/dms/migrations/13.0.1/post-migration.py
+++ b/dms/migrations/13.0.1/post-migration.py
@@ -1,9 +1,29 @@
 # Copyright 2020 Creu Blanca
+# Copyright 2020 ForgeFlow S.L.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openupgradelib import openupgrade  # pylint: disable=W7936
 
 
+def recompute_file_content(env):
+    for file in env["dms.file"].search([]):
+        if "." in file.extension:
+            file.extension = file.extension.replace(".", "")
+        try:
+            file.content
+        except Exception:
+            env.cr.execute("SELECT reference FROM dms_file where id = %s", (file.id,))
+            result = env.cr.fetchone()
+            reference_id = int(result[0].split(",")[1])
+            env.cr.execute(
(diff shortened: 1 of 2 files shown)


etobella:
@pedrobaeza @muk-it @keshrath

dms/__manifest__.py
```diff
@@ -36,4 +36,5 @@
     ],
     "images": ["static/description/banner.png"],
     "application": True,
+    "external_dependencies": {"python": ["openupgradelib"]},
```
pedrobaeza: This is not needed

dms/migrations/13.0.1/post-migration.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2020 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
```
pedrobaeza: Disable this linter that is not suitable for migrations:
```suggestion
from openupgradelib import openupgrade  # pylint: disable=W7936
```

requirements.txt
```diff
@@ -0,0 +1 @@
+openupgradelib
```
pedrobaeza: No need of adding this, as migrations are not executed

dms/migrations/13.0.1/pre-migration.py
```diff
@@ -0,0 +1,88 @@
...
+    ("dms.storage", "dms_storage", "storage_files", "storage_file_ids"),
+    ("dms.access.group", "dms_access_group", "directories", "directory_ids"),
+    ("dms.access.group", "dms_access_group", "parent_group", "parent_group_id"),
+    ("dms.access.group", "dms_access_group", "child_groups", "child_group_ids"),
+    ("dms.access.group", "dms_access_group", "groups", "group_ids"),
+    ("dms.access.group", "dms_access_group", "explicit_users", "explicit_user_ids"),
+    ("dms.access.group", "dms_access_group", "users", "user_ids"),
+]
+
+_field_add = [("complete_name", "dms.directory", "dms_directory", "char", False, "dms")]
```
pedrobaeza, line 70: Why do you need to add this field this way?

pedrobaeza (changes requested):

dms/migrations/13.0.1/pre-migration.py
etobella: Because it was failing the migration due to missing `complete_name`. It was strange, but was the only solution I found in order to fix it.

If I remember properly is that complete_name is used in _order, and this was giving issues.

pedrobaeza: OK, I see. Put a comment about that for avoiding the same question in the future.

etobella: Done!

pedrobaeza (approved):
Seems OK for me

LoisRForgeFlow:
Increased compatibility: https://github.com/tegin/dms/pull/2

LoisRForgeFlow (approved):
Works perfectly :+1:

jarroyomorales (approved):
Working as expected!

PR #7, opened 2020-07-27
[12.0][BACKPORT] DMS

### LINKED ISSUE #4: Backport to 12.0/11.0
As we talked on #1 It is interesting to backport the dms module to 12.0 following the OCA guidelines.

There are several steps to follow:

- [x] Check with @muk-it that there is no issue with that @keshrath 
- v12:
  - [x] Create branch in GitHub
  - [x] Backport webpanel (OCA/web#1629)
  - [ ] Migrate the module (#7 )
- v11:
  - [ ] Create branch in GitHub
  - [ ] Backport webpanel
  - [ ] Migrate the module

etobella:
Backport of `dms` module to 12.0

All `depends` have been changed to `onchange`

#4 #1


.travis.yml
```diff
@@ -5,7 +5,7 @@ cache:
     - $HOME/.cache/pre-commit
 
 python:
-  - "3.5"
+  - "3.6"
```
pedrobaeza: I put this specifically to P3.5, as it's the recommended supported Python version, which is then the most accurated one. Why are you changing it?

etobella: Black is not working on 3.5, so I needed to change it

pedrobaeza: OK, I didn't think about that... Well, I suppose there's no problem using P3.6...

etobella: Let me check If I can make it use 3.6 only on linting :wink:

etobella: Fixed!

brian10048:
During functional testing on runbot I noticed the following issue:

**Steps to reproduce**
Go to Directory view
Click the 'Files' button on a directory
image

**Expected Result:**
Show only the files in this directory.

**Actual Result:**
All files are shown, regardless of their directory
image


**Similarly,** 
clicking the 'Directory' button does not filter to show only the subdirectories
image

results in all directories being shown rather than just the subdirectories
image

etobella:
Thanks!!! Comment attended.

brian10048:
Not a problem! Thank you for working on this.

Issue has been resolved.

gaikaz:
Nice work. 
Tested it on runbot, but found one strange little bug.
Clicking the little preview icon on files that cannot be previewed online, gives the file for download, but the name is sometimes very odd like `datas.dll`. It doesn't do that when downloaded from inside the Form view.

image

image

etobella:
Thanks @gaikaz Now it is solved.The same happens on 13.0, So I will backport the commit to 13.0

dms/models/access_groups.py
```diff
@@ -0,0 +1,134 @@
...
+        compute="_compute_count_directories", string="Count Directories"
+    )
+
+    @api.depends("directory_ids")
+    def _compute_count_directories(self):
+        for record in self:
+            record.count_directories = len(record.directory_ids)
+
+    @api.model
+    def _add_magic_fields(self):
```
jarroyomorales: Since no model inherits from this, is this really needed?

jarroyomorales (commented):
Functional review and works as expected.
Regarding the code review, just a question.

I also found a couple of 404 errors when requesting
`GET http://localhost:8069/dms/static/src/img/banner/documents_onboarding_file.png` and `GET http://localhost:8069/dms/static/src/img/banner/documents_onboarding_directory.png`. This is because these files dont exist, should they be added or should we change the path?

dms/models/access_groups.py
etobella: Yes, the original module was intended to be inherited, but not here

jarroyomorales (approved):

pedrobaeza:
Can you squash migration commits? Should latest change removing magic fields be forward-port it to v13?

etobella:
The latest changes will be ported to 13.0, So I think it is better to keep it on a different commit (I will squash them), I will wait if other changes are necessary before creating the PR

pedrobaeza:
OK, merging then:

/ocabot merge nobump

Issue #2, opened 2020-06-08
Roadmap


mileo:
Hi, 

I would like to help with the migration. Do we have any plans for this?

Could I migrate the module https://github.com/muk-it/muk_dms/tree/12.0/muk_dms_attachment ?

Thanks in advance

pedrobaeza:
In fact I would consider this as a must in core module. I think there's not too much sense in embedding the file contents in DB, but putting it as `attachment=True`, and no need to add all of these burden, having automatically the benefits of attachments:

- No duplicated content
- No DB size increase

What do you think, @etobella @keshrath ?

etobella:
Well, Right now it is contempled on the original `dms` module. When you define the `type` of the `dms.storage` you can choose:
* Database: files are embedded on the database
* Filestore: files are managed as attachments, so I think it is not necessary to continue on this.

However, I think it would be interesting to check the version system in order to proceed that way

pedrobaeza:
I was thinking that filestore was for importing an existing server path and parse its contents... but if it's already attachments, I think the other type (database), it's not interesting to be kept.

mileo:
> MuK Documents is a module to create, manage and view files directly within Odoo. MuK Documents Attachment allows to store Odoo Attachments inside the Document Management System. Attachments stored in the document management system are still available as attachments in Odoo. As with all other files, the storage type can be set via the setttings.

I believe that we can have a configuration where the user specifies which models he wants his attachments to become documents in the DMS module.

In addition, the possibility to classify them with tags in a simple way.

Document data in DMS should be easy to search, for example by model (model_id and rec_name).

With the document data well structured we can for example create customizations:

a - Documents attached to the product category, tagged with MRP;
b - Documents attached to the product, tagged with MRP;
c - Documents attached to the Bill of Material, tagged with MRP;

1. Attachments for items a, b, c are automatically displayed in a production order;
2. If you search in DMS by a production order you get the same a,b,c documents;

pedrobaeza:
OK, if the idea is that any attachment in the system in a configurable way - not by default -, becomes a DMS document, I support that idea. Is that what `muk_dms_attachment` does? If so, go ahead with `dms_attachment`!

keshrath:
MuK DMS Attachment uses the DMS to store Odoo Attachments. The basic idea behind that is that you can have all the Odoo Attachment in a structured way.

pedrobaeza:
[...]

PR #9, opened 2020-08-13
[13.0][IMP] Attachment Integration

luismalta:
This feature aims to integrate dms with ir.attachment. That way, every time a new attachment is created, a dms.file is also created with a category according to the model the attachment is in.

### DIFF
--- a/dms/models/__init__.py
+++ b/dms/models/__init__.py
@@ -11,5 +11,4 @@ from . import tag
 
 from . import res_company
 from . import res_config_settings
-
 from . import ir_attachment
--- a/dms/models/__init__.py
+++ b/dms/models/__init__.py
@@ -11,3 +11,5 @@ from . import tag
 
 from . import res_company
 from . import res_config_settings
+
+from . import ir_attachment
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -158,6 +158,12 @@ class File(models.Model):
         ondelete="cascade"
     )
 
+    record_ref = fields.Reference(
+        string='Record',
+        selection='_select_reference',
+        readonly=True,
+    )
+
     # ----------------------------------------------------------
     # Helper
     # ----------------------------------------------------------
@@ -649,3 +655,7 @@ class File(models.Model):
                 )
             else:
                 record.update({"is_locked": False, "is_lock_editor": False})
+
+    def _select_reference(self):
+        model_ids = self.env['ir.model'].search([])
+        return [(r['model'], r['name']) for r in model_ids] + [('', '')]
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -580,14 +580,18 @@ class File(models.Model):
     def _create_model_attachment(self, vals):
 
         if "default_directory_id" in self._context:
-            default_directory_id = self.env["dms.directory"].search(
-                [("id", "=", self._context["default_directory_id"])]
+            default_directory_id = (
+                self.env["dms.directory"]
+                .with_user(SUPERUSER_ID)
+                .browse(self._context["default_directory_id"])
             )
 
             vals["directory_id"] = default_directory_id.id
 
-        directory_id = self.env["dms.directory"].search(
-            [("id", "=", vals["directory_id"])]
+        directory_id = (
+            self.env["dms.directory"]
+            .with_user(SUPERUSER_ID)
+            .browse(vals["directory_id"])
         )
 
         if directory_id and directory_id.record_ref:
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -578,6 +578,14 @@ class File(models.Model):
                 self.browse(ids).write(dict(vals))
 
     def _create_model_attachment(self, vals):
+
+        if "default_directory_id" in self._context:
+            default_directory_id = self.env["dms.directory"].search(
+                [("id", "=", self._context["default_directory_id"])]
+            )
+
+            vals["directory_id"] = default_directory_id.id
+
         directory_id = self.env["dms.directory"].search(
             [("id", "=", vals["directory_id"])]
         )
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -150,6 +150,13 @@ class File(models.Model):
         attachment=True, string="Content File", prefetch=False, invisible=True
     )
 
+    attachment_id = fields.Many2one(
+        comodel_name='ir.attachment',
+        string='Attachment File',
+        prefetch=False,
+        invisible=True
+    )
+
     # ----------------------------------------------------------
     # Helper
     # ----------------------------------------------------------
@@ -172,6 +179,8 @@ class File(models.Model):
         )
         if self.storage_id.save_type == "file":
             new_vals["content_file"] = self.content
+        elif self.storage_id.save_type == "attachment":
+            new_vals['attachment_id'] = self.content
         elif self.storage_id.save_type == "database":
             new_vals["content_binary"] = self.content and binary
         return new_vals
@@ -356,15 +365,20 @@ class File(models.Model):
                 mimetype = guess_mimetype(binary, default="application/octet-stream")
             record.res_mimetype = mimetype
 
-    @api.depends("content_binary", "content_file")
+    @api.depends("content_binary", "content_file", "attachment_id")
     def _compute_content(self):
         bin_size = self.env.context.get("bin_size", False)
         for record in self:
             if record.content_file:
                 context = {"human_size": True} if bin_size else {"base64": True}
                 record.content = record.with_context(context).content_file
-            else:
+            elif record.content_binary:
                 record.content = base64.b64encode(record.content_binary)
+            else:
+                context = {"human_size": True} if bin_size else {
+                    "base64": True}
+                record.content = \
+                    record.with_context(context).attachment_id.datas
 
     @api.depends("content_binary", "content_file")
     def _compute_save_type(self):
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -151,11 +151,11 @@ class File(models.Model):
     )
 
     attachment_id = fields.Many2one(
-        comodel_name='ir.attachment',
-        string='Attachment File',
+        comodel_name="ir.attachment",
+        string="Attachment File",
         prefetch=False,
         invisible=True,
-        ondelete="cascade"
+        ondelete="cascade",
     )
 
     record_ref = fields.Reference(
@@ -185,7 +185,7 @@ class File(models.Model):
         if self.storage_id.save_type == "file":
             new_vals["content_file"] = self.content
         elif self.storage_id.save_type == "attachment":
-            new_vals['attachment_id'] = self.content
+            new_vals["attachment_id"] = self.content
         elif self.storage_id.save_type == "database":
             new_vals["content_binary"] = self.content and binary
         return new_vals
@@ -380,10 +380,8 @@ class File(models.Model):
             elif record.content_binary:
                 record.content = base64.b64encode(record.content_binary)
             else:
-                context = {"human_size": True} if bin_size else {
-                    "base64": True}
-                record.content = \
-                    record.with_context(context).attachment_id.datas
+                context = {"human_size": True} if bin_size else {"base64": True}
+                record.content = record.with_context(context).attachment_id.datas
 
     @api.depends("content_binary", "content_file")
     def _compute_save_type(self):
@@ -632,7 +630,7 @@ class File(models.Model):
 
     @api.model
     def create(self, vals):
-        if not "record_ref" in vals:
+        if "record_ref" not in vals:
             self._create_model_attachment(vals)
 
         return super(File, self).create(vals)
@@ -687,5 +685,5 @@ class File(models.Model):
                 record.update({"is_locked": False, "is_lock_editor": False})
 
     def _select_reference(self):
-        model_ids = self.env['ir.model'].search([])
-        return [(r['model'], r['name']) for r in model_ids] + [('', '')]
+        model_ids = self.env["ir.model"].search([])
+        return [(r["model"], r["name"]) for r in model_ids] + [("", "")]
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -159,9 +159,7 @@ class File(models.Model):
     )
 
     record_ref = fields.Reference(
-        string='Record',
-        selection='_select_reference',
-        readonly=True,
+        string="Record Referenced", selection="_select_reference", readonly=True,
     )
 
     # ----------------------------------------------------------
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -155,7 +155,7 @@ class File(models.Model):
         string='Attachment File',
         prefetch=False,
         invisible=True,
-        ondelete = "cascade"
+        ondelete="cascade"
     )
 
     # ----------------------------------------------------------
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -581,6 +581,31 @@ class File(models.Model):
             for vals, ids in updates.items():
                 self.browse(ids).write(dict(vals))
 
+    def _create_model_attachment(self, vals):
+        directory_id = self.env["dms.directory"].search(
+            [("id", "=", vals["directory_id"])]
+        )
+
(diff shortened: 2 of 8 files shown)


pedrobaeza:
Thanks for the improvement, but not sure if this should go in the base module. @etobella what do you think? In any case, better documentation about this (in the README) and UI tips are needed.

mileo:
Hi, This feature is related with my comments in #2 

I was talking to @luismalta  a few hours ago about this feature and he has already developed this proof of concept to open this PR so that you can also suggest ideas. We are thinking about the following features:

1. User can configure a model at dms.category (Category / Folder), each attachment created in this template is automatically created in this folder.  For example:
         - HR (dms.category) related to hr.employee;
         - HR / Files here;
2. Create a boolean field at dms.category to create a sub-folder with the record model name; ( This behavior can be the default behavior, and we remove the boolean and avoiding the example above that you have multiple files from multiple records in the same folder.)
         - HR / Michel Admin / Files;
2.1 If you create a new file at HR / Michel Admin i'ts automatically attached to hr.employee record of Michel Admin;
3. In the folder or in the files you can click in a link to go to the record.
4. We can have a default folder to the attachments that don't have the model defined in dms.category be saved.

Any suggestions are welcome. Best regards

etobella:
@luismalta interesting approach, but I think this should go on a new module, like `dms_attachment`.

When we integrated the module `muk_dms_attachment` directly to `dms` was about storing dms files using attachments, not creating dms files from attachments.

However, what will happen if two categories or two directories exists for this model?

If you are interested, we are currently working on a new `dms` module called `dms_field`. This module will allow to define directories related to records and define a way to show this directories and its files. @jarroyomorales 

The main idea would be to relate one or more storages to a model, so, each root folder of the storage will be related to a single record, now, we will be able to show all the related folders in a simple way. So, a record could have different folders related to it and a user will be able to see the files that its permissions allow.

etobella:
The only problem is that we are developing it on 12.0, but it will be ported to 13.0 as soon as it is approved :wink:

keshrath:
`muk_dms_attachment` (https://github.com/muk-it/muk_dms/tree/12.0/muk_dms_attachment) does not store dms files as attachments. It allows attachments to be stored as dms files.

`muk_dms_field` (https://github.com/muk-it/muk_dms/tree/12.0/muk_dms_field) is similar to the binary field with `attachment=True` but stores the data as a file instead of an attachment.

etobella:
Sorry for the misunderstanding, we crossed the information between muk_dms_file and muk_dms_attachment.

Our module has nothing to do with muk_dms_field, it is a different approach of how to relate dms.files and records.
One way is to relate each file to a record like the approach of this PR or muk_dms_attachment. 
Our proposal is to relate a root directory to a record, this way we can create a structure of folders inside this record. It might be interesting in cases like customers or employees of different systems of files and classification folders could exist and we could see some of them depending on the permissions.

mileo:
Hi Guys,

Talking with @luismalta  **I found that I made a mistake.** 

I specified it to be implemented in the dms.category model instead of the dms.directory model. 

@etobella did you see my comment about the roadmap of this feature? 

Interesting your idea of verifying the user's permissions to the directory model, we can implement it. Do attachments no longer do this automatically? @luismalta is going to test this case.

mileo:
About the multiple directories per model, in my opinion, this should not be allowed. This will keep the feature very simple and minimalist avoiding we to have to create one module per model.

We can create a default behavior to create a directory per model, with model _description automatically too.

Do you think we really need a separate module for this feature? Is it not simpler to leave it in the same DMS and create some settings.

dms/models/__init__.py
```diff
@@ -11,3 +11,5 @@
 
 from . import res_company
 from . import res_config_settings
+
```
mileo: Remove this space please

mileo:
Video: https://drive.google.com/file/d/1iCp2TOjDOWMN-GBdqvJnp_qEcB2_2tli/view

luismalta:
The last commits put the PR in the way suggested by @mileo, missing only the errors of Travis' tests, which I'm working on at the moment. Therefore, I would like to request the review of the PR again. 
Best regards.
@etobella @pedrobaeza @keshrath

etobella:
We tried a different approach in order to make a similar behavior on #12. Now, we can assign one or more directories to a record and we can use the inherited structure there. I think this simplifies a lot how to manage it. Can you check? Also, a new view has been added in order to make it easy enough

etobella:
Maybe we can create a unified view, like:
Use the dms_field as a view and this module in order to store the attachments in the first directory. WDYT?

luismalta:
hello @etobella 
First, I'm sorry for the delay in responding.
I checked the proposal made in PR #12 and I really liked the new functionality. I also liked the proposal to use this new functionality with the integration of the attachments.

I wish I had already made a proposal for the integration of these PR, but I still haven't managed. I'm working on the migration of the web_view_searchpanel module to 13.0 to be able to test PR #12 and make the proposal to integrate its features in this PR.

Best regards.

pedrobaeza:
web_view_searchpanel is native in v13.

luismalta:
I see, sorry for the misunderstood.

luismalta:
Hello people. I'm integrating the features of dms_field with those of this PR, but I was in doubt where to put these features. Talking to @mileo , some alternatives came up:
1- Maintain the functionalities in the base module, as it is already being done in this PR.
2- Create a third module that depends on dms_field
3 - Integrate the features in dms_field

Do you have any other ideas? What would fit best?
Best Regards

pedrobaeza:
I would go for this to be integrated in base, and make it compatible with `dms_field`, but letting that apart.

etobella:
`dms_field` should be migrated in another PR, not here, but it would be interesting to make it work properly together :wink:
However, just as a comment of this approach, I think we are losing one of the biggest parts of dms with the `attachment` as the permissions are not useful anymore, but this is just my comment. If this is an option there might be no issues with that

pedrobaeza:
Yes, being optional fields, if you want, you don't fill those fields and let that integration. You are right about permissions, but the aim of this is to fill flows that are done by lots of persons in one place, and summarize them automatically in the DMS by others.

luismalta:
I created a migration to dms_field: PR # 17.
I am currently working on integrating dms_field with the attachments features and will soon update this PR with a proposal.

victoralmau:
Hi,

Maybe need to make some changes to aprrove this PR:

- Add [IMP] in subject (this change is an Implementation)
- Update description to add more info (description, usage, demo images, Video link). It's important to know about this functionalitty to review changes.
- Only 1 commit per PR 

This functionallity is very interesting, good job!! 
Now, i check patiently to try solved travis errors and help it.

dms/models/dms_file.py
```diff
@@ -560,6 +577,39 @@ def _inverse_content(self):
             for vals, ids in updates.items():
                 self.browse(ids).write(dict(vals))
 
+    def _create_model_attachment(self, vals):
+
+        if "default_directory_id" in self._context:
+            default_directory_id = self.env["dms.directory"].search(
+                [("id", "=", self._context["default_directory_id"])]
+            )
```
victoralmau: Change is necessary to solved travis
```
            default_directory_id = self.env["dms.directory"].with_user(SUPERUSER_ID).browse(self._context["default_directory_id"])
```
IMO this is a problem with _search function in directory that is not correct (isn't override and is needed) because not apply groups correctlly (not in this PR).

dms/models/dms_file.py
```diff
@@ -560,6 +577,39 @@ def _inverse_content(self):
...
+
+            vals["directory_id"] = default_directory_id.id
+
+        directory_id = self.env["dms.directory"].search(
+            [("id", "=", vals["directory_id"])]
+        )
```
victoralmau: Change is necessary to solved travis
```
        directory_id = self.env["dms.directory"].with_user(SUPERUSER_ID).browse(vals["directory_id"])
```

IMO this is a problem with _search function in directory that is not correct (isn't override and is needed) because not apply groups correctlly (not in this PR).

victoralmau (commented):
With this changes, travis OK

dms/models/dms_file.py
pedrobaeza: Why searching instead of a `browse`?

victoralmau: > Why searching instead of a `browse`?

Thanks, update my comments

luismalta:
@victoralmau Thanks a lot for the help. I will work on these changes to make PR approvable

victoralmau:
Add exactly same functionallity in V12: https://github.com/OCA/dms/pull/24

victoralmau:
Add test in similar PR https://github.com/OCA/dms/pull/24 and maybe is interesting to add it and increase codecov results.

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,85 @@
...
+        file, ext = re.split(r"\.", file_name)
+        return file + " ({}).".format(count) + ext if count > 1 else file_name
+
+    def _create_record_dir(self, storage_id, attachment_id):
+        dms_field = self.env["ir.module.module"].search([("name", "=", "dms_field")])
+
+        directory_fields = {
+            "name": attachment_id.res_name.replace("/", "-"),
+            "is_root_directory": True,
+            "parent_id": False,
```
jarroyomorales, line 27: Wouldnt it be better to add all these directories under a common model directory `Contacts Root` for example instead of making every record root directory?

dms/models/dms_file.py
```diff
@@ -172,6 +184,8 @@ def _update_content_vals(self, vals, binary):
         )
         if self.storage_id.save_type == "file":
             new_vals["content_file"] = self.content
+        elif self.storage_id.save_type == "attachment":
+            new_vals["attachment_id"] = self.content
```
victoralmau, line 188: It's very strange because in V12 don't work https://github.com/OCA/dms/pull/24#issuecomment-703899342.
Solved (V12) with this changes:
```suggestion
                if self.storage_id.save_type in ["file", "attachment"]:
            new_vals["content_file"] = self.content
```

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,85 @@
...
+            "root_storage_id": storage_id.id,
+        }
+
+        if dms_field and dms_field.state == "installed":
+            directory_fields["res_id"] = attachment_id.res_id
```
victoralmau, line 32: IMO it's unnecessary because of (in the future) other addon use it, it's not necessary to add new field `res_id` in directory because `record_ref` contain it already.

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,85 @@
...
+
+        return self.env["dms.directory"].create(directory_fields)
```
victoralmau, line 34: IMO it's unnecessary but need to change some things in `_create`

dms/models/ir_attachment.py
pedrobaeza: I think the standard is to have `res_model` and `res_id` field. The `record_ref` can be a computed one, composed over the other 2.

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,85 @@
...
+                            "name": file_name,
+                            "directory_id": attachments_dir_id.id,
+                            "attachment_id": attachment_id.id,
+                            "record_ref": "{},{}".format(
+                                attachment_id.res_model, attachment_id.res_id
+                            ),
+                        }
+                    )
+
+            return attachment_id
```
victoralmau, line 85: In V12 change it for some reasons:
- It's possible to exists some attachment storage
- It's necessary to auto-create root directory related to model and storage
- It's necessary to create sub directory (related to root directory previously auto-create) related to model and res_id with record_ref field.
- Dms fIeld need to related to previous directory.

```suggestion
       @api.model
    def create(self, vals):
        if "dms_file" in vals and vals["dms_file"]:
            del vals["dms_file"]
            return super(IrAttachment, self).create(vals)
        else:
            attachment_id = super(IrAttachment, self).create(vals)
            if attachment_id.res_model and attachment_id.res_id:
                storage_ids = self.env["dms.storage"].search(
                    [
                        ("model_id.model", "=", attachment_id.res_model),
                        ("save_type", "=", "attachment"),
                    ]
                )
                if storage_ids:
                    record_ref = "{},{}".format(
                        attachment_id.res_model, attachment_id.res_id,
                    )
                    directory_ids = self.env["dms.directory"].search(
                        [
                            ("storage_id", "in", storage_ids.ids),
                            ("record_ref", "=", record_ref),
                        ]
                    )
                    if directory_ids:
                        directory_id = directory_ids[0]
                    else:
                        root_directory_ids = self.env["dms.directory"].search(
                            [
                                ("storage_id", "=", storage_ids[0].id),
                                ("is_root_directory", "=", True),
                            ]
                        )
                        if root_directory_ids:
                            root_directory_id = root_directory_ids[0]
                        else:
                            root_directory_id = self.env["dms.directory"].create(
                                {
                                    "name": attachment_id.res_model,
                                    "is_root_directory": True,
                                    "storage_id": storage_ids[0].id,
                                    "root_storage_id": storage_ids[0].id,
                                }
                            )
                        # Create directory related to record_ref
                        directory_id = self.env["dms.directory"].create(
                            {
                                "name": attachment_id.res_name,
                                "parent_id": root_directory_id.id,
                                "storage_id": root_directory_id.storage_id.id,
                                "record_ref": record_ref,
                            }
                        )
                    # File operations + dms_file create
                    file_name = self._dms_file_name(directory_id, vals["name"])
                    self.env["dms.file"].create(
                        {
                            "name": file_name,
                            "directory_id": directory_id.id,
                            "attachment_id": attachment_id.id,
                            "record_ref": record_ref,
                        }
                    )
            return attachment_id
```

victoralmau (commented):
Additionally need to create test about it with some checks:

- Migration process

- Create `dms_file` and check if attachment create it and related it

- Create attachment related to model previously configured in storage and check if `dms_file` work fine

In V12 https://github.com/OCA/dms/pull/24 exists test about it.

dms/models/ir_attachment.py
victoralmau: Ok, but actually don't exists this fields.
Do you think that it's necessary to remove `record_ref` and add: `res_model` and `res_id` fields?
It's necessary to add this fields only in directory or in file too? (now group all attachment in only one directory, it's enought?)

pedrobaeza: I think it's better to go with `res_model` and `res_id` fields, and the other one computed (stored). If `res_model` can be computed stored as well from the `dms.directory`, perfect. I understand that `dms.directory` can contain `res_model` and optionally `res_id`, isn't it?

Rad0van:
I was just going to ask in contributors mailing list about existence of module that would allow to categorize and/or add tags to attachments. This PR serves nicely for that purpose I guess. What I miss would be integration back to attachment list. What I mean is that I would like to have let's say product.template's attachments displayed in a tree view with category and tags visible/editable. I'd like to use the category and/or tags later to selectively display some of the attachments to e-commerce for example (which would be up to its own module). Is this feasible?

victoralmau:
> I was just going to ask in contributors mailing list about existence of module that would allow to categorize and/or add tags to attachments. This PR serves nicely for that purpose I guess. What I miss would be integration back to attachment list. What I mean is that I would like to have let's say product.template's attachments displayed in a tree view with category and tags visible/editable. I'd like to use the category and/or tags later to selectively display some of the attachments to e-commerce for example (which would be up to its own module). Is this feasible?

If i understand it correctly this PR don't allow it.
Only integrate with attachment files but don't appear in e-commerce (for example). 

Maybe, if you prefer create a new issue about it and explain a lot and everybody comment about it and the best way to do.

victoralmau:
About this functionality apply now in V12 in this PR https://github.com/OCA/dms/pull/24

IMO the best way to add this functionality to V13 is close this PR, create new PR and apply cherry-pick about https://github.com/OCA/dms/pull/24

What do you think about it?

etobella:
I guess @victoralmau is right.

What do you think @luismalta ?

pedrobaeza:
@luismalta @mileo do you agree then to supersed this with #37 ?

mileo:
> @luismalta @mileo do you agree then to supersed this with #37 ?

Yes, Would it be possible to put @luismalta as a contributor on #37?

pedrobaeza:
It's already included with all its content on the first commit of that PR :smiley: 

Closing then this as continuing in #37 

Thanks all

PR #12, opened 2020-08-27
[ADD] dms_field

etobella:
This module creates a new widget that can be used as a view and a field widget.

dms_field 2

Still working on mobile view and a better readme, but it is usable right now.

The view is added as demo on `res.partner`


etobella:
Preview is not working due to #13

pedrobaeza:
I think this one should be harmonized with #9

luismalta (approved):

jarroyomorales:
@etobella I found this minor view mistake on runbot. When installing `web_responsive` it is fixed.

Captura de pantalla de 2020-09-14 13-48-54

dms_field/static/src/xml/preview.xml
```diff
@@ -0,0 +1,95 @@
...
+<templates xml:space="preserve">
+    <t t-name="dms.DocumentTreeViewDirectoryPreview">
+        <div class="o_preview_directory">
+            <div class="o_preview_directory_header">
+                <h3>
+                    <t t-esc="dms_object.data.name" />
+                </h3>
+            </div>
+            <div class="o_preview_directory_body row">
+                <div class="col-sm-3 o_preview_directory_icon row" align="center">
```
jarroyomorales: With `col-sm-5` here and `col-sm-7` in line 36, everything fits better

etobella:
I have improved the view in order to make it responsive. I think now is :green_apple::green_apple::green_apple::green_apple::green_apple:

dms_field/static/src/xml/preview.xml
etobella: I think the current solution is even clearer :wink:

jarroyomorales (approved):

etobella:
@pedrobaeza @luismalta @mileo
According to https://github.com/OCA/dms/pull/9#issuecomment-691976262 I think this is right now. What do you think?
Should we proceed with merge?

mileo (changes requested):
Hi Guys, 

Sorry the delay.

I tested locally and the module works well. 

But i think that one of the model relations can be improved.  For each model we must create a storage, i don't like this structure.

I believe it would be much better if it were possible to associate a directory, or a storage could be associated with multiple models.

For example: Product: Chair of the category: Furniture

The user create a new storage: **Sales and Manufacturing** with the following folders and files.

```
sales/ (Storage/Folder of the Sales Team)
└── furniture (Product Category)
    ├── chair-part-number-1234 (Product  Product)
    │   └── sales-folder.pdf
    ├── furniture-sales-folder.pdf
    └── furniture-sales-instructions.pdf
manufacturing/ (Storage/Folder of the MRP Team)
└── furniture
    ├── chair-part-number-1234
    │   └── assembly-instructions.pdf
    └── wood-furniture-painting-instrucitions.pdf

```
It would be wonderful if this module allowed the creation of this type of directory structure.

etobella:
It is not hard to add integration to both, but I don't understand why this could be better :thinking:
However, @jarroyomorales Let's proceed with the change :wink:

mileo:
> It is not hard to add integration to both, but I don't understand why this could be better 🤔
> However, @jarroyomorales Let's proceed with the change 😉

But you understand that is impossible to create a folder structure like I describe with the actual code?

etobella:
I understand it is impossible, but it was made impossible on purpose. Let me explain why.
Navigation using your proposal is hard, and it is difficult to decide where to add or how to add a new folder related to a element. Also, decisions according how the permissions are managed, where hard to be made. Also, it is sensitive to changes, using your example, if we change the category of the product, what should happen with the folder?

For this reason, we decided to define folders that can be related to records as root folders, this way, all the elements inside this folder are assigned to a record, and just one record. No problems with decisions.

With the current code, we can define multiple kinds of records on a single storage.

dms_field/static/src/js/base/dms_tree_renderer.js
```diff
@@ -0,0 +1,460 @@
...
+                    self._generateActionButton(node, action, $buttons);
+                });
+            }
+        },
+        _generateActionButton: function(node, action, $buttons) {
+            if (action.action) {
+                var $button = $("<button>", {
+                    type: "button",
+                    class: "btn btn-secondary " + action.icon,
+                    "data-toggle": "dropdown",
```
jarroyomorales, line 201: Here it would be great to add `"title": action.label` to get button information on hovering

etobella: Fixed :wink:

victoralmau:
Although i haven't review yet probably cause conflicts with https://github.com/OCA/dms/pull/24 because some PRs use `res_model` and `res_id` in `dms.directory`, `model_ids` in `dms.storage`, etc

dms_field/models/dms_directory.py
```diff
@@ -0,0 +1,243 @@
...
+                    ("id", "!=", directory.id),
+                    ("res_id", "=", directory.res_id),
+                    ("res_model", "=", directory.res_model),
+                    ("is_root_directory", "=", True),
+                ],
+                limit=1,
+            ):
+                raise ValidationError(
+                    _("This record is already related in this storage")
+                )
```
victoralmau: This constraint isn't compatible with https://github.com/OCA/dms/pull/24

dms_field/models/dms_storage.py
```diff
@@ -0,0 +1,56 @@
...
+            if storage.model_ids and self.env["dms.directory"].search(
+                [
+                    ("root_storage_id", "=", storage.id),
+                    ("is_root_directory", "=", True),
+                    ("res_model", "=", False),
+                ]
+            ):
+                raise ValidationError(
+                    _("There are directories not associated to a record")
+                )
```
victoralmau: This constraint isn't compatible with https://github.com/OCA/dms/pull/24

victoralmau:
It is probably necessary to review this addon because something "similar" already exists in dms.
Currently, in dms there is the option of storage attachments (type). Taking this into account, part of the code already exists (define models in storage, res_id and res_model in directories, etc).

How does it currently work?
- Create a storage with type "attachment" + define the models: res.partner + hr.employee (for example).
- Create a root folder for the above storage and relate it to the res.partner model.
Now, when an attachment is added to any contact record (https://github.com/OCA/dms/blob/12.0/dms/models/ir_attachment.py#L63-L68) the folder for that contact is created (in case it does not exist, as a subfolder of the root folder of contacts) and the dms file is created.

etobella:
"Similar" but really different. 
Using your words from #24 
> This feature aims to integrate dms with ir.attachment. That way, every time a new attachment is created, a dms.file is also created with a category according to the model the attachment is in.

However, this module tries to create a widget to show your assigned folders and files.
Why it is interesting, because #24 is creating a dependency between DMS and attachment. So, all your files, will be visible from the attachment view. Using the standard behavior of DMS you can add permissions on a folder (obviously, it can be improved). This can be interesting for HR and also for not mixing data (You can create a storage for medical permissions and will be only accessible by RRHH). Right now, if you create a file from dms, an attachment is created and can be viewed by all users that can access the file. Obviously, this is not perfect IMO.

I will check how to make it work with your change, as, IMO, desired behavior of both approaches are different.

victoralmau:
Ok, considering that they are different things, don't you think it would be a good idea to use the already existing fields of: `res_model` and `res_id` that already exist in files and folders and NOT link them with "attachment" type storages? (That way the files would NOT be synced but they would be linked).

It's just a quick idea.

etobella:
Yes, probably true, using the difference between `database`, `file` and `attachment` it might be possible. I will check. Thanks

etobella:
I needed to apply some changes on `dms` in order to make it work properly. Not a big deal

etobella:
@victoralmau I think now is fixed

dms_field/tests/test_dms_field.py
```diff
@@ -0,0 +1,116 @@
...
+from odoo.tests import SavepointCase
+
+
+class TestDmsField(SavepointCase):
+    def setUp(self):
+        super().setUp()
+        self.storage = self.env["dms.storage"].create(
+            {
+                "name": "DMS Storage",
+                "model_ids": [(4, self.env.ref("base.model_res_partner").id)],
```
victoralmau: Add `type` to prevent future errors (if default value related to type field change probably test fail).

dms_field/tests/test_dms_field.py
```diff
@@ -0,0 +1,116 @@
...
+    def test_check_constrain_multi_directory(self):
+        self.env["dms.directory"].create(
+            {
+                "root_storage_id": self.storage.id,
+                "is_root_directory": True,
+                "name": "First Directory",
+                "res_model": self.partner._name,
+                "res_id": self.partner.id,
+            }
+        )
```
```suggestion
    def _create_directory_vals(self, record):
        return {
            "root_storage_id": self.storage.id,
            "is_root_directory": True,
            "name": record.display_name,
            "res_model": record._name,
            "res_id": record.id,
        }
    
    def test_check_constrain_multi_directory(self):
        self.env["dms.directory"].create(self._create_directory_vals(self.partner))
```
Something like these help to reduce code (re-use `_create_directory_vals` function or similar).

victoralmau (changes requested):
Good job!
Code review OK (runbot not work and is not possible to review fine now).

Some questions:
- What happen if some directory exists related to any record (partner for example) but user haven't access to these directory because access groups restricted read?
- It's necessary to add custom fonts to work? (static/lib/jsTree/themes/proton)

etobella:
Thanks!!! Comments attended

etobella:
>     * What happen if some directory exists related to any record (partner for example) but user haven't access to these directory because access groups restricted read?
The folder will not be shown. Probably a second PR will be necessary to add a group on storages in order to fix/improve this. 
>     * It's necessary to add custom fonts to work? (static/lib/jsTree/themes/proton)
Yes, this is the necessary for the new view.

victoralmau (approved):

victoralmau:
Excuse me @etobella, reviewing 12.0 I have seen that this (https://github.com/OCA/dms/blob/12.0/dms_field/views/dms_storage.xml#L25) makes it the first to load when documents app open. 
IMO you should keep the one with DMS and therefore change `sequence` to 99 (for example).
What do you think about it?

etobella:
Ouch!!! My fault. I might try to make a PR to solve it today or tomorrow

victoralmau:
PR to solve it https://github.com/OCA/dms/pull/63

PR #13, opened 2020-08-27
[12.0][FIX] dms: Fix preview widget

etobella:
This fixes the behavior of the previewer

### DIFF
--- a/dms/static/src/js/fields/preview.js
+++ b/dms/static/src/js/fields/preview.js
@@ -9,7 +9,10 @@ odoo.define("dms.preview", function(require) {
 
     var basic_fields = require("web.basic_fields");
     var registry = require("web.field_registry");
+    var core = require("web.core");
     var DocumentViewer = require("mail.DocumentViewer");
+
+    var QWeb = core.qweb;
 
     var FieldPreviewViewer = DocumentViewer.extend({
         template: "FieldBinaryPreview",
@@ -18,18 +21,32 @@ odoo.define("dms.preview", function(require) {
             this.modelName = model;
             this.fieldName = field;
         },
+        /*
+            We need to overwrite this function in order to ensure that the
+            correct template is used
+        */
+        _updateContent: function() {
+            this.$(".o_viewer_content").html(
+                QWeb.render("FieldBinaryPreview.Content", {widget: this})
+            );
+            this.$(".o_viewer_img").load(_.bind(this._onImageLoaded, this));
+            this.$('[data-toggle="tooltip"]').tooltip({delay: 0});
+            this._reset();
+        },
         _onDownload: function(e) {
             e.preventDefault();
-            window.location =
+            var url = new URL(
                 "/web/content/" +
-                this.modelName +
-                "/" +
-                this.activeAttachment.id +
-                "/" +
-                this.fieldName +
-                "/" +
-                "datas" +
-                "?download=true";
+                    this.modelName +
+                    "/" +
+                    this.activeAttachment.id +
+                    "/" +
+                    this.fieldName,
+                window.location.href
+            );
+            url.searchParams.set("download", true);
+            url.searchParams.set("filename", this.activeAttachment.name);
+            window.location = url.href;
         },
     });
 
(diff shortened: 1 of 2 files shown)


dms/static/src/js/fields/preview.js
```diff
@@ -28,7 +28,7 @@ odoo.define("dms.preview", function(require) {
                 "/" +
                 this.fieldName +
                 "/" +
-                "datas" +
+                this.activeAttachment.name +
```
pedrobaeza: Are you sure about this?

dms/static/src/xml/views.xml
```diff
@@ -57,24 +57,44 @@
...
-    <t t-name="FieldBinaryPreview" t-extend="DocumentViewer">
-        <t t-jquery=".o_document_viewer_content_call" t-operation="replace">
-            <t
-                class="o_document_viewer_content_call"
-                t-call="FieldBinaryPreview.Content"
-            />
-        </t>
+    <t t-name="FieldBinaryPreview">
+        <div
+            class="modal o_modal_fullscreen"
```
pedrobaeza: Why this change?

dms/static/src/js/fields/preview.js
etobella: Yes, I am using https://github.com/odoo/odoo/blob/12.0/addons/web/controllers/main.py#L1029

pedrobaeza: But is that route trustable? May there be conflicts with the name (due to slug)?

pedrobaeza: Isn't better to use the route without that part?

dms/static/src/xml/views.xml
```diff
@@ -57,24 +57,21 @@
...
             </div>
         </t>
     </t>
     <t t-name="FieldBinaryPreview" t-extend="DocumentViewer">
-        <t t-jquery=".o_document_viewer_content_call" t-operation="replace">
-            <t
-                class="o_document_viewer_content_call"
-                t-call="FieldBinaryPreview.Content"
-            />
+        <t t-jquery="[t-call='DocumentViewer.Content']" t-operation="replace">
```
pedrobaeza, line 77: A replace operation is not usually advisable. Isn't there any other option?

etobella: I know that it's a replace, but the new element inherits from DocumentViewer.Content, It was the only way I found easy enough in order to change the logic :thinking:

pedrobaeza: OK if you don't see a better approach, but always remember to add a comment explaining the whys for better later comprehension.

dms/static/src/js/fields/preview.js
etobella: With your proposal the downloaded file will be named as datas.pdf or datas.jpg. 
With mine, the name of the file, will be the name of the file on odoo.

Not a big issue for me, I can change it if you think is better :wink:

pedrobaeza: Ah, OK. Is this handling correctly accented chars and so on?

etobella: I tried several options and I found one that could give problems, so I am checking again :wink:

pedrobaeza: OK, let me know :smiley:

pedrobaeza: Maybe a call to `slug` over that name (is available at this level? or you need a computed field for that?)

etobella: Well, I used the URLSearchParams in order to remove the problem, so, now it should work

pedrobaeza:
Great, for testing this functionally, what should I look before the patch and after?

etobella:
imagen
 This button was not working properly

pedrobaeza:
I get a weird behavior using arrow keys in the preview:

Peek 28-08-2020 20-19

etobella:
Yes, thanks, I have found the issue, now it should be solved :wink:

pedrobaeza (approved):
Is this applicable to v13?

etobella:
Yes, it is the last part :disappointed:

pedrobaeza:
Let's merge this for fixing the problem:

/ocabot merge patch

PR #17, opened 2020-09-14
[13.0][MIG] dms_field

luismalta:


etobella:
@luismalta Travis is red, can you review it? do yo need some help? Do you prefer if we supersede it?

albariera:
Superseded at #132 with @luismalta as co-author

PR #20, opened 2020-09-28
[IMP] dms: Integrate with mail_preview_base

### LINKED ISSUE #15: Preview for audio files
I'm missing a preview for audio files (WAV, MP3, OGG...), which browsers supports natively, but not when trying to preview files from these extensions in DMS. I think this can be a general issue in Odoo preview panel, but not sure right now.

etobella:
This code was removed from dms on OCA/social#604 following the comments of #15 
Adding the dependencies in order to make everything work as expected


dms/__manifest__.py
```diff
@@ -9,7 +9,13 @@
     "license": "LGPL-3",
     "website": "http://github.com/OCA/dms",
     "author": "MuK IT,Odoo Community Association (OCA)",
-    "depends": ["web_drop_target", "mail", "web_view_searchpanel", "http_routing"],
+    "depends": [
+        "web_drop_target",
+        "mail",
+        "web_view_searchpanel",
+        "http_routing",
+        "mail_preview_base",
```
pedrobaeza, line 18: Why adding it as dependency if not technically needed? You are already stating that it's recommended to install, but it shouldn't be mandatory.

etobella: The base is required in order to use the `preview_binary` widget

imagen

I have changed the readme in order to improve the explanation

pedrobaeza: But I think we shouldn't add a new dependency in the middle of the release...

dms/README.rst
```diff
@@ -67,6 +67,14 @@ In order to achieve it you need to:
 You can check all the files that still needs to be migrated from all storages
 and migrate them manually on *Documents -> Configuration -> Migration*
 
+
+Preview
+~~~~~~~
+
+`mail_preview_base` is required for DMS but it is recommended to install all
```
pedrobaeza: Put this in INSTALL.rst anyway

dms/__manifest__.py
pedrobaeza: OK, I see that you are removing some JS code here. Uhm, difficult balance to add or not to add... Question, with current JS code, mail_preview_* work? Does this code interfere with `mail_preview_base`?

etobella: The `mail_preview_base` defines a base of functions that will be used to simplify how to preview files.

It also defines a new widget that will allow to preview the file if we can preview it (preview_binary). This widget is used on DMS and the logic will also be reused on `dms_field` module.

So, base is required in order to make this widget reusable, keeping the widget on `dms` or in `mail_preview_base` is optional, but I preferred to keep it all on the same place.

pedrobaeza: OK, let's go with it

jarroyomorales (approved):

houssine78 (approved):
LGTM no test

PR #22, opened 2020-10-05
[12.0][IMP] dms: Portal

victoralmau:
**Portal functionallity**

Users with portal access view link to "Documents"
In Documents show all "root" directories with explicit access to portal and files it.

This new feature use Access Group field in directories and files. It's neccessary "read" permission to show in portal.

If directory have access group to portal, by default, all subdirectories too, but if "not inherit groups" field not selected dissapear.

Add too "Share" link to Directories and Files to show it with custom link

dms_1
dms_2
dms_3
dms_share

Please @pedrobaeza review it

V13: https://github.com/OCA/dms/pull/25

@Tecnativa TT25647


etobella (changes requested):
The download button did not work on portal. Great Addition :wink:

victoralmau:
> The download button did not work on portal. Great Addition wink

Thanks, updated and work fine now

ernestotejeda (commented):
Great job, thanks. 
I found only two issues: review 'Sort by and filter by' functionalities.

Peek 2020-10-05 15-58

dms/models/directory.py
```diff
@@ -193,6 +205,52 @@ class DmsDirectory(models.Model):
                 """,
     )
 
+    @api.multi
+    def _get_share_url(self, redirect=False, signup_partner=False, pid=None):
+        self.ensure_one()
+        params = {"access_token": self._portal_ensure_token()}
+        url = "/my/dms/directory/%s" % self.id
+        return "{}?{}".format(url, url_encode(params))
```
ernestotejeda: Why not to override _compute_access_url method instead? 
simila to this: https://github.com/OCA/rma/blob/12.0/rma/models/rma.py#L414

victoralmau:
> Great job, thanks.
> I found only two issues: review 'Sort by and filter by' functionalities.
> 
> Peek 2020-10-05 15-58

Thanks, "filter by" removed now (because in searchbar it's possible to select only name) and sort by solved.

jarroyomorales:
Hi @victoralmau! Tested it on runbot and works as expected! The only thing I'm wondering is if it is okay to put it on the base module or if we should create a `dms_portal` module, what do you think?

pedrobaeza:
@jarroyomorales portal functionality is in base for all the core modules (`sale`, `account`, etc), so it's reasonable to have it also bundled together, and the code addition is not too much.

victoralmau:
> Hi @victoralmau! Tested it on runbot and works as expected! The only thing I'm wondering is if it is okay to put it on the base module or if we should create a `dms_portal` module, what do you think?

Thanks! I think that it's necessary to include in base module because "portal feature" haven't dependencies. Portal feature only work if you like.

jarroyomorales:
@victoralmau @pedrobaeza Okay then!

ernestotejeda (commented):
Now everything works very fine.
But, ... I'm not quite sure how it is done to determine what a portal user should and shouldn't see. Anyway, I think there should be an access rule for portal users so that they can only see the documents that they are follower of? What do you think?

victoralmau:
> Now everything works very fine.
> But, ... I'm not quite sure how it is done to determine what a portal user should and shouldn't see. Anyway, I think there should be an access rule for portal users so that they can only see the documents that they are follower of? What do you think?

Thank you for review and comment about it.
In DMS, ACLS are override in dms_security_mixin.py, the idea is create dms_access_group and apply to directories (this groups override ACLs) and simplify "rule access" rewriting ACLs.
Use followers is a good idea and will allow create a general rule about portal BUT it's more difficult to aplly (in a directory for example) if compare it with access group that contains some groups and explicit users. 
Exist a issue https://github.com/OCA/dms/issues/23 about ACL system modify.

dms/controllers/portal.py
```diff
@@ -0,0 +1,175 @@
...
+            if item_sudo.access_token:
+                if not consteq(item_sudo.access_token, access_token):
+                    if model == "dms.file":
+                        if not consteq(
+                            item_sudo.directory_id.access_token, access_token
+                        ):
+                            return False
+                    else:
+                        return False
+            elif not item_sudo.access_token and model == "dms.file":
```
pedrobaeza:
```suggestion
            elif model == "dms.file":
```

pedrobaeza: I think anyways that there should be 2 separate methods for this, as the code gets ugly with no real gain having together (on contrary, you need to constantly check `model == "dms.file"`). What do you think?

dms/models/directory.py
```diff
@@ -36,6 +37,12 @@ class DmsDirectory(models.Model):
 
     _parent_store = True
     _parent_name = "parent_id"
+    """
+    Fix partner id to solved
+    https://github.com/odoo/odoo/blob/12.0/addons/portal/models/portal_mixin.py
+    because partner_id is required
+    """
+    partner_id = fields.Many2one(comodel_name="res.partner", store=False)
```
pedrobaeza: This can be already removed as it's merged in 12.0

ernestotejeda (approved):
Sorry, and thanks for the clarification.
I was confused because I thought that the demo 'DMS Access Groups' (dms.access.group) named 'Admin' and 'Demo' were referring to the Admin and demo users, but I see that they are not. But, I already saw that 'Admin' includes the admin and demo users, and 'Demo' includes the portal user.

dms/models/dms_file.py
```diff
@@ -25,13 +28,20 @@ class File(models.Model):
...
         "mail.activity.mixin",
     ]
 
     _order = "name asc"
+    """
+    Fix partner id to solved
+    https://github.com/odoo/odoo/blob/12.0/addons/portal/models/portal_mixin.py
+    because partner_id is required
+    """
+    partner_id = fields.Many2one(comodel_name="res.partner", store=False)
```
pedrobaeza: To be removed

dms/readme/DESCRIPTION.rst
```diff
@@ -2,3 +2,4 @@ DMS is a module for creating, managing and viewing document files directly
 within Odoo.
 This module is only the basis for an entire ecosystem of apps that extend and
 seamlessly integrate with the document management system.
+This module add portal functionallity to directories and files (related to access group that allow it to read portal group)
```
pedrobaeza:
```suggestion

This module adds portal functionality for directories and files for allowed users, both portal or internal users. You can get as well a tokenized link from a directory or a file for sharing it with any anonymous user.
```

dms/readme/ROADMAP.rst
```diff
@@ -0,0 +1,4 @@
+- Tour test to portal
```
pedrobaeza: Isn't this finally done?

dms/readme/USAGE.rst
```diff
@@ -1,2 +1,6 @@
 The best way to manage the documents is to switch to the Documents view.
 Existing documents can be managed there and new documents can be created.
+
+Portal functionallity
+~~~~~~~~~~~~~~~~~~~~~
+The best way to active portal is create Access group that allow read some users (or groups) and apply to some directories
```
pedrobaeza:
```suggestion
Portal functionality
~~~~~~~~~~~~~~~~~~~~

You can add any portal user to DMS access groups, and then allow that group in directories, so they will see in the portal such directories and their files.

Another possibility is to click on "Share" button inside a directory or a file for obtaining a tokenized link for single access to that resource, no matter if logged or not.
```

dms/security/ir.model.access.csv
```diff
@@ -10,3 +10,7 @@ access_dms_directory_user,dms_directory_user,model_dms_directory,group_dms_user,
 access_dms_file_user,dms_file_user,model_dms_file,group_dms_user,1,1,1,1
 
 access_security_access_groups_user,access_security_access_groups_user,model_dms_access_group,base.group_user,1,1,1,1
+
```
pedrobaeza, line 13: This file shouldn't contain any empty line. Please remove all.

dms/tests/__init__.py
```diff
@@ -1,3 +1,4 @@
+"""
```
pedrobaeza: Remove this that is avoiding to execute the rest of the tests

dms/views/assets.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- License AGPL-3.0 (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+    <template id="dms_frontend_demo" inherit_id="web.assets_frontend">
```
pedrobaeza: Please include this as demo data, as there's not too much sense to load this for production deployments.

pedrobaeza (changes requested):
Latest changes and we can go on!

dms/controllers/portal.py
victoralmau: All changes apply, thanks!
About this suggestion now function contains some custom logic, separated now in 2 similar functions (every function with your model).

Maybe a solution to remove all this logic is describe in https://github.com/OCA/dms/issues/23

ernestotejeda:
Fix the commit name

dms/readme/DESCRIPTION.rst
```diff
@@ -2,3 +2,4 @@ DMS is a module for creating, managing and viewing document files directly
 within Odoo.
 This module is only the basis for an entire ecosystem of apps that extend and
 seamlessly integrate with the document management system.
+This module adds portal functionality for directories and files for allowed users, both portal or internal users. You can get as well a tokenized link from a directory or a file for sharing it with any anonymous user.
```
pedrobaeza, line 6: For proper paragraph recognition in RST, you must leave an empty blank line in between, so please reformat this properly.

dms/readme/USAGE.rst
```diff
@@ -1,2 +1,7 @@
 The best way to manage the documents is to switch to the Documents view.
 Existing documents can be managed there and new documents can be created.
+
+Portal functionality
+~~~~~~~~~~~~~~~~~~~~
+You can add any portal user to DMS access groups, and then allow that group in directories, so they will see in the portal such directories and their files.
```
pedrobaeza, line 7: Missing empty lines for paragraphs (in both).

pedrobaeza (approved):

pedrobaeza:
@etobella can you change your review?

pedrobaeza:
As @etobella doesn't answer, but his comments are honored, let's continue with the merge:

/ocabot merge major

mileo:
Nice work!

PR #24, opened 2020-10-05
[12.0] [IMP] dms: Attachment Integration

victoralmau:
This feature aims to integrate dms with ir.attachment. That way, every time a new attachment is created, a dms.file is also created with a category according to the model the attachment is in.

Related to (V13): https://github.com/OCA/dms/pull/9

Please @pedrobaeza review it

@Tecnativa TT25646

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -271,6 +271,9 @@ class DmsDirectory(models.Model):
         if self.env.user.has_group("base.group_public"):
             res = True
         return res
+    record_ref = fields.Reference(
+        string="Record Referenced", selection="_select_reference", readonly=True,
+    )
 
     @api.model
     def _search_is_hidden(self, operator, value):
@@ -654,3 +657,7 @@ class DmsDirectory(models.Model):
                 DmsDirectory, self.sudo().search([("id", "child_of", self.ids)])
             ).unlink()
         return super().unlink()
+
+    def _select_reference(self):
+        model_ids = self.env["ir.model"].search([])
+        return [(r["model"], r["name"]) for r in model_ids] + [("", "")]
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -271,9 +271,52 @@ class DmsDirectory(models.Model):
         if self.env.user.has_group("base.group_public"):
             res = True
         return res
-    record_ref = fields.Reference(
-        string="Record Referenced", selection="_select_reference", readonly=True,
+
+    allowed_model_ids = fields.Many2many(
+        compute="_compute_allowed_model_ids", comodel_name="ir.model", store=False
     )
+    model_id = fields.Many2one(
+        comodel_name="ir.model",
+        domain="[('id', 'in', allowed_model_ids)]",
+        compute="_compute_model_id",
+        inverse="_inverse_model_id",
+        string="Model",
+        store=True,
+    )
+    res_model = fields.Char(string="Linked attachments model")
+    res_id = fields.Integer(string="Linked attachments record ID")
+    record_ref = fields.Reference(
+        string="Record Referenced", compute="_compute_record_ref", selection=[]
+    )
+    storage_id_save_type = fields.Selection(related="storage_id.save_type", store=False)
+
+    @api.depends("root_storage_id", "storage_id")
+    def _compute_allowed_model_ids(self):
+        for record in self:
+            if record.root_storage_id:
+                record.allowed_model_ids = record.root_storage_id.model_ids.ids
+            elif record.storage_id:
+                record.allowed_model_ids = record.storage_id.model_ids.ids
+
+    @api.depends("res_model")
+    def _compute_model_id(self):
+        for record in self:
+            if not record.res_model:
+                record.model_id = False
+                continue
+            record.model_id = self.env["ir.model"].search(
+                [("model", "=", record.res_model)]
+            )
+
+    def _inverse_model_id(self):
+        for record in self:
+            record.res_model = record.model_id.model
+
+    @api.depends("res_model", "res_id")
+    def _compute_record_ref(self):
+        for record in self:
+            if record.res_model and record.res_id:
+                record.record_ref = "{},{}".format(record.res_model, record.res_id)
 
     @api.model
     def _search_is_hidden(self, operator, value):
@@ -657,7 +700,3 @@ class DmsDirectory(models.Model):
                 DmsDirectory, self.sudo().search([("id", "child_of", self.ids)])
             ).unlink()
         return super().unlink()
-
-    def _select_reference(self):
-        model_ids = self.env["ir.model"].search([])
-        return [(r["model"], r["name"]) for r in model_ids] + [("", "")]
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -186,6 +186,17 @@ class File(models.Model):
                         if directory_item.id == self.directory_id.id:
                             return True
         return res
+    attachment_id = fields.Many2one(
+        comodel_name="ir.attachment",
+        string="Attachment File",
+        prefetch=False,
+        invisible=True,
+        ondelete="cascade",
+    )
+
+    record_ref = fields.Reference(
+        string="Record Referenced", selection="_select_reference", readonly=True,
+    )
 
     # ----------------------------------------------------------
     # Helper
@@ -209,6 +220,8 @@ class File(models.Model):
         )
         if self.storage_id.save_type == "file":
             new_vals["content_file"] = self.content
+        elif self.storage_id.save_type == "attachment":
+            new_vals["attachment_id"] = self.content
         elif self.storage_id.save_type == "database":
             new_vals["content_binary"] = self.content and binary
         return new_vals
@@ -393,7 +406,7 @@ class File(models.Model):
                 mimetype = guess_mimetype(binary, default="application/octet-stream")
             record.res_mimetype = mimetype
 
-    @api.depends("content_binary", "content_file")
+    @api.depends("content_binary", "content_file", "attachment_id")
     def _compute_content(self):
         bin_size = self.env.context.get("bin_size", False)
         for record in self:
@@ -406,6 +419,9 @@ class File(models.Model):
                     if bin_size
                     else base64.b64encode(record.content_binary)
                 )
+            else:
+                context = {"human_size": True} if bin_size else {"base64": True}
+                record.content = record.with_context(context).attachment_id.datas
 
     @api.depends("content_binary", "content_file")
     def _compute_save_type(self):
@@ -617,6 +633,43 @@ class File(models.Model):
             for vals, ids in updates.items():
                 self.browse(ids).write(dict(vals))
 
+    def _create_model_attachment(self, vals):
+
+        if "default_directory_id" in self._context:
+            default_directory_id = (
+                self.env["dms.directory"]
+                .with_user(SUPERUSER_ID)
+                .browse(self._context["default_directory_id"])
+            )
+
+            vals["directory_id"] = default_directory_id.id
+
+        directory_id = (
+            self.env["dms.directory"]
+            .with_user(SUPERUSER_ID)
+            .browse(vals["directory_id"])
+        )
+
+        if directory_id and directory_id.record_ref:
+
+            if directory_id.record_ref:
+                model_name = directory_id.record_ref._name
+                model_id = directory_id.record_ref.id
+
+                attachment_id = self.env["ir.attachment"].create(
+                    {
+                        "name": vals["name"],
+                        "datas": vals["content"],
+                        "res_model": model_name,
+                        "res_id": model_id,
+                        "dms_file": True,
+                    }
+                )
+
+                vals["attachment_id"] = attachment_id.id
+                vals["record_ref"] = "{},{}".format(model_name, model_id)
+                del vals["content"]
+
     def copy(self, default=None):
         self.ensure_one()
         default = dict(default or [])
@@ -642,6 +695,13 @@ class File(models.Model):
         # We need to do sudo because we don't know when the related groups
         # will be deleted
         return super(File, self.sudo()).unlink()
+
+    @api.model
+    def create(self, vals):
+        if "record_ref" not in vals:
+            self._create_model_attachment(vals)
+
+        return super(File, self).create(vals)
 
     # ----------------------------------------------------------
     # Locking fields and functions
@@ -691,3 +751,7 @@ class File(models.Model):
                 )
             else:
                 record.update({"is_locked": False, "is_lock_editor": False})
+
+    def _select_reference(self):
+        model_ids = self.env["ir.model"].search([])
+        return [(r["model"], r["name"]) for r in model_ids] + [("", "")]
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -186,6 +186,7 @@ class File(models.Model):
                         if directory_item.id == self.directory_id.id:
                             return True
         return res
+
     attachment_id = fields.Many2one(
         comodel_name="ir.attachment",
         string="Attachment File",
@@ -193,10 +194,22 @@ class File(models.Model):
         invisible=True,
         ondelete="cascade",
     )
-
+    res_model = fields.Char(string="Linked attachments model")
+    res_id = fields.Integer(string="Linked attachments record ID")
     record_ref = fields.Reference(
-        string="Record Referenced", selection="_select_reference", readonly=True,
+        string="Record Referenced",
+        compute="_compute_record_ref",
+        selection=[],
+        readonly=True,
+        store=False,
     )
+    storage_id_save_type = fields.Selection(related="storage_id.save_type", store=False)
+
+    @api.depends("res_model", "res_id")
+    def _compute_record_ref(self):
+        for record in self:
+            if record.res_model and record.res_id:
+                record.record_ref = "{},{}".format(record.res_model, record.res_id)
 
     # ----------------------------------------------------------
     # Helper
@@ -218,11 +231,9 @@ class File(models.Model):
                 "size": binary and len(binary) or 0,
             }
         )
-        if self.storage_id.save_type == "file":
+        if self.storage_id.save_type in ["file", "attachment"]:
             new_vals["content_file"] = self.content
-        elif self.storage_id.save_type == "attachment":
-            new_vals["attachment_id"] = self.content
-        elif self.storage_id.save_type == "database":
+        else:
             new_vals["content_binary"] = self.content and binary
         return new_vals
 
@@ -419,7 +430,7 @@ class File(models.Model):
                     if bin_size
                     else base64.b64encode(record.content_binary)
                 )
-            else:
+            elif record.attachment_id:
                 context = {"human_size": True} if bin_size else {"base64": True}
                 record.content = record.with_context(context).attachment_id.datas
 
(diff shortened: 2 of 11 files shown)


pedrobaeza:
Aren't you able to cherry-pick v13 commit and then put changes for v12 on top of? This way, you preserve the attribution.

ernestotejeda (commented):
Code and functional review :+1: . Looks good.
But, I think you should include unit tests for the new functionality

ernestotejeda:
@victoralmau I got an error creating a file in an type-attachment storage:
Peek 2020-10-05 17-24

victoralmau:
> @victoralmau I got an error creating a file in an type-attachment storage:
> Peek 2020-10-05 17-24

Thanks for report this error, solved now.

victoralmau:
> Code and functional review +1 . Looks good.
> But, I think you should include unit tests for the new functionality

Thanks, althought in original PR not exist tests, added now.

ernestotejeda (approved):

pedrobaeza:
Please don't solve conflicts through merge commits as stated other times.

victoralmau:
> Please don't solve conflicts through merge commits as stated other times.

Sorry, solved it (rebase)

dms/models/directory.py
```diff
@@ -272,6 +272,71 @@ def check_access(self, operation, raise_exception=False):
...
+    )
+    model_id = fields.Many2one(
+        comodel_name="ir.model",
+        domain="[('id', 'in', allowed_model_ids)]",
+        compute="_compute_model_id",
+        inverse="_inverse_model_id",
+        string="Model",
+        store=False,
+    )
+
```
pedrobaeza: Why leaving this empty lines in between?

dms/models/directory.py
```diff
@@ -272,6 +272,71 @@ def check_access(self, operation, raise_exception=False):
...
+    res_model = fields.Char(string="Res model")
```
pedrobaeza: Better to put a proper label

```suggestion
    res_model = fields.Char(string="Linked attachments model")
```

dms/models/directory.py
```diff
@@ -272,6 +272,71 @@ def check_access(self, operation, raise_exception=False):
...
+
+    res_id = fields.Integer(string="Res id")
```
```suggestion
    res_id = fields.Integer(string="Linked attachments record ID")
```

dms/models/directory.py
```diff
@@ -272,6 +272,71 @@ def check_access(self, operation, raise_exception=False):
...
+
+    record_ref = fields.Reference(
+        string="Record Referenced",
+        compute="_compute_record_ref",
+        selection=[],
+        readonly=True,
+        store=False,
```
pedrobaeza: These two attributes are the default one, so you can omit them if you want

dms/models/directory.py
```diff
@@ -272,6 +272,71 @@ def check_access(self, operation, raise_exception=False):
...
+    )
+
+    storage_id_save_type = fields.Char(
+        compute="_compute_storage_id_save_type", store=False
```
pedrobaeza: This can be a related field:
```suggestion
        related="storage_id.save_type", store=False
```

dms/models/directory.py
```diff
@@ -272,6 +272,71 @@ def check_access(self, operation, raise_exception=False):
...
+
+    def _compute_allowed_model_ids(self):
+        for record in self:
+            if record.root_storage_id:
+                record.allowed_model_ids = record.root_storage_id.model_ids.ids
+            elif record.storage_id:
+                record.allowed_model_ids = record.storage_id.model_ids.ids
+
+    @api.onchange("root_storage_id")
+    def _onchange_root_storage_id(self):
```
pedrobaeza: This mustn't be this way, but adding `@api.depends("root_storage_id")` on the compute method

dms/models/directory.py
```diff
@@ -272,6 +272,71 @@ def check_access(self, operation, raise_exception=False):
...
+        self._compute_allowed_model_ids()
+
+    @api.onchange("storage_id")
+    def _onchange_storage_id(self):
+        self._compute_allowed_model_ids()
```
pedrobaeza: The same as previous comment

dms/models/directory.py
```diff
@@ -272,6 +272,71 @@ def check_access(self, operation, raise_exception=False):
...
+    allowed_model_ids = fields.Many2many(
+        compute="_compute_allowed_model_ids", comodel_name="ir.model", store=False
+    )
+    model_id = fields.Many2one(
+        comodel_name="ir.model",
+        domain="[('id', 'in', allowed_model_ids)]",
+        compute="_compute_model_id",
+        inverse="_inverse_model_id",
+        string="Model",
+        store=False,
```
pedrobaeza: I'm thinking that for performance reasons, this can be stored.

dms/models/directory.py
pedrobaeza: Maybe for perfomance reasons is better to store this

dms/models/directory.py
```diff
@@ -272,6 +272,71 @@ def check_access(self, operation, raise_exception=False):
...
+    def _onchange_root_storage_id(self):
+        self._compute_allowed_model_ids()
+
+    @api.onchange("storage_id")
+    def _onchange_storage_id(self):
+        self._compute_allowed_model_ids()
+
+    @api.depends("res_model")
+    def _compute_model_id(self):
+        for record in self:
```
pedrobaeza, line 303:
```suggestion
        for record in self:
            if not record.res_model:
                record.model_id = False
                continue
```

dms/models/dms_file.py
```diff
@@ -187,6 +187,41 @@ def check_access_token(self, access_token=False):
                             return True
         return res
 
+    attachment_id = fields.Many2one(
+        comodel_name="ir.attachment",
+        string="Attachment File",
```
pedrobaeza, line 192: The same about empty lines and field labels.

dms/models/dms_file.py
```diff
@@ -187,6 +187,41 @@ def check_access_token(self, access_token=False):
...
+    record_ref = fields.Reference(
+        string="Record Referenced",
+        compute="_compute_record_ref",
+        selection=[],
+        readonly=True,
+        store=False,
+    )
+
+    storage_id_save_type = fields.Char(
+        compute="_compute_storage_id_save_type", store=False
```
pedrobaeza: related

dms/models/dms_file.py
```diff
@@ -617,6 +655,30 @@ def _inverse_content(self):
             for vals, ids in updates.items():
                 self.browse(ids).write(dict(vals))
 
+    def _create_model_attachment(self, vals):
+        if "default_directory_id" in self._context:
```
pedrobaeza: Don't use `self._context`, but `self.env.context`. I see also that the context is something very weak, and more treated as a default value. It should be in `vals` already if such context is there.

Another thing is that this create must be @api.model_create_multi for not losing performance.

dms/models/dms_file.py
```diff
@@ -617,6 +655,30 @@ def _inverse_content(self):
...
+            vals["directory_id"] = default_directory_id.id
+        directory_id = self.env["dms.directory"].sudo().browse(vals["directory_id"])
+        if directory_id and directory_id.res_model and directory_id.res_id:
+            attachment_id = self.env["ir.attachment"].create(
+                {
+                    "name": vals["name"],
+                    "datas": vals["content"],
+                    "res_model": directory_id.res_model,
+                    "res_id": directory_id.res_id,
+                    "dms_file": True,
```
pedrobaeza: This trick is not the usual and can be conflicting. Please use instead a context key.

dms/models/dms_file.py
```diff
@@ -617,6 +655,30 @@ def _inverse_content(self):
...
 
+    def _create_model_attachment(self, vals):
+        if "default_directory_id" in self._context:
+            default_directory_id = (
+                self.env["dms.directory"]
+                .sudo()
+                .browse(self._context["default_directory_id"])
+            )
+            vals["directory_id"] = default_directory_id.id
+        directory_id = self.env["dms.directory"].sudo().browse(vals["directory_id"])
```
pedrobaeza: This is a recordset, not an id, so the variable must be `directory`.

dms/models/dms_file.py
```diff
@@ -406,6 +441,9 @@ def _compute_content(self):
                     if bin_size
                     else base64.b64encode(record.content_binary)
                 )
+            else:
```
pedrobaeza:
```suggestion
            elif record.attachment_id:
```

dms/models/dms_file.py
```diff
@@ -643,6 +705,12 @@ def unlink(self):
         # will be deleted
         return super(File, self.sudo()).unlink()
 
+    @api.model
```
pedrobaeza: Do it multi-record creation compatible

dms/models/dms_file.py
```diff
@@ -617,6 +655,30 @@ def _inverse_content(self):
...
+            attachment_id = self.env["ir.attachment"].create(
+                {
+                    "name": vals["name"],
+                    "datas": vals["content"],
+                    "res_model": directory_id.res_model,
+                    "res_id": directory_id.res_id,
+                    "dms_file": True,
+                }
+            )
+            vals["attachment_id"] = attachment_id.id
```
pedrobaeza: Instead of modifying current `vals` dictionary, it's always a better practice to copy the dictionary and return that modified copy.

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,59 @@
...
+
+    _inherit = "ir.attachment"
+
+    def _get_dms_directories(self, res_model, res_id):
+        return self.env["dms.directory"].search(
+            [("res_model", "=", res_model), ("res_id", "=", res_id)]
+        )
+
+    @api.model
+    def create(self, vals):
```
pedrobaeza: Multi-creation aware please

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,59 @@
...
+        if "dms_file" in vals and vals["dms_file"]:
+            del vals["dms_file"]
+            return super(IrAttachment, self).create(vals)
+        else:
+            attachment_id = super(IrAttachment, self).create(vals)
```
pedrobaeza: The variable is a recordset, so without `_id`

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,59 @@
...
+            if attachment_id.res_model and attachment_id.res_id:
+                directory_ids = self._get_dms_directories(
+                    attachment_id.res_model, attachment_id.res_id
+                )
+                if not directory_ids:
```
pedrobaeza: Please try to improve this ugly code with so many cyclomatic complexity and bad variable names.

dms/models/storage.py
```diff
@@ -80,6 +84,8 @@ class Storage(models.Model):
         compute="_compute_count_storage_files", string="Count Files"
     )
 
+    model_ids = fields.Many2many("ir.model", string="Models")
```
pedrobaeza:
```suggestion
    model_ids = fields.Many2many("ir.model", string="Linked Models")
```

pedrobaeza (changes requested):

pedrobaeza:
Put in the roadmap to add a migration procedure for converting an storage to attachment one for populating existing records with attachments as folders. Another roadmap point is to add a link from attachment view in chatter to linked documents.

dms/models/directory.py
```diff
@@ -272,6 +272,40 @@ def check_access(self, operation, raise_exception=False):
             res = True
         return res
 
+    allowed_model_ids = fields.Many2many(
+        compute="_compute_allowed_model_ids", comodel_name="ir.model", store=False
+    )
+    model_id = fields.Many2one(
+        comodel_name="ir.model",
+        domain="[('id', 'in', allowed_model_ids)]",
+        inverse="_inverse_model_id",
```
pedrobaeza, line 282: No computed and stored? This will remain in unsync content if you change `res_model`.

dms/models/dms_file.py
```diff
@@ -643,6 +699,13 @@ def unlink(self):
         # will be deleted
         return super(File, self.sudo()).unlink()
 
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if "res_model" not in vals and "res_id" not in vals:
+                vals = self._create_model_attachment(vals)
+        return super(File, self).create(vals_list)
```
pedrobaeza: This is not doing the changes, as the new `vals` is not on `vals_list`. You have to build a new vals_list:
```suggestion
        new_vals_list = []
        for vals in vals_list:
            if "res_model" not in vals and "res_id" not in vals:
                vals = self._create_model_attachment(vals)
            new_vals_list.append(vals)
        return super(File, self).create(new_vals_list)
```

dms/models/dms_file.py
```diff
@@ -617,6 +644,35 @@ def _inverse_content(self):
             for vals, ids in updates.items():
                 self.browse(ids).write(dict(vals))
 
+    def _create_model_attachment(self, vals):
+        res_vals = vals.copy()
+        if self.env.context.get("default_directory_id"):
```
pedrobaeza: This is still using the infamous context key already commented about. Why not using `vals["directory_id"]`? And why you need to browse and then get only the ID?

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,64 @@
...
+    _inherit = "ir.attachment"
+
+    def _get_dms_directories(self, res_model, res_id):
+        return self.env["dms.directory"].search(
+            [("res_model", "=", res_model), ("res_id", "=", res_id)]
+        )
+
+    def _dms_operations(self, attachment_ids):
+        for attachment in attachment_ids:
+            if attachment.res_model and attachment.res_id:
```
pedrobaeza: Save one indentation level, as this still has a big cyclomatic complexity:
```suggestion
            if not attachment.res_model or not attachment.res_id:
                continue
```

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,64 @@
...
+                directories = self._get_dms_directories(
+                    attachment.res_model, attachment.res_id
+                )
+                if not directories:
+                    # Auto-create directories
+                    directories_without_res_id = self._get_dms_directories(
```
pedrobaeza: Maybe it's better to extract this part to a method.

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,64 @@
...
+                                "parent_id": directory.id,
+                                "storage_id": directory.storage_id.id,
+                            }
+                        )
+                    # Get dms_directories again (with items previously created)
+                    directories = self._get_dms_directories(
+                        attachment.res_model, attachment.res_id
+                    )
+                # Auto-create_files
+                if directories:
```
pedrobaeza: No need of this `if`. If no `directories`, `for` loop doesn't iterate. If the problem is that `directories` can be `False` or `None`, then please change that method for always returning a recordset, but empty. It's a good practice to be homogeneous on the returned value for a method.

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,64 @@
...
+                        )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        if self.env.context.get("dms_file"):
+            return super(IrAttachment, self).create(vals_list)
+        else:
+            res = super(IrAttachment, self).create(vals_list)
+            self._dms_operations(res)
+            return res
```
pedrobaeza: Go to more elegant code:
```suggestion
        records = super().create(vals_list)
        if not self.env.context.get("dms_file"):
            records._dms_operations()
        return records
```

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,64 @@
...
+
+    _inherit = "ir.attachment"
+
+    def _get_dms_directories(self, res_model, res_id):
+        return self.env["dms.directory"].search(
+            [("res_model", "=", res_model), ("res_id", "=", res_id)]
+        )
+
+    def _dms_operations(self, attachment_ids):
+        for attachment in attachment_ids:
```
pedrobaeza: This is not IDs, but recordset. Please name properly variables, or later these names will provoke code errors. Anyway, you are already on the proper model, so no need of extra argument (see the later suggestion):
```suggestion
    def _dms_operations(self):
        for attachment in self:
```

pedrobaeza (changes requested):

dms/models/directory.py
victoralmau: I think in another review you suggested removing `store=False` and not computed (to improve performance).
The `res_model` field should not be defined from anywhere other than through the `model_id` so I'm not sure this is a problem.

pedrobaeza: No, I only meant to be store=True, but not removing the computation. Yes, it can be a problem for programatically adding things.

dms/models/dms_file.py
victoralmau: Honestly i don't understand because `default_directory_id` is needed but now is used in some places:

https://github.com/OCA/dms/blob/12.0/dms/models/res_company.py#L109
https://github.com/OCA/dms/blob/12.0/dms/static/src/js/views/many_drop_target.js#L83
https://github.com/OCA/dms/blob/12.0/dms/views/directory.xml#L45

Try to change it and i search a better way?

pedrobaeza: As said, that's a context value that serves for having default values populated, but you mustn't relay on it, but on vals dictionary, as if you programatically create a record, you must provide 2 times same data. Do you want to meet for more explanations?

dms/models/ir_attachment.py
victoralmau: Yes, it's a good idea, separated now.

dms/models/directory.py
victoralmau: Ok, sorry for my confusion about it.

dms/models/directory.py
```diff
@@ -272,6 +272,49 @@ def check_access(self, operation, raise_exception=False):
...
+                record.allowed_model_ids = record.root_storage_id.model_ids.ids
+            elif record.storage_id:
+                record.allowed_model_ids = record.storage_id.model_ids.ids
+
+    @api.depends("res_model")
+    def _compute_model_id(self):
+        for record in self:
+            if not record.res_model:
+                record.model_id = False
+                continue
```
pedrobaeza, line 306: And the rest of the code?

dms/models/ir_attachment.py
```diff
@@ -0,0 +1,61 @@
...
+        items = self._get_dms_directories(self.res_model, False)
+        for item in items:
+            model_item = self.env[self.res_model].sudo().browse(self.res_id)
+            ir_model_item = self.env["ir.model"].search(
+                [("model", "=", self.res_model)]
+            )
+            self.env["dms.directory"].create(
+                {
+                    "name": model_item.display_name,
+                    "model_id": ir_model_item.id,
```
pedrobaeza, line 24: It was OK setting as well `res_model`. The problem you found is because the _compute_model_id is not completed right now.

pedrobaeza (changes requested):

pedrobaeza (approved):

pedrobaeza:
OK, now it's good to continue. I merge as talked with @etobella by chat:

/ocabot merge major

pedrobaeza:
Now I think it's better to cherry-pick this into v13 than continue on the other PR. Please tell in the PR to see what the other contributors think.

PR #28, opened 2020-10-18
[13.0][ADD] DMS Version Control

### LINKED ISSUE #3: Versioning
After the merge of #1, it is interesting to talk about versions. Original @Muk-it had a custom module but I think it is on their private repos. I am interested on migrating/create this set of modules and any starting work is welcome. :smile: 

@keshrath @pedrobaeza

keshrath:
image
image


keshrath:
I could use a little help to get the Travis CI green. Currently I get the following error if I try to run pre-commit:

```
Return code: 1
Expected return code: 0
Output: (none)
Errors:
    npm ERR! code ERESOLVE
    npm ERR! ERESOLVE unable to resolve dependency tree
    npm ERR!
    npm ERR! While resolving: prettier@1.19.1
    npm ERR! Found: @angular/compiler@8.2.13
    npm ERR! node_modules/@angular/compiler
    npm ERR!   @angular/compiler@"8.2.13" from the root project
    npm ERR!
    npm ERR! Could not resolve dependency:
    npm ERR! peer @angular/compiler@"^6.0.0 || ^7.0.0" from angular-estree-parser@1.1.5
    npm ERR! node_modules/angular-estree-parser
    npm ERR!   angular-estree-parser@"1.1.5" from the root project
```

dms_version/models/file.py
```diff
@@ -0,0 +1,54 @@
...
+_logger = logging.getLogger(__name__)
+
+class File(models.Model):
+    
+    _inherit = 'dms.file'
+    
+    version_ids = fields.One2many(
+        comodel_name='dms.version',
+        inverse_name="file_id",
+        stiring="Versions",
```
etobella: string

dms_version/models/storage.py
```diff
@@ -0,0 +1,80 @@
...
+        help="Compress versions to minimize the used space.")
+    
+    incremental_versions = fields.Boolean(
+        string="Only Save Changes", 
+        default=False,
+        help="Only save the incremental changes between two versions.")
+            
+    def action_clean_file_versions(self):
+        versions = self.env['dms.version'].sudo()
+        files = self.mapped('storage_files').sudo()
```
etobella: 'storage_file_ids'

dms_version/models/storage.py
```diff
@@ -0,0 +1,80 @@
...
+        versions.clean_versions(files)
+        
+    def action_delete_file_versions(self):
+        self.mapped('storage_files.versions').sudo().unlink()
```
etobella: 'storage_file_ids.version_ids'

etobella:
Interesting approach :wink: some small issues were found, but easily resolved.

About pre-commit, it is weird :cry: if you need it, we can check it and resolve the pre-commit, but it will be weird if you need some changes :thinking: Did you tried to use a virtualenv? Just as an idea...

dms_version/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "website": "http://github.com/OCA/dms",
+    "author": "MuK IT,Odoo Community Association (OCA)",
+    "depends": ["dms"],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/file.xml",
+        "views/version.xml",
+        "views/storage.xml",
+    ],
+    "application": False,
```
victoralmau, line 19:
```suggestion
    "application": False,
    "external_dependencies": {"python": ["bsdiff4"]},
```
Need to add bsdiff4 dependency to addon work fine.

dms_version/models/storage.py
```diff
@@ -0,0 +1,86 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
+
+try:
+    import bsdiff4
+except ImportError:
+    bsdiff4 = False
```
victoralmau:
```suggestion
import bsdiff4
```
If you define `external_dependencies` in manifest about it not necessary make try except (because except is impossible and because it's impossible to codecov check too)

dms_version/models/storage.py
```diff
@@ -0,0 +1,86 @@
...
+        domain = [("has_versioning", "=", True), ("clean_versions", "=", "autovacuum")]
+        self.search(domain).action_clean_file_versions()
+
+    @api.constrains("incremental_versions")
+    def _check_incremental_versions(self):
+        for record in self:
+            if record.incremental_versions and not bsdiff4:
+                raise ValidationError(
+                    _("Incremental Version Control needs the `bsdiff4` Python library.")
+                )
```
victoralmau:
```suggestion
```

IMO this constraint is not necessary with previous suggestions (because it's impossible).

dms_version/models/version.py
```diff
@@ -0,0 +1,185 @@
...
+
+from odoo import _, api, fields, models, tools
+from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
+
+try:
+    import bsdiff4
+except ImportError:
+    bsdiff4 = False
```
victoralmau:
```suggestion
import bsdiff4
```

Similar to previous change.

dms_version/models/version.py
```diff
@@ -0,0 +1,185 @@
...
+        new_version = self.create(
+            {
+                "name": file.name,
+                "file_id": file.id,
+                "content": content,
+                "previous_version_id": record.id,
+            }
+        )
+        record.write({"next_version_id": new_version.id})
+        return new_version
```
victoralmau: IMO always create two versions and maybe other similar options is something like this

```suggestion
        domain = [
            ("file_id", "=", file.id),
        ]
        record = self.search(domain, order='id desc' limit=1)
        if not record:
            return self.create(
                {"name": file.name, "file_id": file.id, "content": content}
            )
        else:        
            new_version = self.create(
                {
                    "name": file.name,
                    "file_id": file.id,
                    "content": content,
                    "previous_version_id": record.id,
                }
            )
            record.write({"next_version_id": new_version.id})
            return new_version
```

dms_version/tests/test_storage.py
```diff
@@ -0,0 +1,169 @@
...
+from odoo.addons.dms.tests.common import multi_users
+from odoo.addons.dms.tests.test_storage import StorageTestCase
+
+_path = os.path.dirname(os.path.dirname(__file__))
+_logger = logging.getLogger(__name__)
+
+try:
+    import bsdiff4
+except ImportError:
+    bsdiff4 = False
```
victoralmau:
```suggestion
import bsdiff4
```

victoralmau (changes requested):
Good job!
Check Travis and try to solved it

dms_version/__manifest__.py
keshrath: The bsdiff4 is not a strict requirement. You can use the version control without it as well. Only if you want to have incremental versions the dependency ist needed.

etobella: If it is not required, maybe we can do two modules (just to avoid this kind of checks), `dms_version` and `dms_version_incremental`, that will implement the incremental part. What do you think?

pedrobaeza: I personally think that directly the library requirement should be mandatory and don't maintain 2 versions of code, as if you need versioning, the diff/incremental way is always more efficient, and the requirement is very light (only a library). WDYT?

keshrath: IMO: To make an extra module seems excessive for such a small extension. I have no problem making it a strict requirement. Technically it is not, but I guess most will use incremental versions to reduce the needed disk space.

pedrobaeza: Yeah, that's why I say directly to not have other option except incremental.

etobella: One version then :wink:

dms_version/models/version.py
```diff
@@ -0,0 +1,182 @@
...
+                    }
+                )
+                remove_versions.unlink()
+
+    @api.model
+    def add_version(self, file, content):
+        domain = [
+            ("file_id", "=", file.id),
+        ]
+        record = self.search(domain, order='id desc' limit=1)
```
victoralmau:
```suggestion
        record = self.search(domain, order='id desc', limit=1)
```

dms_version/models/version.py
```diff
@@ -0,0 +1,182 @@
...
+import logging
+import zlib
+from collections import defaultdict
+
+from odoo import _, api, fields, models, tools
+from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
+
+import bsdiff4
```
victoralmau:
```suggestion
import base64
import logging
import zlib
from collections import defaultdict
import bsdiff4

from odoo import _, api, fields, models, tools
from odoo.exceptions import ValidationError

_logger = logging.getLogger(__name__)
```

dms_version/models/storage.py
```diff
@@ -0,0 +1,76 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import logging
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
+
+import bsdiff4
```
victoralmau:
```suggestion
import logging

from odoo import _, api, fields, models
from odoo.exceptions import ValidationError

_logger = logging.getLogger(__name__)
```

'bsdiff4' imported but unused

dms_version/tests/test_storage.py
```diff
@@ -0,0 +1,166 @@
...
+import os
+import unittest
+
+from odoo.addons.dms.tests.common import multi_users
+from odoo.addons.dms.tests.test_storage import StorageTestCase
+
+_path = os.path.dirname(os.path.dirname(__file__))
+_logger = logging.getLogger(__name__)
+
+import bsdiff4
```
victoralmau:
```suggestion
import logging
import os
import unittest
import bsdiff4

from odoo.addons.dms.tests.common import multi_users
from odoo.addons.dms.tests.test_storage import StorageTestCase

_path = os.path.dirname(os.path.dirname(__file__))
_logger = logging.getLogger(__name__)
```

victoralmau (changes requested):
My new suggestions help to put travis and runbot green.

Too, it's impotant to add to this file: https://github.com/OCA/dms/blob/13.0/requirements.txt this line:
`bsdiff4`

keshrath:
> My new suggestions help to put travis and runbot green.
> 
> Too, it's impotant to add to this file: https://github.com/OCA/dms/blob/13.0/requirements.txt this line:
> `bsdiff4`

Thanks for the review and the changes. Travis now has only one failed test case, but this one is a bit strange. On the one hand it is one of the tests in the basic DMS module and then the stacktrace is not very informative.

pedrobaeza:
I think the problem is in the test structure: as you are inheriting from other test file, both tests are performed: old and new. You need to change the `dms` test structure, having 2 test classes: one base, with only initialization and helper methods, and a second one inheriting the first that contains the `test_` methods. Example:

Base: https://github.com/OCA/l10n-spain/blob/2d911fed4742561171fcd45e10e0e70d63f6b5c9/l10n_es_aeat_mod303/tests/test_l10n_es_aeat_mod303.py#L15

Extension class: https://github.com/OCA/l10n-spain/blob/2d911fed4742561171fcd45e10e0e70d63f6b5c9/l10n_es_aeat_mod303/tests/test_l10n_es_aeat_mod303.py#L445

victoralmau:
Superseded by https://github.com/OCA/dms/pull/84

PR #29, opened 2020-10-19
[12.0][FIX] dms: /my/dms change to user and separate some urls in some functions

victoralmau:
PR in V13: https://github.com/OCA/dms/pull/30

Separate urls in some functions

- "my/dms" changes to auth=user
- "my/dms/directory/<int:dms_directory_id>" optimized and simplified.

Please @pedrobaeza and @etobella  review it

@Tecnativa TT26319

PR #30, opened 2020-10-19
[13.0][FIX] dms: /my/dms change to user and separate some urls in some functions

victoralmau:
PR in V12: https://github.com/OCA/dms/pull/29

Separate urls in some functions

- "my/dms/" changes to auth=user
- "my/dms/directory/<int:dms_directory_id>" optimized and simplified.

Please @pedrobaeza and @etobella  review it

@Tecnativa TT26319

### DIFF
--- a/dms/controllers/portal.py
+++ b/dms/controllers/portal.py
@@ -38,12 +38,69 @@ class CustomerPortal(CustomerPortal):
         return values
 
     @http.route(
-        ["/my/dms/", "/my/dms/directory/<int:dms_directory_id>"],
+        ["/my/dms"], type="http", auth="user", website=True,
+    )
+    def portal_my_dms(
+        self, sortby=None, filterby=None, search=None, search_in="name", **kw
+    ):
+        values = self._prepare_portal_layout_values()
+        searchbar_sortings = {"name": {"label": _("Name"), "order": "name asc"}}
+        # default sortby br
+        if not sortby:
+            sortby = "name"
+        sort_br = searchbar_sortings[sortby]["order"]
+        # search
+        searchbar_inputs = {
+            "name": {"input": "name", "label": _("Name")},
+        }
+        if not filterby:
+            filterby = "name"
+        # domain
+        domain = [
+            (
+                "id",
+                "in",
+                request.env["dms.directory"]._get_own_root_directories(
+                    request.env.user.id
+                ),
+            )
+        ]
+        # search
+        if search and search_in:
+            search_domain = []
+            if search_in == "name":
+                search_domain = OR([search_domain, [("name", "ilike", search)]])
+            domain += search_domain
+        # content according to pager and archive selected
+        items = (
+            request.env["dms.directory"]
+            .with_user(request.env.user.id)
+            .search(domain, order=sort_br)
+        )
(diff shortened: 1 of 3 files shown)


dms/controllers/portal.py
```diff
@@ -38,12 +38,69 @@ def _prepare_portal_layout_values(self):
...
+        items = (
+            request.env["dms.directory"]
+            .with_user(request.env.user.id)
+            .search(domain, order=sort_br)
+        )
+        request.session["my_dms_folder_history"] = items.ids
+        # values
+        values.update(
+            {
+                "dms_directories": items,
```
pedrobaeza: In v12 this is:

```suggestion
                "dms_directories": items.sudo(),
```

dms/controllers/portal.py
```diff
@@ -38,12 +38,69 @@ def _prepare_portal_layout_values(self):
...
+                "page_name": "dms_directory",
+                "default_url": "/my/dms/directories",
```
pedrobaeza: In v12 this is:
```suggestion
                "default_url": "/my/dms",
```

dms/controllers/portal.py
```diff
@@ -75,85 +131,64 @@ def portal_my_dms(
...
             search_domain = []
             if search_in == "name":
                 search_domain = OR([search_domain, [("name", "ilike", search)]])
             domain += search_domain
         # content according to pager and archive selected
-        items = (
+        dms_directory_items = (
             request.env["dms.directory"]
-            .with_user(request.env.user.id)
+            .sudo(request.env.user.id)
```
pedrobaeza:
```suggestion
            .with_user(request.env.user.id)
```

dms/controllers/portal.py
```diff
@@ -75,85 +131,64 @@ def portal_my_dms(
...
-            )
+        request.session["my_dms_file_history"] = dms_file_items.ids
+        dms_parent_categories = dms_directory_sudo.sudo(
+            request.env.user.id
+        )._get_parent_categories(access_token)
+        # values
+        values = {
+            "dms_directories": dms_directory_items.sudo(),
+            "page_name": "dms_directory",
+            "default_url": "/my/dms/directories",
```
pedrobaeza:
```suggestion
            "default_url": "/my/dms",
```

pedrobaeza (changes requested):

dms/controllers/portal.py
victoralmau: In V13 is needed use `with_user(id)` instead of `sudo(id)`.

dms/controllers/portal.py
victoralmau: In V12 fix it isn't change.

dms/controllers/portal.py
pedrobaeza: Exactly, that's why I'm requesting this change...

dms/controllers/portal.py
pedrobaeza: Yeah, I know. It's not critical and I have seen it later, but better to put a valid URL in this version.

pedrobaeza (approved):

pedrobaeza:
Merging as well:

/ocabot merge patch

PR #33, opened 2020-10-20
[12.0][FIX] dms: Fix mimetype error in some files (.xlsx for example)

victoralmau:
In some case, if upload some extension type (.xlsx for example), in some operations (when update "_Inherit groups_" directory) error appear related to not defined mimetype.

Simplifying `_compute_mimetype` function is the option to solve it.

Related PR (V13): https://github.com/OCA/dms/pull/35

Please @pedrobaeza review it

@Tecnativa TT26273

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -6,7 +6,6 @@ import base64
 import hashlib
 import json
 import logging
-import mimetypes
 from collections import defaultdict
 
 from odoo import SUPERUSER_ID, _, api, fields, models, tools
@@ -387,11 +386,10 @@ class File(models.Model):
     @api.depends("name", "content")
     def _compute_mimetype(self):
         for record in self:
-            mimetype = record.name and mimetypes.guess_type(record.name)[0]
-            if not mimetype and record.content:
-                binary = base64.b64decode(record.with_context({}).content or "")
-                mimetype = guess_mimetype(binary, default="application/octet-stream")
-            record.res_mimetype = mimetype
+            binary = base64.b64decode(record.with_context({}).content or "")
+            record.res_mimetype = guess_mimetype(
+                binary, default="application/octet-stream"
+            )
 
     @api.depends("content_binary", "content_file")
     def _compute_content(self):


pedrobaeza:
This simplification means losing of possibilities? If `guess_mimetype` using as well the `mimetypes` library for trying to detect the mimetype?

victoralmau:
> This simplification means losing of possibilities? If `guess_mimetype` using as well the `mimetypes` library for trying to detect the mimetype?

This is my solution but probably exists other much better.
Looking odoo code about `ir.attachment` i found this:
https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/ir_attachment.py#L250

`guess_mimetype` function DON'T use `mimetypes` (https://github.com/odoo/odoo/blob/12.0/odoo/tools/mimetypes.py#L135)

Maybe is better solution to adapt code to make similar, what do you think?

pedrobaeza:
@keshrath do you know the reason for using `mimetypes` library?

@victoralmau any way, can't the code be enclosed in a try and fallback to the other when failing?

dms/models/dms_file.py
```diff
@@ -387,11 +386,10 @@ def _compute_extension(self):
...
         for record in self:
-            mimetype = record.name and mimetypes.guess_type(record.name)[0]
-            if not mimetype and record.content:
-                binary = base64.b64decode(record.with_context({}).content or "")
-                mimetype = guess_mimetype(binary, default="application/octet-stream")
-            record.res_mimetype = mimetype
+            binary = base64.b64decode(record.with_context({}).content or "")
+            record.res_mimetype = guess_mimetype(
+                binary, default="application/octet-stream"
+            )
```
victoralmau, line 392: Perhaps another possible solution (taking as a reference what odoo does in `ir.attachment`) is something similar to this:

```suggestion
            mimetype = record.name and mimetypes.guess_type(record.name)[0]
            if not mimetype or mimetype == 'application/octet-stream':
                mimetype = guess_mimetype(base64.b64decode(record.content))                
            record.res_mimetype = mimetype
```

victoralmau:
> @victoralmau any way, can't the code be enclosed in a try and fallback to the other when failing?

There is really no error in the code, and showing logs always returns a mimetype (which solved very strange) and that is why I have simplified the function.

pedrobaeza:
Let's wait then for @keshrath's answer.

etobella (approved):
LGTM

pedrobaeza:
@keshrath can you please comment about this?

etobella:
@keshrath

pedrobaeza:
Let's merge it for now for avoiding the problem while we receive the answer:

/ocabot merge patch

@victoralmau I understand that this is also needed in v13, isn't it?

pedrobaeza:
@Yajo @sbidoul we have in this branch pre-commit. Can you tell me what are the changes to make it work now with the change of prettier repo?

sbidoul:
@pedrobaeza 

```
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'checkout', '1.19.1')
return code: 1
expected return code: 0
stdout: (none)
stderr:
    error: pathspec '1.19.1' did not match any file(s) known to git
```

That's because that 1.19.1 tag does not exist in prettier/pre-commit. v1.19.1 should work.

victoralmau:
I have created new PR related it in V13: https://github.com/OCA/dms/pull/35

marcelsavegnago (approved):

PR #46, opened 2021-01-18
[12.0][FIX] dms: Remove request website user try

### LINKED ISSUE #42: [12.0] Current user id overwritten by website user (id=4)
Hi there,
i've just installed DMS module and everything works properly at first. But I found something in the code that i cannot understand:
https://github.com/OCA/dms/blob/3a6380e7e93d9365b177c77fa5229f7114777e33/dms/controllers/portal.py#L117
https://github.com/OCA/dms/blob/3a6380e7e93d9365b177c77fa5229f7114777e33/dms/controllers/portal.py#L205

Why current user id (logged user) is overwrite by website user? I mean, whenever I try to access a directory, it throws me an error because user id (4) doesn't have permissions.
As a workaround, the solution is simple:

image

But I would like to know the purpose of that.
Thanks in advance.

victoralmau:
Remove request website user try (it's not necessary).

Fixes #42 

Locked by:

- [x] https://github.com/OCA/dms/pull/47

@Tecnativa TT27782

### DIFF
--- a/dms/controllers/portal.py
+++ b/dms/controllers/portal.py
@@ -110,14 +110,7 @@ class CustomerPortal(CustomerPortal):
         access_token=None,
         **kw
     ):
-        """Process user's consent acceptance or rejection."""
         ensure_db()
-        try:
-            # If there's a website, we need a user to render the template
-            request.uid = request.website.user_id.id
-        except AttributeError:
-            # If there's no website, the default is OK
-            pass
         # operations
         searchbar_sortings = {"name": {"label": _("Name"), "order": "name asc"}}
         # default sortby br


etobella:
pre-commit is not right :wink:

pedrobaeza:
@etobella see #47

pedrobaeza:
@victoralmau why are you removing that code for being used when `website` is installed?

victoralmau:
It's not really necessary, if some user try to access some directory or file without permission auto-redirect to home. 
No other addon use it.

pedrobaeza (approved):

PR #47, opened 2021-01-18
Update configuration files

joao-p-marques:
Pre-commit needed update because it was causing pip install errors (see https://travis-ci.org/github/OCA/dms/jobs/754985038 for example)

This applies the same configurations as defined in https://github.com/OCA/oca-addons-repo-template/ (without actually bootstraping the copier template, as it is not available for v12.0). It also ignores configurations that would introduce unwanted changes (like migrating from Travis to GH Actions)

The changes include updates or creation of other files that the template defines.

ping @victoralmau @pedrobaeza @Yajo


yajo:
I'm not sure we want this kind of entropy in a `12.0` branch... 🤔 

It seems to me that it would be better to just obey linter rules manually in this repo as all other v12 ones do.

joao-p-marques:
Ok, but the problem is that the current pre-commit configuration is outdated and causing problems when installing because of the new pip dependency resolver, so we need to update the config anyway, right?

pedrobaeza (approved):
For me it's OK. @Yajo take into account that this 12.0 branch was created to be like 13.0, as it was created later and coming from 13.0, so for not having to undo the pre-commit style/line length.

pedrobaeza:
Well, when Travis is green...

etobella (approved):

etobella:
I made a PR fixing travis #48 but I closed it as this is older, can you apply the same change, that will fix Travis and then we can continue or  do you prefer to close this and continue with my PR?

joao-p-marques:
From what I can see this would also fix Travis, updating pre-commit

etobella:
No, it is not fixing it, as you can see on the results of the PR

joao-p-marques:
Sorry, MB. The Travis results indicated linting errors, and not an installation error with pip, which was the previous error. The 2nd commit should fix those.

The only difference between your PR and this with regards to the `pre-commit` config is that you bump the version of `setuptools-odoo` to **2.6.0** and this keeps it at **2.5.2**. As that is the version that comes from the repos template, I'd suggest to keep the lower version. WDYT?

etobella:
Sure

PR #51, opened 2021-02-01
[12.0][FIX] dms: Fix error when  web_drop_target addon used to create new file + remove content field in tree

victoralmau:
1- Fix error when  `web_drop_target` addon used to create new file

dms-error-web_drop_target

2- Remove `content` field in tree view to prevent download all files.

This error also happens in 13.0

Please @pedrobaeza , @joao-p-marques  and @Yajo review it.

@Tecnativa TT28040

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -644,7 +644,12 @@ class File(models.Model):
 
     def _create_model_attachment(self, vals):
         res_vals = vals.copy()
-        directory = self.env["dms.directory"].sudo().browse(res_vals["directory_id"])
+        directory = False
+        directory_model = self.env["dms.directory"]
+        if "directory_id" in res_vals:
+            directory = directory_model.browse(res_vals["directory_id"])
+        elif self.env.context.get("active_id"):
+            directory = directory_model.browse(self.env.context.get("active_id"))
         if directory and directory.res_model and directory.res_id:
             attachment = (
                 self.env["ir.attachment"]
--- a/dms/views/dms_file.xml
+++ b/dms/views/dms_file.xml
@@ -298,7 +298,7 @@
                 <field name="is_lock_editor" invisible="1" />
                 <field name="name" />
                 <field name="write_date" />
-                <field name="content" />
+                <field name="size"/>
                 <field name="res_mimetype" />
                 <field name="path_names" widget="path_names" string="Path" />
             </tree>


etobella (approved):

dms/models/dms_file.py
```diff
@@ -644,7 +644,12 @@ def _inverse_content(self):
 
     def _create_model_attachment(self, vals):
         res_vals = vals.copy()
-        directory = self.env["dms.directory"].sudo().browse(res_vals["directory_id"])
+        directory = False
+        directory_model = self.env["dms.directory"].sudo()
+        if "directory_id" in res_vals:
+            directory = directory_model.browse(res_vals["directory_id"])
+        elif self.env.context.get("active_id"):
+            directory = directory_model.browse(self.env.context.get("active_id"))
```
joao-p-marques, line 652: Perhaps you could save `self.env.context.get("active_id")` in a variable to avoid having to fetch it twice?
```suggestion
        active_id_from_ctx = self.env.context.get("active_id")
        if "directory_id" in res_vals:
            directory = directory_model.browse(res_vals["directory_id"])
        elif active_id_from_ctx:
            directory = directory_model.browse(active_id_from_ctx)
```

joao-p-marques (approved):
LGTM :+1: 

Just a small suggestion

dms/models/dms_file.py
yajo: This is not bad per se, but it's not gonna give you any more performance. 🤷🏼‍♂️

yajo (approved):

dms/models/dms_file.py
```diff
@@ -644,7 +644,12 @@ def _inverse_content(self):
 
     def _create_model_attachment(self, vals):
         res_vals = vals.copy()
-        directory = self.env["dms.directory"].sudo().browse(res_vals["directory_id"])
+        directory = False
+        directory_model = self.env["dms.directory"].sudo()
```
yajo: I'm not sure sudo is a good idea here. You don't want your files to be uploaded to folders where you don't have permissions, right? 🤔

yajo (changes requested):

dms/models/dms_file.py
joao-p-marques: Yes, of course. It is just a small suggestion, not really important :smile:

dms/models/dms_file.py
pedrobaeza: There was previously a `sudo`, maybe for the attachment thing. Although in general is not good idea to require `sudo` (and more in v12 that changes the uid), do you think this removal can provoke any error?

marcelsavegnago (approved):

dms/models/dms_file.py
yajo: I tested functionally after this removal and it seems to work fine 👍🏼

yajo (approved):

pedrobaeza:
@Yajo have you tried configuring a directory with attachment auto-creation? @victoralmau can you answer my previous question?

yajo:
I just tried with all 3 types of storage provided. All seem to work fine.

Uploads bigger than 25MB work also, when using drag and drop. I think the form's UI should have some kind of warning about that problem, reminding the user that he must upload with drag & drop for big files.

victoralmau:
I have tried attachment storage process and work fine.

dms-attachment_storage

pedrobaeza:
/ocabot merge patch

Is it needed in v13?

dms/models/dms_file.py
kos94ok-3D: Is more readable?
<pre>
directory_id = res_vals.get("directory_id", self._context.get("active_id"))
directory = directory_id and self.env["dms.directory"].browse(directory_id)
</pre>

PR #52, opened 2021-02-02
[13.0][FIX] dms: Fix error when web_drop_target addon used to create new file + remove content field in tree

victoralmau:
Related to 12.0: https://github.com/OCA/dms/pull/51

1- Fix error when  `web_drop_target` addon used to create new file

dms-error-web_drop_target

2- Remove `content` field in tree view to prevent download all files.

@Tecnativa TT28040

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -638,7 +638,12 @@ class File(models.Model):
 
     def _create_model_attachment(self, vals):
         res_vals = vals.copy()
-        directory = self.env["dms.directory"].sudo().browse(res_vals["directory_id"])
+        directory = False
+        directory_model = self.env["dms.directory"]
+        if "directory_id" in res_vals:
+            directory = directory_model.browse(res_vals["directory_id"])
+        elif self.env.context.get("active_id"):
+            directory = directory_model.browse(self.env.context.get("active_id"))
         if directory and directory.res_model and directory.res_id:
             attachment = (
                 self.env["ir.attachment"]
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -638,13 +638,9 @@ class File(models.Model):
 
     def _create_model_attachment(self, vals):
         res_vals = vals.copy()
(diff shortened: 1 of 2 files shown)


dms/models/dms_file.py
```diff
@@ -638,7 +638,12 @@ def _inverse_content(self):
 
     def _create_model_attachment(self, vals):
         res_vals = vals.copy()
-        directory = self.env["dms.directory"].sudo().browse(res_vals["directory_id"])
+        directory = False
+        directory_model = self.env["dms.directory"]
+        if "directory_id" in res_vals:
+            directory = directory_model.browse(res_vals["directory_id"])
+        elif self.env.context.get("active_id"):
+            directory = directory_model.browse(self.env.context.get("active_id"))
```
kos94ok-3D: Is more readable?
<pre>
directory_id = res_vals.get("directory_id", self.env.context.get("active_id"))
directory = directory_id and self.env["dms.directory"].browse(directory_id)
</pre>

etobella (approved):

dms/models/dms_file.py
```diff
@@ -638,7 +638,8 @@ def _inverse_content(self):
 
     def _create_model_attachment(self, vals):
         res_vals = vals.copy()
-        directory = self.env["dms.directory"].sudo().browse(res_vals["directory_id"])
+        directory_id = res_vals.get("directory_id", self.env.context.get("active_id"))
+        directory = directory_id and self.env["dms.directory"].browse(directory_id)
         if directory and directory.res_model and directory.res_id:
```
yajo: You can safely do this:

```suggestion
        directory = self.env["dms.directory"].browse(directory_id)
        if directory.res_model and directory.res_id:
```

It is safe to do it, because in case `directory_id` is `False` or `None`, then `.browse()` will return an empty recordset, which itself returns always falsy values when accessing its fields.

yajo (approved):

yajo (approved):

PR #53, opened 2021-02-05
[12.0][FIX] dms: Use sudo() in attachment operacion to prevent error in attachment creation

victoralmau:
**What happen?**
Permissions error because to user haven't access to dms models.

**When?**
If user without dms permissions try to create a attachment in any model. For example, try to email send with attachment(s).

**Why don't appear it until now?**
Probably because all people that use dms have dms permissions (user for example) or haven't create new attachment yet with users without dms permissions.

**Why?**
In attachment creation dms https://github.com/OCA/dms/blob/12.0/dms/models/ir_attachment.py#L61-L66 make some operations related to: search/create dms directories and search/create dms files (create if it's necessary according to dms configuration).

**But, why these functionality?**
DMS has the functionality of creating dms files and folders as attachments are created.
If a storage configured as attachment and related to a model is created (partner for example), a root directory is created in that storage linked to the partner model, when an attachment is created in a partner, it will be auto-created (if it does not exist ) the corresponding partner folder, and the dms file of that attachment will also be created in that folder. 

**And, it's really necessary to use `sudo()`?**
IMO yes, because the user who creates the attached file may not have the dms permissions and even if they did, they may not have access to the directories in which the dms file must be created (according to what is defined in dms access group). 

@Tecnativa TT28162

### DIFF
--- a/dms/models/ir_attachment.py
+++ b/dms/models/ir_attachment.py
@@ -7,8 +7,10 @@ class IrAttachment(models.Model):
     _inherit = "ir.attachment"
 
     def _get_dms_directories(self, res_model, res_id):
-        return self.env["dms.directory"].search(
-            [("res_model", "=", res_model), ("res_id", "=", res_id)]
+        return (
+            self.env["dms.directory"]
+            .sudo()
+            .search([("res_model", "=", res_model), ("res_id", "=", res_id)])
         )
 
     def _dms_directories_create(self):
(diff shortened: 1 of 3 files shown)


yajo (approved):

yajo:
I think the problem this patch fixes is not very obvious. Could you add a test?

pedrobaeza:
Yes, please explain the steps to reproduce the error, and I don't quite really like it, as it's to put `sudo` everywhere for doing everything with all the permissions, but that's not the best approach:

- Reduce the scope of the sudo to the minimum.
- Why the sudo is needed? Isn't this reflecting the problem in other part of the code?

victoralmau:
I have updated my first comment to explain everything.

victoralmau:
Added tests.

pedrobaeza (approved):

PR #59, opened 2021-02-15
[13.0][FIX] dms: Use sudo() in attachment operacion to prevent error in attachment creation

victoralmau:
Related to 12.0: https://github.com/OCA/dms/pull/53

**What happen?**
Permissions error because to user haven't access to dms models.

**When?**
If user without dms permissions try to create a attachment in any model. For example, try to email send with attachment(s).

**Why don't appear it until now?**
Probably because all people that use dms have dms permissions (user for example) or haven't create new attachment yet with users without dms permissions.

**Why?**
In attachment creation dms https://github.com/OCA/dms/blob/13.0/dms/models/ir_attachment.py#L57-L62 make some operations related to: search/create dms directories and search/create dms files (create if it's necessary according to dms configuration).

**But, why these functionality?**
DMS has the functionality of creating dms files and folders as attachments are created.
If a storage configured as attachment and related to a model is created (partner for example), a root directory is created in that storage linked to the partner model, when an attachment is created in a partner, it will be auto-created (if it does not exist ) the corresponding partner folder, and the dms file of that attachment will also be created in that folder. 

**And, it's really necessary to use `sudo()`?**
IMO yes, because the user who creates the attached file may not have the dms permissions and even if they did, they may not have access to the directories in which the dms file must be created (according to what is defined in dms access group). 

@Tecnativa TT28162

### DIFF
--- a/dms/models/ir_attachment.py
+++ b/dms/models/ir_attachment.py
@@ -1,3 +1,4 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 from odoo import api, models
 
@@ -12,13 +13,13 @@ class IrAttachment(models.Model):
         )
 
     def _dms_directories_create(self):
-        items = self._get_dms_directories(self.res_model, False)
+        items = self.sudo()._get_dms_directories(self.res_model, False)
         for item in items:
             model_item = self.env[self.res_model].sudo().browse(self.res_id)
             ir_model_item = self.env["ir.model"].search(
                 [("model", "=", self.res_model)]
             )
-            self.env["dms.directory"].create(
+            self.env["dms.directory"].sudo().create(
                 {
                     "name": model_item.display_name,
                     "model_id": ir_model_item.id,
@@ -33,18 +34,18 @@ class IrAttachment(models.Model):
         for attachment in self:
             if not attachment.res_model or not attachment.res_id:
                 continue
-            directories = self._get_dms_directories(
+            directories = self.sudo()._get_dms_directories(
                 attachment.res_model, attachment.res_id
             )
             if not directories:
                 attachment._dms_directories_create()
                 # Get dms_directories again (with items previously created)
-                directories = self._get_dms_directories(
+                directories = self.sudo()._get_dms_directories(
                     attachment.res_model, attachment.res_id
                 )
             # Auto-create_files
             for directory in directories:
-                self.env["dms.file"].create(
+                self.env["dms.file"].sudo().create(
                     {
                         "name": attachment.name,
                         "directory_id": directory.id,
--- a/dms/tests/__init__.py
+++ b/dms/tests/__init__.py
@@ -6,4 +6,5 @@ from . import test_file_database
 from . import test_file
 from . import test_benchmark
 from . import test_directory_mail
+from . import test_ir_attachment
 from . import test_portal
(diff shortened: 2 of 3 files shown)


pedrobaeza (approved):

dms/models/ir_attachment.py
```diff
@@ -7,8 +8,10 @@ class IrAttachment(models.Model):
     _inherit = "ir.attachment"
 
     def _get_dms_directories(self, res_model, res_id):
-        return self.env["dms.directory"].search(
-            [("res_model", "=", res_model), ("res_id", "=", res_id)]
+        return (
+            self.env["dms.directory"]
+            .sudo()
```
yajo: I don't think this is a good idea.

The method itself is not clear about it's doing this. I mean, you might be using in your module:

```py
self._get_dms_directories('res.partner', [1,2,3]).unlink()
```

... and expect to unlink only those directories that your current user has access to. The method is called `_get_dms_directories`, not `_get_dms_directories_sudo`, so the fact that the returned recordset is sudoed is, IMHO, an unexpected side effect and a gateway for security problems.

The correct fix would be, in this case, to prepend the `sudo()` _to the method call_:

```py
self.sudo()._get_dms_directories('res.partner', [1,2,3]).unlink()
```

In this case, all is predictable and expected.

yajo (changes requested):

dms/models/ir_attachment.py
pedrobaeza: You approved the same in #53 ...

pedrobaeza: I don't think it's too much problem, as it's an internal method and it wasn't renamed for not losing compatibility with inherited modules.

yajo: Well, I didn't realize before. I still think this is a problem.

pedrobaeza: Your proposal is not correct. In any case, it would be:

```python
self._get_dms_directories('res.partner', [1,2,3]).sudo().unlink()
```

but that won't get the same results, as we need that the search already contains all possible results.

yajo: AFAIK, my proposal is correct 🤷🏼‍♂️ 

The sudo property is saved in the env, which is later used by the method, so it gets sudoed already.

pedrobaeza: As a new handler is obtained with `self.env["dms.directory"]`, I would say the `sudo` is not kept, but anyway, please don't block this for this detail that doesn't impact and let's fix the underlying problem.

yajo: I don't consider this a detail, because it can lead to security problems. However, to avoid blocking, I pushed the fix. Sometimes, it's faster to do it than to talk about it. 😃

pedrobaeza:
@Yajo, what you have done is what we didn't want to do for having all around `sudo`. Take into account that this one is an internal method only used for this.

dms/models/ir_attachment.py
pedrobaeza: It's not a question of talking, but don't put an ugly code that we didn't want. My premise to Victor was: do the minimal `sudo` operations.

yajo: I'm adding no more sudo operations. I'm just being explicit about them.

pedrobaeza: 3 against 1. You are adding more indeed.

yajo: Those 3 were already being done inside the method, so there were 3 being done in any case. Readers just didn't realize.

pedrobaeza: OK, do both what you think. I retire, but I don't want 2 commits for such simple thing.

victoralmau:
Finally, what happen with these PR? 
It's important to add to solved it and add `test_ir_attachment` file.

pedrobaeza:
Although I don't agree with the spread of the `sudo`, keep it for not blocking the PR, but at least squash the commits together keeping if you want the co-authorship.

yajo (approved):

Issue #60, opened 2021-02-16
[13.0] Record Ref selection bug


ghost:
With attachment type storage, cannot open files and directories, because record_ref fields have "selection=[ ]".

In local i solved with a function like this:

    @api.model
    def _get_ref_selection(self):
        models = self.env['ir.model'].search([])
        return [(model.model, model.name) for model in models]
        
 Hope is useful

pedrobaeza:
But why do you need `record_ref`? I don't think it's a required field.

ghost:
Is a computed field:

```
    record_ref = fields.Reference(
            string="Record Referenced",
            compute="_compute_record_ref",
            selection=[],
            readonly=True,
            store=False,
        )
```

pedrobaeza:
But why is a problem then?

ghost:
Because, when you open a file or a directory gives an error and you can't open them

victoralmau:
Please, can you describe steps to reproduce in runbot?
I have try to open a directory or file in runbot 13.0 and don't get any problem.

ghost:
Yes, of course:

- Create a storage with "Save Type" = attachment, select a model from Linked Models (e.g res.partner)
- Create a directory from this storage and save it, edit it and set the model field (e.g res.partner)
- Now open a contact and upload a file from the chatter box as attachment
- Go to Documents and try to open that file, or the directory generated

victoralmau:
We can close this now, right?

PR #61, opened 2021-02-17
[13.0][REF+FIX] dms: Record_ref fixed when access to item (directory or file) in view

victoralmau:
Related to https://github.com/OCA/dms/issues/60

- Refactor code to create `abstract.dms.mixin` to contain fields or functions that is used in some models (in these case some fields appear in directory and field and some functions about it).
- Fixed problem happen when access to file or folder view if record ref is set (file or folder has been created in attachment creation in partner for example)

Although it's work fine in 12.0 IMO it's necessary to apply 12.0 to standardize code.

@Tecnativa TT28337

### DIFF
--- a/dms/models/abstract_dms_mixin.py
+++ b/dms/models/abstract_dms_mixin.py
@@ -1,0 +1,50 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class AbstractDmsMixin(models.AbstractModel):
+    _name = "abstract.dms.mixin"
+    _description = "Abstract Dms Mixin"
+
+    name = fields.Char(string="Name", required=True, index=True)
+    res_model = fields.Char(string="Linked attachments model")
+    res_id = fields.Integer(string="Linked attachments record ID")
+    record_ref = fields.Reference(
+        string="Record Referenced",
+        compute="_compute_record_ref",
+        selection=lambda self: self._get_ref_selection(),
+    )
+    # Only defined to prevent error in other fields that related it
+    storage_id = fields.Many2one(
+        comodel_name="dms.storage", string="Storage", store=True,
+    )
+    is_hidden = fields.Boolean(
+        string="Storage is Hidden", related="storage_id.is_hidden", readonly=True
+    )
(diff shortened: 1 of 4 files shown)


pedrobaeza:
I see that you are doing a refactoring at the same time. Please at least separate in 2 commits. I suppose the real patch is to put `False` instead of `None` and this highlights why is a bad idea to put `None` in Odoo fields.

victoralmau:
Sorry Pedro it's a confusion, i did not want to upload that change, it really is not necessary (change `None` to `False`).

pedrobaeza:
Then I don't get why this refactoring is fixing the problem...

dms/models/abstract_dms_mixin.py
```diff
@@ -0,0 +1,28 @@
...
+class AbstractDmsMixin(models.AbstractModel):
+    _name = "abstract.dms.mixin"
+    _description = "Abstract Dms Mixin"
+
+    res_model = fields.Char(string="Linked attachments model")
+    res_id = fields.Integer(string="Linked attachments record ID")
+    record_ref = fields.Reference(
+        string="Record Referenced",
+        compute="_compute_record_ref",
+        selection=lambda self: self._get_ref_selection(),
```
etobella, line 16: That is the real change that is fixing this. The refactoring is just made to minimize the amount of code

etobella (approved):
Code Review LGTM

dms/models/abstract_dms_mixin.py
pedrobaeza: OK, understood. Then the commit message should be `[REF+FIX] ` and explaining in the commit message the real fix and the refactoring.

victoralmau: Changed it now. 
Probably it's possible to optimize a lot of code that is used in some models (directory and field specially).

pedrobaeza:
OK, then take a look and do it now for not having several of these commits.

pedrobaeza:
@etobella do you agree with the new changes?

victoralmau:
Ping @etobella, do you agree with the new changes?

etobella:
Yes, good for me

Issue #62, opened 2021-02-18
[13.0] Exception while changing name or file content


LorisBalsamo:
When I try to change the name or the file content of a file in DMS, it shows this exception:

The problem exist in runbot, not only on my local instance.

`Odoo Server Error
Traceback (most recent call last):
  File "/.repo_requirements/odoo/odoo/http.py", line 624, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/.repo_requirements/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 669, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/odoo/http.py", line 350, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 915, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/.repo_requirements/odoo/addons/web/controllers/main.py", line 1327, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/.repo_requirements/odoo/addons/web/controllers/main.py", line 1319, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 387, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 374, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/models.py", line 6255, in onchange
    for name in nametree
  File "/.repo_requirements/odoo/odoo/models.py", line 6256, in <listcomp>
    if name not in done and snapshot0.has_changed(name)
  File "/.repo_requirements/odoo/odoo/models.py", line 6101, in has_changed
    return self[name] != record[name]
[...]

PR #64, opened 2021-02-18
[13.0][FIX] dms: Prevent error when change name or content

victoralmau:
Prevent error when change name or content.

Related to https://github.com/OCA/dms/issues/62

Please @pedrobaeza and @etobella review it.

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -408,13 +408,10 @@ class File(models.Model):
         for record in self:
             record.extension = file.guess_extension(record.name)
 
-    @api.depends("name", "content")
+    @api.depends("content")
     def _compute_mimetype(self):
         for record in self:
-            binary = base64.b64decode(record.with_context({}).content or "")
-            record.res_mimetype = guess_mimetype(
-                binary, default="application/octet-stream"
-            )
+            record.res_mimetype = guess_mimetype(record.content or b"")
 
     @api.depends("content_binary", "content_file", "attachment_id")
     def _compute_content(self):


dms/models/dms_file.py
```diff
@@ -408,10 +408,13 @@ def _compute_extension(self):
         for record in self:
             record.extension = file.guess_extension(record.name)
 
-    @api.depends("name", "content")
+    @api.depends("name", "content", "content_binary")
     def _compute_mimetype(self):
         for record in self:
-            binary = base64.b64decode(record.with_context({}).content or "")
+            if record.content_binary:
+                binary = record.content_binary
```
pedrobaeza: No decode needed here?

pedrobaeza:
I don't get the fix. `content` is a computed field that gets the proper value when binary:

https://github.com/OCA/dms/blob/b012c200e543a71a42feb071c7d5b39c7906bf78/dms/models/dms_file.py#L423

pedrobaeza:
Was this previously working with the decoding?

victoralmau:
This works fine in 12.0, probably in 13.0 nobody had reviewed it yet (create a file and edit the name or content).
The change I had previously proposed also worked because `content_binary` is an existing field although I think the current change is simpler.

pedrobaeza:
On new files work as well?

victoralmau:
Now work fine on new files and edit files to other content/name.

dms/models/dms_file.py
```diff
@@ -411,10 +411,10 @@ def _compute_extension(self):
...
     def _compute_mimetype(self):
         for record in self:
-            binary = base64.b64decode(record.with_context({}).content or "")
-            record.res_mimetype = guess_mimetype(
-                binary, default="application/octet-stream"
-            )
+            if record.content:
+                record.res_mimetype = guess_mimetype(
+                    record.content, default="application/octet-stream"
+                )
```
pedrobaeza: and the else part?

victoralmau: It's really necessary? It's a compute and content field is required.

pedrobaeza: I think so, unless a default value is provided in the field definition. If not, you will be hit by `Missing cache value` sooner or later.

victoralmau: Ok, change it. 
Remove default because is the same that in function https://github.com/odoo/odoo/blob/13.0/odoo/tools/mimetypes.py#L136

dms/models/dms_file.py
```diff
@@ -411,10 +411,9 @@ def _compute_extension(self):
     @api.depends("name", "content")
     def _compute_mimetype(self):
         for record in self:
-            binary = base64.b64decode(record.with_context({}).content or "")
-            record.res_mimetype = guess_mimetype(
-                binary, default="application/octet-stream"
-            )
+            record.res_mimetype = "application/octet-stream"
+            if record.content:
+                record.res_mimetype = guess_mimetype(record.content)
```
pedrobaeza: Why not doing?
```suggestion
            record.res_mimetype = guess_mimetype(record.content or b"")
```

pedrobaeza (approved):

dms/models/dms_file.py
```diff
@@ -411,10 +411,7 @@ def _compute_extension(self):
     @api.depends("name", "content")
     def _compute_mimetype(self):
         for record in self:
-            binary = base64.b64decode(record.with_context({}).content or "")
-            record.res_mimetype = guess_mimetype(
-                binary, default="application/octet-stream"
-            )
+            record.res_mimetype = guess_mimetype(record.content or b"")
```
etobella, line 414: Good for me, just a question, why the name on the depends?

etobella (approved):

dms/models/dms_file.py
victoralmau: In my opinion I think that `name` field is not necessary because it is not really used to generate the `res_mimetype` field BUT as it was already like this (`name` field as depends) in the code I have not removed it.
What do you think @pedrobaeza  about it?

pedrobaeza: Maybe it was a trick for getting that recomputed, but it shouldn't be needed.

PR #67, opened 2021-03-30
[13.0][FIX] dms: Prevent parent_id error in directories in some UI cases

victoralmau:
Prevent `parent_id` error in directories in some UI cases

dms-error

Changes:
- [FIX] dms: Prevent `parent_id` error in directories in some UI cases
- [FIX] dms: Change `allowed_model_ids` field in directory to related
- [FIX] dms: Change `res_model` + `res_id` fields in files to related

Steps to reproduce:

- Create root directory according to some storage.
- Enter in directory created and go to "Subdirectories"
- Create new directory (error appear).

@Tecnativa TT29019

### DIFF
--- a/dms/__manifest__.py
+++ b/dms/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Document Management System",
     "summary": """Document Management System for Odoo""",
-    "version": "13.0.3.2.5",
+    "version": "13.0.4.0.0",
     "category": "Document Management",
     "license": "LGPL-3",
     "website": "http://github.com/OCA/dms",
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -1,5 +1,6 @@
 # Copyright 2017-2019 MuK IT GmbH.
 # Copyright 2020 Creu Blanca
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import base64
@@ -9,7 +10,7 @@ import operator
 from collections import defaultdict
 
 from odoo import _, api, fields, models, tools
-from odoo.exceptions import AccessError, ValidationError
+from odoo.exceptions import AccessError, UserError, ValidationError
 
 from odoo.addons.http_routing.models.ir_http import slugify
 
@@ -48,23 +49,18 @@ class DmsDirectory(models.Model):
         parent inherits the settings form its parent.""",
     )
 
-    root_storage_id = fields.Many2one(
-        comodel_name="dms.storage",
-        string="Root Storage",
-        ondelete="restrict",
-        compute="_compute_directory_type",
-        store=True,
-        readonly=False,
-        copy=True,
-    )
     # Override acording to defined in AbstractDmsMixin
     storage_id = fields.Many2one(
-        compute="_compute_storage",
+        compute="_compute_storage_id",
+        compute_sudo=True,
+        readonly=False,
         comodel_name="dms.storage",
         string="Storage",
         ondelete="restrict",
         auto_join=True,
         store=True,
+        copy=True,
+        default=False,
     )
 
     parent_id = fields.Many2one(
@@ -76,9 +72,17 @@ class DmsDirectory(models.Model):
         index=True,
         store=True,
         readonly=False,
-        compute="_compute_directory_type",
+        compute="_compute_parent_id",
         copy=True,
+        default=lambda self: self._default_parent_id(),
     )
+
+    def _default_parent_id(self):
+        context = self.env.context
+        if context.get("active_model") == self._name and context.get("active_id"):
+            return context["active_id"]
+        else:
+            return False
 
     complete_name = fields.Char(
         "Complete Name", compute="_compute_complete_name", store=True
@@ -261,7 +265,7 @@ class DmsDirectory(models.Model):
         return res
 
     allowed_model_ids = fields.Many2many(
-        compute="_compute_allowed_model_ids", comodel_name="ir.model", store=False
+        related="storage_id.model_ids", comodel_name="ir.model",
     )
     model_id = fields.Many2one(
         comodel_name="ir.model",
@@ -271,16 +275,6 @@ class DmsDirectory(models.Model):
         string="Model",
         store=True,
     )
-    storage_id_save_type = fields.Selection(related="storage_id.save_type", store=False)
-
-    @api.depends("root_storage_id", "storage_id")
-    def _compute_allowed_model_ids(self):
-        for record in self:
-            record.allowed_model_ids = False
-            if record.root_storage_id and record.root_storage_id.model_ids:
-                record.allowed_model_ids = record.root_storage_id.model_ids.ids
-            elif record.storage_id and record.storage_id.model_ids:
-                record.allowed_model_ids = record.storage_id.model_ids.ids
 
     @api.depends("res_model")
     def _compute_model_id(self):
@@ -375,13 +369,14 @@ class DmsDirectory(models.Model):
             else:
                 category.complete_name = category.name
 
-    @api.depends("root_storage_id", "parent_id")
-    def _compute_storage(self):
+    @api.depends("parent_id")
+    def _compute_storage_id(self):
         for record in self:
-            if record.is_root_directory:
-                record.storage_id = record.root_storage_id
-            else:
+            if record.parent_id:
                 record.storage_id = record.parent_id.storage_id
+            else:
+                # HACK: Not needed in v14 due to odoo/odoo#64359
+                record.storage_id = record.storage_id
 
     @api.depends("user_star_ids")
     def _compute_starred(self):
@@ -469,12 +464,13 @@ class DmsDirectory(models.Model):
     # ----------------------------------------------------------
 
     @api.depends("is_root_directory")
-    def _compute_directory_type(self):
+    def _compute_parent_id(self):
         for record in self:
             if record.is_root_directory:
                 record.parent_id = None
             else:
-                record.root_storage_id = None
+                # HACK: Not needed in v14 due to odoo/odoo#64359
+                record.parent_id = record.parent_id
 
     @api.depends("category_id")
     def _compute_tags(self):
@@ -494,19 +490,21 @@ class DmsDirectory(models.Model):
             raise ValidationError(_("Error! You cannot create recursive directories."))
         return True
 
-    @api.constrains("is_root_directory", "root_storage_id", "parent_id")
+    @api.constrains("is_root_directory", "storage_id")
     def _check_directory_storage(self):
         for record in self:
-            if record.is_root_directory and not record.root_storage_id:
-                raise ValidationError(_("A root directory has to have a root storage."))
-            if not record.is_root_directory and not record.parent_id:
-                raise ValidationError(_("A directory has to have a parent directory."))
-            if record.parent_id and (
-                record.is_root_directory or record.root_storage_id
-            ):
+            if record.is_root_directory and not record.storage_id:
+                raise ValidationError(_("A root directory has to have a storage."))
+
+    @api.constrains("is_root_directory", "parent_id")
+    def _check_directory_parent(self):
+        for record in self:
+            if record.is_root_directory and record.parent_id:
                 raise ValidationError(
                     _("A directory can't be a root and have a parent directory.")
                 )
+            if not record.is_root_directory and not record.parent_id:
+                raise ValidationError(_("A directory has to have a parent directory."))
 
     @api.constrains("parent_id")
     def _check_directory_access(self):
@@ -525,7 +523,7 @@ class DmsDirectory(models.Model):
             if not check_name(record.name):
                 raise ValidationError(_("The directory name is invalid."))
             if record.is_root_directory:
-                childs = record.sudo().root_storage_id.root_directory_ids.name_get()
+                childs = record.sudo().storage_id.root_directory_ids.name_get()
             else:
                 childs = record.sudo().parent_id.child_directory_ids.name_get()
             if list(
@@ -556,14 +554,11 @@ class DmsDirectory(models.Model):
     def copy(self, default=None):
         self.ensure_one()
         default = dict(default or [])
-        if "root_storage_id" in default:
-            storage = self.env["dms.storage"].browse(default["root_storage_id"])
-            names = storage.sudo().root_directory_ids.mapped("name")
-        elif "parent_id" in default:
+        if "parent_id" in default:
             parent_directory = self.browse(default["parent_id"])
             names = parent_directory.sudo().child_directory_ids.mapped("name")
         elif self.is_root_directory:
-            names = self.sudo().root_storage_id.root_directory_ids.mapped("name")
+            names = self.sudo().storage_id.root_directory_ids.mapped("name")
         else:
             names = self.sudo().parent_id.child_directory_ids.mapped("name")
         default.update({"name": unique_name(self.name, names)})
@@ -622,8 +617,6 @@ class DmsDirectory(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         for vals in vals_list:
-            if vals.get("root_storage_id", False):
-                vals["storage_id"] = vals["root_storage_id"]
             if vals.get("parent_id", False):
                 parent = self.browse([vals["parent_id"]])
                 data = next(iter(parent.sudo().read(["storage_id"])), {})
@@ -631,6 +624,17 @@ class DmsDirectory(models.Model):
         return super().create(vals_list)
 
     def write(self, vals):
+        if vals.get("storage_id"):
+            for item in self:
+                if item.storage_id.id != vals["storage_id"]:
+                    raise UserError(_("It is not possible to change the storage."))
+        if vals.get("parent_id"):
+            parent = self.browse([vals["parent_id"]])
+            for item in self:
+                if item.parent_id.storage_id != parent.storage_id:
+                    raise UserError(
+                        _("It is not possible to change parent to other storage.")
+                    )
         # Groups part
         if any(key in vals for key in ["group_ids", "inherit_group_ids"]):
             with self.env.norecompute():
@@ -641,19 +645,6 @@ class DmsDirectory(models.Model):
             records.recompute()
         else:
             res = super().write(vals)
-
-        if self and any(
-            field for field in vals if field in ["root_storage_id", "parent_id"]
-        ):
-            records = self.sudo().search([("id", "child_of", self.ids)]) - self
-            if "root_storage_id" in vals:
-                records.write({"storage_id": vals["root_storage_id"]})
-            elif "parent_id" in vals:
-                parent = self.browse([vals["parent_id"]])
-                data = next(iter(parent.sudo().read(["storage_id"])), {})
-                records.write(
-                    {"storage_id": self._convert_to_write(data).get("storage_id")}
-                )
         return res
 
     def unlink(self):
--- a/dms/views/directory.xml
+++ b/dms/views/directory.xml
@@ -440,14 +440,8 @@
                             <field
                                 name="storage_id"
                                 options="{'no_open': True}"
-                                attrs="{'invisible':['|', ('is_root_directory','=',True), ('parent_id','=',False)]}"
-                            />
-                            <field
-                                name="root_storage_id"
-                                options="{'no_open': True}"
-                                readonly="1"
-                                invisible="context.get('hide_root_storage', False)"
-                                attrs="{'invisible':[('is_root_directory','=',False)], 'required':[('is_root_directory','=',True)]}"
+                                attrs="{'readonly':[('is_root_directory','=',False)], 'required':[('is_root_directory','=',True)]}"
+                                force_save="1"
                             />
                         </group>
                         <group>
@@ -625,10 +619,6 @@
             <field name="is_root_directory" position="attributes">
                 <attribute name="invisible">0</attribute>
             </field>
-            <field name="root_storage_id" position="attributes">
-                <attribute name="readonly">0</attribute>
-                <attribute name="options">{'no_quick_create': True}</attribute>
-            </field>
             <field name="storage_id" position="attributes">
                 <attribute name="options">{}</attribute>
             </field>
@@ -679,8 +669,10 @@
                     <group>
                         <field name="is_root_directory" readonly="1" />
                         <field
-                            name="root_storage_id"
+                            name="storage_id"
                             options="{'no_quick_create': True}"
+                            attrs="{'readonly':[('is_root_directory','=',False)], 'required':[('is_root_directory','=',True)]}"
+                            force_save="1"
                         />
                     </group>
                     <group>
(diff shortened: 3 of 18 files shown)


pedrobaeza:
Why are you changing again to onchange? Please leave the computed writable, but fix the problem using such mechanism (probably a missing default or initialization value). Also the steps to reproduce the problem are important.

victoralmau:
> Why are you changing again to onchange? Please leave the computed writable, but fix the problem using such mechanism (probably a missing default or initialization value). Also the steps to reproduce the problem are important.

IMO these field it's not necessary to define as compute because really not define correctly https://github.com/OCA/dms/blob/13.0/dms/models/directory.py#L471-L477
Related to these field and these "strange" work appear `parent_id` field that it's change in previous compute.
I think it's need to change to onchange because only in some cases it's necessary to define `root_storage_id` or `parent_id` to None but not always.
What do you think about it?

pedrobaeza:
A computed writable is a writable field that acts like a computed on onchanges, so it's correct as is. Just fix the problematic case and please see the training about computed writable fields: https://www.youtube.com/watch?v=lE4e4G45BHo

joao-p-marques (approved):

dms/models/directory.py
```diff
@@ -475,6 +475,8 @@ def _compute_directory_type(self):
                 record.parent_id = None
             else:
                 record.root_storage_id = None
+                if self.env.context.get("active_id"):
```
pedrobaeza: Very weak condition. Please check at least that `active_model` is matching

pedrobaeza:
Anyway, the problem I'm seeing is that the fields hasn't been properly converted in v13. Previously, all computed fields resets to default values when being recomputed. Now, this is not happening, and no default value is provided, so they can have an uncertain state if the conditions to fill data is not met. You should provide a sane default.

dms/models/directory.py
```diff
@@ -475,6 +475,14 @@ def _compute_directory_type(self):
                 record.parent_id = None
```
pedrobaeza, line 470: In this code branch, there's `root_storage_id` assignation. For what is this field useful? Shouldn't it be the storage of the folder?

dms/models/directory.py
```diff
@@ -475,6 +475,14 @@ def _compute_directory_type(self):
...
             else:
                 record.root_storage_id = None
+                if self.env.context.get(
+                    "active_model"
+                ) == self._name and self.env.context.get("active_id"):
+                    record.parent_id = self.env.context.get("active_id")
+                else:
+                    record.parent_id = self.env["dms.directory"].search(
```
pedrobaeza: Searching for the first directory is OK, being from any storage?

dms/models/directory.py
victoralmau: All directories have `parent_id` except if directory is root (`is_root_directory`).
`storage_id` field is compute writable according to 
```
@api.depends("root_storage_id", "parent_id")
    def _compute_storage(self):
        for record in self:
            if record.is_root_directory:
                record.storage_id = record.root_storage_id
            else:
                record.storage_id = record.parent_id.storage_id
```
and it's reason to set None to `root_storage_id` field (show in form only if `is_root_directory` is checked)

dms/models/directory.py
victoralmau: IMO yes because not set yet storage (storage auto-defined when `parent_id` is set).
According to create directories, exists 3 possibilities:
- Create manual directory from storage smartbutton called "_Directories_": In these case auto-set `is_root_directory` to True and `parent_id` is not necessary.
- Create manual directory from other directory smartbutton called "_Subdirectories_": In these case `active_id` auto-set these record to `parent_id`.
- Create manual directory: Auto-set first root directory search from any storage.

pedrobaeza:
Pending on the comments on v12 version. The ideal is to have uniform values, and not to make this `if` according other fields.

victoralmau:
> Pending on the comments on v12 version. The ideal is to have uniform values, and not to make this `if` according other fields.

According to https://github.com/OCA/dms/pull/68#discussion_r616393645 maybe it's a good idea to merge it to solve explained problem and after in other PR we can try to improve it (12.0 + 13.0).

pedrobaeza:
Well, the computed writable can only be put in this version, not in 12.0, so maybe it's a good idea to apply here the PoC.

dms/models/directory.py
```diff
@@ -48,22 +48,13 @@ class DmsDirectory(models.Model):
...
-        ondelete="restrict",
-        compute="_compute_directory_type",
-        store=True,
-        readonly=False,
-        copy=True,
-    )
     # Override acording to defined in AbstractDmsMixin
     storage_id = fields.Many2one(
         compute="_compute_storage",
+        compute_sudo=True,
```
pedrobaeza, line 55: I miss here `readonly=False`

dms/models/directory.py
pedrobaeza: The name of the compute should now be changed, isn't it?

dms/views/directory.xml
```diff
@@ -678,10 +668,7 @@
                 <group>
                     <group>
                         <field name="is_root_directory" readonly="1" />
-                        <field
-                            name="root_storage_id"
-                            options="{'no_quick_create': True}"
-                        />
+                        <field name="storage_id" options="{'no_quick_create': True}" />
```
pedrobaeza: This needs to be readonly=1 if parent_id and force_save=1

pedrobaeza:
Ah, you need a migration script for moving data from `root_storage_id` to `storage_id`.

victoralmau:
@pedrobaeza any idea about solve error in `test_unlink_directory`?

dms/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Document Management System",
     "summary": """Document Management System for Odoo""",
-    "version": "13.0.3.2.3",
+    "version": "13.0.4.0.0",
```
pedrobaeza, line 7: Why are you manually changing the version in the manifest?

pedrobaeza: OK, I see the migration script

dms/migrations/13.0.4/pre-migration.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
```
pedrobaeza: Incorrect folder name. It should `13.0.4.0.0`.

dms/migrations/13.0.4/pre-migration.py
```diff
@@ -0,0 +1,20 @@
...
+from openupgradelib import openupgrade
+
+
+def convert_root_storage_id_in_storage_id(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE dms_directory
+        SET storage_id = root_storage_id
+        WHERE root_storage_id IS NOT NULL
```
pedrobaeza: `AND parent_id IS NULL`?

dms/models/directory.py
```diff
@@ -48,23 +48,16 @@ class DmsDirectory(models.Model):
...
+        readonly=False,
```
pedrobaeza, line 56: `default=False` will be probably needed as well.

dms/models/directory.py
```diff
@@ -643,17 +630,17 @@ def write(self, vals):
...
-                records.write({"storage_id": vals["root_storage_id"]})
-            elif "parent_id" in vals:
+            if "parent_id" in vals:
                 parent = self.browse([vals["parent_id"]])
                 data = next(iter(parent.sudo().read(["storage_id"])), {})
                 records.write(
                     {"storage_id": self._convert_to_write(data).get("storage_id")}
                 )
+            elif "parent_id" not in vals and "storage_id" in vals:
+                records.write({"storage_id": vals["storage_id"]})
```
pedrobaeza: This is writing again the same thanon the `super`

dms/models/directory.py
```diff
@@ -230,13 +224,9 @@ def _get_parent_categories(self, access_token):
                 current_directory = current_directory.parent_id
         return directories[::-1]
 
-    def _get_own_root_directories(self, user_id):
+    def _get_own_root_directories(self):
```
pedrobaeza: Why introducing here this API change?

dms/models/directory.py
victoralmau: It's necessary to prevent error in `test_storage` test (apply `storage_id` to all child directories).

pedrobaeza: OK, I see, but this will trigger N * X writes if there are N levels of depth, and each level has X children. You should prevent this with a context.

dms/models/directory.py
```diff
@@ -494,19 +492,17 @@ def _check_directory_recursion(self):
             raise ValidationError(_("Error! You cannot create recursive directories."))
         return True
 
-    @api.constrains("is_root_directory", "root_storage_id", "parent_id")
+    @api.constrains("is_root_directory", "storage_id", "parent_id")
     def _check_directory_storage(self):
```
joao-p-marques, line 494: Shouldn't this checks be in separate constraints? They are testing different things, and maybe we should avoid calling this method too many times...

joao-p-marques:
@victoralmau you mention this happens in some UI cases, but could you add a test that tries to reproduce it?

victoralmau:
> @victoralmau you mention this happens in some UI cases, but could you add a test that tries to reproduce it?

IMO is not necessary to create test because now only exists `storage_id` and is not possible to other field change it.

dms/models/directory.py
```diff
@@ -48,30 +48,24 @@ class DmsDirectory(models.Model):
...
+        copy=True,
+        default=False,
     )
 
     parent_id = fields.Many2one(
         comodel_name="dms.directory",
         domain="[('permission_create', '=', True)]",
         string="Parent Directory",
-        ondelete="restrict",
+        ondelete="cascade",
```
pedrobaeza: Why changing this? It's very risky to allow by default to unlink recursively. I think that should be an option under Actions, with double confirmation.

dms/models/directory.py
```diff
@@ -474,7 +464,15 @@ def _compute_directory_type(self):
             if record.is_root_directory:
                 record.parent_id = None
             else:
-                record.root_storage_id = None
+                record.storage_id = None
+                if self.env.context.get(
+                    "active_model"
+                ) == self._name and self.env.context.get("active_id"):
+                    record.parent_id = self.env.context.get("active_id")
```
pedrobaeza: Why are you assigning `parent_id` in a method called `_compute_directory_type`, and mixing also `storage_id`?

dms/models/dms_security_mixin.py
```diff
@@ -237,6 +237,8 @@ def check_access_groups(self, operation):
             return None
         group_ids = set(self.ids) - set(self._get_ids_without_access_groups(operation))
         if group_ids:
+            # Hack to prevent error in some cases (unlink subdirectory) re-compute
+            self.complete_group_ids
```
pedrobaeza: Please explain the problem and why is this a hack. I think more than this triggers the recomputation of that, but this denotes a bug in the way things are computed.

dms/models/directory.py
victoralmau: Ok, I don't change it.

dms/models/directory.py
victoralmau: Ok, i have changed compute function name and remove related to `storage_id`.

dms/models/dms_security_mixin.py
victoralmau: Now, it's not necessary, sorry

dms/models/directory.py
```diff
@@ -375,12 +367,10 @@ def _compute_complete_name(self):
             else:
                 category.complete_name = category.name
 
-    @api.depends("root_storage_id", "parent_id")
+    @api.depends("parent_id")
     def _compute_storage(self):
```
pedrobaeza: Use the convention:
```suggestion
    def _compute_storage_id(self):
```

dms/models/directory.py
```diff
@@ -469,12 +459,19 @@ def _compute_groups(self):
...
 
     @api.depends("is_root_directory")
-    def _compute_directory_type(self):
+    def _compute_parent_id(self):
         for record in self:
             if record.is_root_directory:
                 record.parent_id = None
             else:
-                record.root_storage_id = None
+                if self.env.context.get(
```
pedrobaeza: This seems more a default value than something that should be put in a computed

dms/models/directory.py
```diff
@@ -469,12 +459,19 @@ def _compute_groups(self):
...
+                    "active_model"
+                ) == self._name and self.env.context.get("active_id"):
+                    record.parent_id = self.env.context.get("active_id")
+                else:
```
pedrobaeza: The same for this part.

dms/models/directory.py
```diff
@@ -643,17 +637,23 @@ def write(self, vals):
             res = super().write(vals)
 
         if self and any(
-            field for field in vals if field in ["root_storage_id", "parent_id"]
+            field for field in vals if field in ["storage_id", "parent_id"]
```
pedrobaeza: I still don't have this part clear, as it means you can change the storage in the parent directory, and be propagated to children, but no other action is taken, so the storage is inconsistent. Imagine you change from filestore to DB, but the files already contained are still in filestore with no transformation. Am I wrong?

Maybe we should just restrict the change of the storage.

dms/models/directory.py
```diff
@@ -76,10 +70,25 @@ class DmsDirectory(models.Model):
         index=True,
         store=True,
         readonly=False,
-        compute="_compute_directory_type",
         copy=True,
+        default=lambda self: self._default_parent_id(),
     )
 
+    def _default_parent_id(self):
+        if self._context.get("active_model") == self._name and self._context.get(
```
pedrobaeza: `self._context` is an internal variable you shouldn't use. Cache the value:
```python
context = self.env.context
```
and then use it across the code.

dms/models/directory.py
```diff
@@ -76,10 +70,25 @@ class DmsDirectory(models.Model):
...
+            "active_id"
+        ):
+            return self._context.get("active_id")
```
```suggestion
            return context["active_id"]
```

dms/models/directory.py
```diff
@@ -76,10 +70,25 @@ class DmsDirectory(models.Model):
...
+        else:
+            return self.env["dms.directory"].search(
```
pedrobaeza: You are returning a recordset here and an ID on the other branch

dms/models/directory.py
```diff
@@ -273,13 +282,11 @@ def check_access(self, operation, raise_exception=False):
...
 
-    @api.depends("root_storage_id", "storage_id")
+    @api.depends("storage_id")
     def _compute_allowed_model_ids(self):
         for record in self:
             record.allowed_model_ids = False
-            if record.root_storage_id and record.root_storage_id.model_ids:
-                record.allowed_model_ids = record.root_storage_id.model_ids.ids
-            elif record.storage_id and record.storage_id.model_ids:
+            if record.storage_id and record.storage_id.model_ids:
```
pedrobaeza: No need of this double condition:
```suggestion
            if record.storage_id.model_ids:
```
If `record.storage_id` is an empty recordset, any subfield with dot notation will be populated as empty.

pedrobaeza: Even more, why don't you assign directly without if?:
```python
        record.allowed_model_ids = record.storage_id.model_ids.ids
```

dms/models/directory.py
```diff
@@ -622,15 +618,23 @@ def _process_message(self, msg_dict, extra_values=False):
...
-            if vals.get("root_storage_id", False):
-                vals["storage_id"] = vals["root_storage_id"]
             if vals.get("parent_id", False):
                 parent = self.browse([vals["parent_id"]])
                 data = next(iter(parent.sudo().read(["storage_id"])), {})
                 vals["storage_id"] = self._convert_to_write(data).get("storage_id")
         return super().create(vals_list)
 
     def write(self, vals):
+        if "storage_id" in vals and vals["storage_id"]:
```
pedrobaeza: Simplify expressions whenever you are able to:
```suggestion
        if vals.get("storage_id"):
```

dms/models/directory.py
victoralmau: Correct, you should not be able to change the storage if it is of a different type.

dms/models/directory.py
```diff
@@ -622,15 +618,23 @@ def _process_message(self, msg_dict, extra_values=False):
...
+            storage = self.env["dms.storage"].browse(vals["storage_id"])
+            for item in self:
+                if item.storage_id == storage:
```
pedrobaeza: This `if` is not needed

dms/models/directory.py
victoralmau: I have changed now to default and onchange and i have removed compute.

dms/models/directory.py
```diff
@@ -622,15 +618,23 @@ def _process_message(self, msg_dict, extra_values=False):
...
+                    continue
+                if item.storage_id.save_type != storage.save_type:
```
pedrobaeza: I think we should forbid changing the storage at all, no matter the `save_type`, as we don't want to maintain the storage propagation code. Or there is a use case for allowing to change the storage?

dms/models/directory.py
pedrobaeza: Well, the compute can be correct for avoiding the onchange. My comment was about the default value. I think you can keep the compute for the other part. Or is there any problem on that?

victoralmau: Yes, if we keep compute function exists some possibilities: 
- Option A: Only compute values according to `is_root_directory=True`. In these case appear error describe in these PR.
- Option B: Compute values according to `is_root_directory=True` and in other cases (else) call to default function.
- Option C: Remove compute but NOT create onchange function. If we created directories (specially in tests) show constraint according to is set `parent_id` in root directories.

Do you think in other option?

dms/models/directory.py
victoralmau: I think that if you should allow changing the storage because for example you want to re-organize the folders (as long as the storage type is the same).

pedrobaeza: We should forbid the change of the storage at all, as having the same `save_type` doesn't mean that data are saved in the same place. Imagine a new addon for storing in `Google Drive`. The `save_type` will be the same, but the specific parameters of the storage will redirect to a different account/path. This is also happening with the attachments, changing the model on each specific storage.

victoralmau:
IMO it's ready to merge, i will not add more changes.

dms/models/directory.py
```diff
@@ -622,15 +620,18 @@ def _process_message(self, msg_dict, extra_values=False):
...
-                vals["storage_id"] = vals["root_storage_id"]
             if vals.get("parent_id", False):
                 parent = self.browse([vals["parent_id"]])
                 data = next(iter(parent.sudo().read(["storage_id"])), {})
                 vals["storage_id"] = self._convert_to_write(data).get("storage_id")
         return super().create(vals_list)
 
     def write(self, vals):
+        if vals.get("storage_id"):
+            storage = self.env["dms.storage"].browse(vals["storage_id"])
```
pedrobaeza: Don't browse, but compare with `.id` instead

dms/models/directory.py
```diff
@@ -643,17 +644,23 @@ def write(self, vals):
...
 
         if self and any(
-            field for field in vals if field in ["root_storage_id", "parent_id"]
+            field for field in vals if field in ["storage_id", "parent_id"]
         ):
             records = self.sudo().search([("id", "child_of", self.ids)]) - self
-            if "root_storage_id" in vals:
-                records.write({"storage_id": vals["root_storage_id"]})
-            elif "parent_id" in vals:
+            if "parent_id" in vals:
```
pedrobaeza: You should prevent the same to change the parent if the target is from another storage.

dms/models/directory.py
```diff
@@ -643,17 +644,23 @@ def write(self, vals):
...
                 records.write(
                     {"storage_id": self._convert_to_write(data).get("storage_id")}
                 )
+            elif (
+                "parent_id" not in vals
+                and "storage_id" in vals
+                and not self.env.context.get("bypass_update_childs_storage_id")
+            ):
+                records.with_context(bypass_update_childs_storage_id=True).write(
+                    {"storage_id": vals["storage_id"]}
```
pedrobaeza: This code has no sense now.

dms/models/directory.py
```diff
@@ -76,10 +71,22 @@ class DmsDirectory(models.Model):
...
-        compute="_compute_directory_type",
+        compute="_compute_parent_id",
         copy=True,
+        default=lambda self: self._default_parent_id(),
     )
 
+    def _default_parent_id(self):
+        context = self.env.context
+        if context.get("active_model") == self._name and context.get("active_id"):
+            return context.get["active_id"]
```
yajo: This is a bug.

```suggestion
            return context["active_id"]
```

dms/models/directory.py
```diff
@@ -76,10 +71,22 @@ class DmsDirectory(models.Model):
...
+        else:
+            return (
+                self.env["dms.directory"]
+                .search([("is_root_directory", "=", True)], limit=1)
+                .id
+            )
```
yajo: 🤔 Do you think this makes any sense?

I mean... **by default** every new directory should be created as a children of some random root directory? Wouldn't it make sense to leave it as `False` instead?

This is my suggestion:

```suggestion
        return False
```

---

If I'm right and this must be changed, please take that chance to put fields first and then, later, methods. At least with the new ones introduced. This follows Odoo guidelines (upon which OCA's are based): https://www.odoo.com/documentation/13.0/reference/guidelines.html#symbols-and-conventions

yajo (changes requested):

yajo:
> Steps to reproduce:
> 
>     * Create root directory according to some storage.
> 
>     * Enter in directory created and go to "Subdirectories"
> 
>     * Create new directory (error appear).

Important detail: I tested these STR and it worked 👍🏼

victoralmau:
It's ready to merge now, right?

dms/models/directory.py
```diff
@@ -48,23 +49,17 @@ class DmsDirectory(models.Model):
...
     # Override acording to defined in AbstractDmsMixin
     storage_id = fields.Many2one(
-        compute="_compute_storage",
+        compute="_compute_storage_id",
+        compute_sudo=True,
+        readonly=False,
         comodel_name="dms.storage",
         string="Storage",
         ondelete="restrict",
-        auto_join=True,
```
etobella, line 66: Why do you need to remove it?

pedrobaeza:
@victoralmau You should update the PR description as well to what this has finally been.

dms/models/directory.py
victoralmau: Do you remember these change @pedrobaeza ?
IMO these change is needed because we have added `default=False` right?

pedrobaeza: I think it's a mistake, as they are not related.

yajo (approved):
Code OK, runbot doesn't work.

dms/models/directory.py
etobella: So keep it please

etobella (approved):

PR #68, opened 2021-04-05
[12.0][IMP] dms: Inherit permissions from related records

victoralmau:
Add Inherit access from parent record (in storage).

@Tecnativa TT29041

### DIFF
--- a/dms/__manifest__.py
+++ b/dms/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Document Management System",
     "summary": """Document Management System for Odoo""",
-    "version": "12.0.4.0.8",
+    "version": "12.0.5.0.0",
     "category": "Document Management",
     "license": "LGPL-3",
     "website": "https://github.com/OCA/dms",
--- a/dms/controllers/portal.py
+++ b/dms/controllers/portal.py
@@ -16,24 +16,22 @@ _logger = logging.getLogger(__name__)
 class CustomerPortal(CustomerPortal):
     def _dms_check_access(self, model, res_id, access_token=None):
         try:
-            item_sudo = request.env[model].sudo().browse([res_id])
+            item = request.env[model].browse(res_id)
         except (AccessError, MissingError):
             return False
 
         if access_token:
-            if not item_sudo.check_access_token(access_token):
+            if not item.check_access_token(access_token):
                 return False
         else:
-            if not item_sudo.sudo(request.env.user.id).check_access("read", False):
+            if not item.sudo(request.env.user.id).check_access("read", False):
                 return False
 
-        return item_sudo
+        return item
 
     def _prepare_portal_layout_values(self):
         values = super()._prepare_portal_layout_values()
-        ids = request.env["dms.directory"]._get_own_root_directories(
-            request.env.user.id
-        )
+        ids = request.env["dms.directory"]._get_own_root_directories()
         values.update({"dms_directory_count": len(ids)})
         return values
 
@@ -57,13 +55,7 @@ class CustomerPortal(CustomerPortal):
             filterby = "name"
         # domain
         domain = [
-            (
-                "id",
-                "in",
-                request.env["dms.directory"]._get_own_root_directories(
-                    request.env.user.id
-                ),
-            )
+            ("id", "in", request.env["dms.directory"]._get_own_root_directories(),)
         ]
         # search
         if search and search_in:
@@ -81,7 +73,7 @@ class CustomerPortal(CustomerPortal):
         # values
         values.update(
             {
-                "dms_directories": items.sudo(),
+                "dms_directories": items,
                 "page_name": "dms_directory",
                 "default_url": "/my/dms",
                 "searchbar_sortings": searchbar_sortings,
@@ -169,7 +161,7 @@ class CustomerPortal(CustomerPortal):
         )._get_parent_categories(access_token)
         # values
         values = {
-            "dms_directories": dms_directory_items.sudo(),
+            "dms_directories": dms_directory_items,
             "page_name": "dms_directory",
             "default_url": "/my/dms/directories",
             "searchbar_sortings": searchbar_sortings,
@@ -179,7 +171,7 @@ class CustomerPortal(CustomerPortal):
             "filterby": filterby,
             "access_token": access_token,
             "dms_directory": dms_directory_sudo,
-            "dms_files": dms_file_items.sudo(),
+            "dms_files": dms_file_items,
             "dms_parent_categories": dms_parent_categories,
         }
         return request.render("dms.portal_my_dms", values)
@@ -193,12 +185,6 @@ class CustomerPortal(CustomerPortal):
     def portal_my_dms_file_download(self, dms_file_id, access_token=None, **kw):
         """Process user's consent acceptance or rejection."""
         ensure_db()
-        try:
-            # If there's a website, we need a user to render the template
-            request.uid = request.website.user_id.id
-        except AttributeError:
-            # If there's no website, the default is OK
-            pass
         # operations
         res = self._dms_check_access("dms.file", dms_file_id, access_token)
         if not res:
--- a/dms/demo/access_group.xml
+++ b/dms/demo/access_group.xml
@@ -12,8 +12,16 @@
         />
     </record>
     <record id="access_group_02_demo" model="dms.access.group">
-        <field name="name">Demo</field>
+        <field name="name">Portal</field>
         <field name="perm_read">True</field>
         <field name="group_ids" eval="[(6, 0, [ref('base.group_portal')])]" />
+    </record>
+    <record id="access_group_03_demo" model="dms.access.group">
+        <field name="name">Only admin user</field>
+        <field name="perm_read">True</field>
+        <field name="perm_create">True</field>
+        <field name="perm_write">True</field>
+        <field name="perm_unlink">True</field>
+        <field name="explicit_user_ids" eval="[(6, 0, [ref('base.user_admin')])]" />
     </record>
 </odoo>
--- a/dms/demo/storage.xml
+++ b/dms/demo/storage.xml
@@ -10,4 +10,11 @@
         <field name="name">Documents Storage</field>
         <field name="save_type">database</field>
     </record>
+    <record id="storage_attachment_demo" model="dms.storage">
+        <field name="name">Attachment Storage</field>
+        <field name="save_type">attachment</field>
+        <field name="inherit_access_from_parent_record" eval="True" />
+        <field name="include_message_attachments" eval="True" />
+        <field name="model_ids" eval="[(6, 0, [ref('base.model_res_partner')])]" />
+    </record>
 </odoo>
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -1,5 +1,6 @@
 # Copyright 2017-2019 MuK IT GmbH.
 # Copyright 2020 Creu Blanca
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import base64
@@ -194,6 +195,24 @@ class DmsDirectory(models.Model):
                 """,
     )
 
+    def check_access_rule(self, operation):
+        super().check_access_rule(operation)
+        for record in self:
+            if (
+                record.is_root_directory
+                and not record.root_storage_id_inherit_access_from_parent_record
+            ) or (
+                not record.is_root_directory
+                and not record.storage_id_inherit_access_from_parent_record
+            ):
+                record.check_access_groups(operation)
+                continue
+            # Check access to inherited model (and record)
+            model = self.env[record.res_model]
+            model.check_access_rights(operation)
+            if record.res_id:
+                model.browse(record.res_id).check_access_rule(operation)
+
     def _get_share_url(self, redirect=False, signup_partner=False, pid=None):
         self.ensure_one()
         return "/my/dms/directory/{}?access_token={}&db={}".format(
@@ -244,11 +263,9 @@ class DmsDirectory(models.Model):
                 current_directory = current_directory.parent_id
         return directories[::-1]
 
-    def _get_own_root_directories(self, user_id):
+    def _get_own_root_directories(self):
         ids = []
-        items = (
-            self.env["dms.directory"].sudo(user_id).search([("is_hidden", "=", False)])
-        )
+        items = self.env["dms.directory"].search([("is_hidden", "=", False)])
         for item in items:
             current_directory = item
             while (
@@ -261,16 +278,6 @@ class DmsDirectory(models.Model):
                 ids.append(current_directory.id)
 
         return ids
-
-    def check_access(self, operation, raise_exception=False):
-        res = super(DmsDirectory, self).check_access(operation, raise_exception)
-        if self.env.user.has_group("base.group_portal"):
-            if self.id in self._get_ids_without_access_groups(operation):
-                res = False
-        # Fix show breadcrumb with share button (public)
-        if self.env.user.has_group("base.group_public"):
-            res = True
-        return res
 
     allowed_model_ids = fields.Many2many(
         compute="_compute_allowed_model_ids", comodel_name="ir.model", store=False
@@ -289,6 +296,18 @@ class DmsDirectory(models.Model):
         string="Record Referenced", compute="_compute_record_ref", selection=[]
     )
     storage_id_save_type = fields.Selection(related="storage_id.save_type", store=False)
+    root_storage_id_inherit_access_from_parent_record = fields.Boolean(
+        related="root_storage_id.inherit_access_from_parent_record",
+        related_sudo=True,
+        auto_join=True,
+        store=True,
+    )
+    storage_id_inherit_access_from_parent_record = fields.Boolean(
+        related="storage_id.inherit_access_from_parent_record",
+        related_sudo=True,
+        auto_join=True,
+        store=True,
+    )
 
     @api.depends("root_storage_id", "storage_id")
     def _compute_allowed_model_ids(self):
@@ -364,26 +383,6 @@ class DmsDirectory(models.Model):
     # ----------------------------------------------------------
     # Search
     # ----------------------------------------------------------
-    @api.model
-    def _search(
-        self,
-        args,
-        offset=0,
-        limit=None,
-        order=None,
-        count=False,
-        access_rights_uid=None,
-    ):
-        result = super(DmsDirectory, self)._search(
-            args, offset, limit, order, False, access_rights_uid
-        )
-        if result:
-            directory_ids = set(result)
-            if self.env.user.has_group("base.group_portal"):
-                exclude_ids = self._get_ids_without_access_groups("read")
-                directory_ids -= set(exclude_ids)
-                return directory_ids
-        return result
 
     @api.model
     def _search_starred(self, operator, operand):
@@ -417,14 +416,16 @@ class DmsDirectory(models.Model):
     @api.depends("child_directory_ids")
     def _compute_count_directories(self):
         for record in self:
-            directories = len(record.child_directory_ids)
+            directories = len(
+                record.child_directory_ids.filtered(lambda x: x.check_access("read"))
+            )
             record.count_directories = directories
             record.count_directories_title = _("%s Subdirectories") % directories
 
     @api.depends("file_ids")
     def _compute_count_files(self):
         for record in self:
-            files = len(record.file_ids)
+            files = len(record.file_ids.filtered(lambda x: x.check_access("read")))
             record.count_files = files
             record.count_files_title = _("%s Files") % files
 
@@ -510,6 +511,24 @@ class DmsDirectory(models.Model):
             )
             record.tag_ids = tags
 
+    @api.onchange("storage_id")
+    def _onchange_storage_id(self):
+        for record in self:
+            if (
+                not record.is_root_directory
+                and record.storage_id.save_type == "attachment"
+                and record.storage_id.inherit_access_from_parent_record
+            ) or (
+                record.is_root_directory
+                and record.root_storage_id.save_type == "attachment"
+                and record.root_storage_id.inherit_access_from_parent_record
+            ):
+                record.group_ids = False
+
+    @api.onchange("model_id")
+    def _onchange_model_id(self):
+        self._inverse_model_id()
+
     # ----------------------------------------------------------
     # Constrains
     # ----------------------------------------------------------
@@ -519,6 +538,14 @@ class DmsDirectory(models.Model):
         if not self._check_recursion():
             raise ValidationError(_("Error! You cannot create recursive directories."))
         return True
+
+    @api.constrains("storage_id", "model_id")
+    def _check_storage_id_attachment_model_id(self):
+        for record in self:
+            if record.storage_id.save_type == "attachment" and not record.model_id:
+                raise ValidationError(
+                    _("A directory has to have model in attachment storage.")
+                )
 
     @api.constrains("is_root_directory", "root_storage_id", "parent_id")
     def _check_directory_storage(self):
@@ -548,7 +575,7 @@ class DmsDirectory(models.Model):
     @api.constrains("name")
     def _check_name(self):
         for record in self:
-            if not check_name(record.name):
+            if self.env.context.get("check_name", True) and not check_name(record.name):
                 raise ValidationError(_("The directory name is invalid."))
             if record.is_root_directory:
                 childs = record.sudo().root_storage_id.root_directory_ids.name_get()
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -1,5 +1,6 @@
 # Copyright 2020 Antoni Romera
 # Copyright 2017-2019 MuK IT GmbH
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import base64
@@ -150,6 +151,10 @@ class File(models.Model):
     content_file = fields.Binary(
         attachment=True, string="Content File", prefetch=False, invisible=True
     )
+
+    def check_access_rule(self, operation):
+        self.mapped("directory_id").check_access_rule(operation)
+        return super().check_access_rule(operation)
 
     def get_human_size(self):
         return human_size(self.size)
@@ -557,22 +562,6 @@ class File(models.Model):
             records -= self.browse(directory.sudo().mapped("file_ids").ids)
         return records
 
-    def check_access(self, operation, raise_exception=False):
-        res = super(File, self).check_access(operation, raise_exception)
-        try:
-            if self.env.user.has_group("base.group_portal"):
-                res_access = res and self.check_directory_access(operation)
-                return res_access and (
-                    self.directory_id.id
-                    not in self.directory_id._get_ids_without_access_groups(operation)
-                )
-            else:
-                return res and self.check_directory_access(operation)
-        except AccessError:
-            if raise_exception:
-                raise
-            return False
-
     def check_directory_access(self, operation, vals=False, raise_exception=False):
         if not vals:
             vals = {}
@@ -587,6 +576,16 @@ class File(models.Model):
     # ----------------------------------------------------------
     # Constrains
     # ----------------------------------------------------------
+
+    @api.constrains("storage_id", "res_model")
+    def _check_storage_id_attachment_res_model(self):
+        for record in self:
+            if record.storage_id.save_type == "attachment" and not (
+                record.res_model and record.res_id
+            ):
+                raise ValidationError(
+                    _("A file must have model and resource ID in attachment storage.")
+                )
 
     @api.constrains("name")
     def _check_name(self):
--- a/dms/models/dms_security_mixin.py
+++ b/dms/models/dms_security_mixin.py
@@ -1,4 +1,5 @@
 # Copyright 2020 Creu Blanca
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 from collections import defaultdict
@@ -135,6 +136,22 @@ class DmsSecurityMixin(models.AbstractModel):
             )
         query.where_clause += [where_clause]
         query.where_clause_params += [self.env.user.id]
+        # Add _get_directory_ids_with_res_model_without_access
+        if self.env.context.get("use_res_model_without_access", True):
+            custom_ids = self._get_directory_ids_with_res_model_without_access(mode)
+            if custom_ids:
+                field = "id"
+                if self._name == "dms.file":
+                    field = "directory_id"
+                extra_clause = """
+                    "{table}".{field} NOT IN ({ids})
+                """.format(
+                    table=self._table,
+                    field=field,
+                    ids=", ".join(["%s"] * len(custom_ids)),
+                )
+                query.where_clause += [extra_clause]
+                query.where_clause_params += custom_ids.ids
 
     @api.model
     def _apply_ir_rules(self, query, mode="read"):
@@ -222,11 +239,8 @@ class DmsSecurityMixin(models.AbstractModel):
         try:
             access_right = self.check_access_rights(operation, raise_exception)
             access_rule = self.check_access_rule(operation) is None
-            return (
-                access_right
-                and access_rule
-                and self.check_access_groups(operation) is None
-            )
+            access_group = self.check_access_groups(operation) is None
+            return access_right and access_rule and access_group
         except AccessError:
             if raise_exception:
                 raise
@@ -253,9 +267,10 @@ class DmsSecurityMixin(models.AbstractModel):
             result = defaultdict(list)
             for key, val in self.env.cr.fetchall():
                 result[key].append(val)
-            if len(result.keys()) < len(group_ids) or not all(
-                list(map(lambda val: any(val), result.values()))
-            ):
+            if (
+                len(result.keys()) < len(group_ids)
+                or not all(list(map(lambda val: any(val), result.values())))
+            ) and not self.check_access(operation):
                 raise AccessError(
                     _(
                         "The requested operation cannot be completed due "
@@ -288,6 +303,31 @@ class DmsSecurityMixin(models.AbstractModel):
             self.env.cr.execute(sql_query, [self.env.user.id])
             ids_with_access += list(map(lambda val: val[0], self.env.cr.fetchall()))
         return self & self.browse(ids_with_access)
+
+    def _get_directory_ids_with_res_model_without_access(self, operation):
+        """
+        It's necessary to get all directories with res_model related and check
+        if access to related record.
+        Context use_res_model_without_access=False is used to skip some
+        part of functions that use these function and prevent recursion error.
+        """
+        return (
+            self.env["dms.directory"]
+            .with_context(use_res_model_without_access=False)
+            .search(
+                [
+                    ("res_model", "!=", False),
+                    "|",
+                    "&",
+                    ("is_root_directory", "=", True),
+                    ("root_storage_id_inherit_access_from_parent_record", "=", True),
+                    "&",
+                    ("is_root_directory", "=", False),
+                    ("storage_id_inherit_access_from_parent_record", "=", True),
+                ]
+            )
+            .filtered(lambda x: not x.check_access("read"))
+        )
 
     def _write(self, vals):
         self.check_access_groups("write")
--- a/dms/models/ir_attachment.py
+++ b/dms/models/ir_attachment.py
@@ -7,26 +7,23 @@ class IrAttachment(models.Model):
     _inherit = "ir.attachment"
 
     def _get_dms_directories(self, res_model, res_id):
-        return (
-            self.env["dms.directory"]
-            .sudo()
-            .search(
-                [
-                    ("res_model", "=", res_model),
-                    ("res_id", "=", res_id),
-                    ("storage_id.save_type", "=", "attachment"),
-                ]
-            )
-        )
+        domain = [
+            ("res_model", "=", res_model),
+            ("res_id", "=", res_id),
+            ("storage_id.save_type", "=", "attachment"),
+        ]
+        if self.env.context.get("attaching_to_record"):
+            domain += [("storage_id.include_message_attachments", "=", True)]
+        return self.env["dms.directory"].search(domain)
 
     def _dms_directories_create(self):
         items = self._get_dms_directories(self.res_model, False)
         for item in items:
-            model_item = self.env[self.res_model].sudo().browse(self.res_id)
+            model_item = self.env[self.res_model].browse(self.res_id)
             ir_model_item = self.env["ir.model"].search(
                 [("model", "=", self.res_model)]
             )
-            self.env["dms.directory"].sudo().create(
+            self.env["dms.directory"].sudo().with_context(check_name=False).create(
                 {
                     "name": model_item.display_name,
                     "model_id": ir_model_item.id,
@@ -54,17 +51,25 @@ class IrAttachment(models.Model):
                 directories = self._get_dms_directories(
                     attachment.res_model, attachment.res_id
                 )
-            # Auto-create_files
+            # Auto-create_files (if not exists)
             for directory in directories:
-                self.env["dms.file"].sudo().create(
-                    {
-                        "name": attachment.name,
-                        "directory_id": directory.id,
-                        "attachment_id": attachment.id,
-                        "res_model": attachment.res_model,
-                        "res_id": attachment.res_id,
-                    }
+                dms_file_model = self.env["dms.file"].sudo()
+                dms_file = dms_file_model.search(
+                    [
+                        ("attachment_id", "=", attachment.id),
+                        ("directory_id", "=", directory.id),
+                    ]
                 )
+                if not dms_file:
+                    dms_file_model.create(
+                        {
+                            "name": attachment.name,
+                            "directory_id": directory.id,
+                            "attachment_id": attachment.id,
+                            "res_model": attachment.res_model,
+                            "res_id": attachment.res_id,
+                        }
+                    )
 
     @api.model_create_multi
     def create(self, vals_list):
@@ -72,3 +77,11 @@ class IrAttachment(models.Model):
         if not self.env.context.get("dms_file"):
             records._dms_operations()
         return records
+
+    def write(self, vals):
+        res = super().write(vals)
+        if not self.env.context.get("dms_file") and self.env.context.get(
+            "attaching_to_record"
+        ):
+            self._dms_operations()
+        return res
--- a/dms/models/storage.py
+++ b/dms/models/storage.py
@@ -85,6 +85,19 @@ class Storage(models.Model):
     )
 
     model_ids = fields.Many2many("ir.model", string="Linked Models")
+    inherit_access_from_parent_record = fields.Boolean(
+        string="Inherit permissions from related record",
+        default=False,
+        help="Indicate if directories and files access work only with "
+        "related model access (for example, if some directories are related "
+        "with any sale, only users with read access to these sale can acess)",
+    )
+    include_message_attachments = fields.Boolean(
+        string="Create files from message attachments",
+        default=False,
+        help="Indicate if directories and files auto-create in mail "
+        "composition process too",
+    )
 
     # ----------------------------------------------------------
     # Actions
--- a/dms/security/ir.model.access.csv
+++ b/dms/security/ir.model.access.csv
@@ -6,15 +6,16 @@ access_dms_category_user,dms_category_user,model_dms_category,group_dms_user,1,1
 access_dms_storage_user,dms_storage_user,model_dms_storage,group_dms_user,1,0,0,0
 access_dms_storage_manager,dms_storage_manager,model_dms_storage,group_dms_manager,1,1,1,1
 
+access_dms_directory_public,dms_directory_public,model_dms_directory,base.group_public,1,0,0,0
+access_dms_directory_portal,dms_directory_portal,model_dms_directory,base.group_portal,1,0,0,0
+access_dms_directory_base_user,dms_directory_base_user,model_dms_directory,base.group_user,1,0,0,0
 access_dms_directory_user,dms_directory_user,model_dms_directory,group_dms_user,1,1,1,1
+
+access_dms_file_public,dms_file_public,model_dms_file,base.group_public,1,0,0,0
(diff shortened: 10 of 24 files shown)


dms/views/storage.xml
```diff
@@ -166,6 +166,10 @@
                     <group name="data_storage">
                         <group>
                             <field name="is_hidden" />
+                            <field
+                                name="inherit_record_mail_followers"
```
pedrobaeza:
```suggestion
                                name="inherit_access_from_parent_record"
```

dms/demo/access_group.xml
```diff
@@ -12,8 +12,13 @@
...
     <record id="access_group_02_demo" model="dms.access.group">
-        <field name="name">Demo</field>
+        <field name="name">Portal</field>
         <field name="perm_read">True</field>
         <field name="group_ids" eval="[(6, 0, [ref('base.group_portal')])]" />
     </record>
+    <record id="access_group_03_demo" model="dms.access.group">
+        <field name="name">Only admin user</field>
+        <field name="perm_read">True</field>
+        <field name="group_ids" eval="[(6, 0, [ref('base.user_admin')])]" />
```
yajo: Shouldn't this ref point to a group and not to a user?

dms/models/directory.py
```diff
@@ -265,11 +265,15 @@ def _get_own_root_directories(self, user_id):
     def check_access(self, operation, raise_exception=False):
         res = super(DmsDirectory, self).check_access(operation, raise_exception)
         if self.env.user.has_group("base.group_portal"):
-            if self.id in self._get_ids_without_access_groups(operation):
+            if self.ids in self._get_ids_without_access_groups(operation):
```
yajo: This is probably a bug.

Both `self.ids` and `self._get_ids_without_access_groups(operation)` return `list` objects, and the code seems to try to act as if they were `set`s.

See:

```python
>>> a=[1,2]
>>> b=[2,3,1]
>>> a in b
False
>>> c={1,2}
>>> d={2,3,1}
>>> d in c
False
>>> d & c
{1, 2}
>>> bool(d & c)
True
```

Fix: 

```suggestion
            if set(self.ids) & set(self._get_ids_without_access_groups(operation)):
```

dms/models/directory.py
```diff
@@ -265,11 +265,15 @@ def _get_own_root_directories(self, user_id):
...
                 res = False
         # Fix show breadcrumb with share button (public)
         if self.env.user.has_group("base.group_public"):
             res = True
+        # Fix record_ref_mail_follower_ids
```
yajo: Possibly you'll have to remove references to followers, as it's not the main point here.

dms/models/dms_security_mixin.py
```diff
@@ -287,8 +311,30 @@ def filter_access_groups(self, operation):
...
+                [
+                    ("storage_id.inherit_access_from_parent_record", "=", True),
+                    ("res_model", "!=", False),
+                    ("res_id", "!=", 0),
+                ]
+            )
+        )
+        for directory in directories:
+            record = self.env[directory.res_model].browse(directory.res_id)
+            if record.check_access_rights("read") and directory.id not in ids:
```
yajo: AFAICS you have another bug here.

`check_access_rights()` is an `@api.model` method. This means that it's meant to be executed once per model, not per record.

OTOH, you also have `check_access_rule()`, which is `@api.multi` and should be checked per record, but I don't see it being used anywhere.

yajo (changes requested):
TBH I was expecting something more in the lines of:

```python
class DmsDirectory(models.Model):
    # ...
    @api.model
    def check_access_rights(self, operation, raise_exception=True):
        models = self.filtered('inherit_access_from_parent_record').mapped('res_model')
        result = super().check_access_rights(operation, raise_exception)
        for model in models:
            result = result or self.env[model].check_access_rights(operation, raise_exception)
            if result:
                return result
        # continue special stuff regarding access rights for dms.directory

        def check_access_rule(self, operation):
            # Similar logic as above, but for record rules

class DmsFile(models.Model):
    # Same logic as above. To access a file, we override both methods
    # and check the user has access to the file directory by calling the
    # directory equivalent methods.
```

However I'm aware this module does "its own stuff" regarding permissions... I guess it's not the time for a full refactor, but I found some bugs anyway.

BTW remember to change the menu & breadcrumb names in portal.

dms/models/directory.py
victoralmau: Interesting explanation about it, i have changed some things similar.

dms/models/dms_security_mixin.py
victoralmau: I have added some tests to check it and `check_access_rights` and `check_access_rule` don't work fine because allways return True although in some cases need to return False (remember in some case some records it's not accessible, private address for example). For these reasons i have used `record.read()`

victoralmau:
Now, everything work fine and test added. It's ready to review it.

yajo (changes requested):
Change these please:

imagen

imagen

victoralmau:
@Yajo can you review it again?

dms/models/directory.py
```diff
@@ -194,6 +195,40 @@ class DmsDirectory(models.Model):
...
+            try:
+                if record.res_id:
+                    self.env[model].browse(record.res_id).check_access_rule(operation)
+                else:
+                    self.env[model].check_access_rights(operation)
+            except AccessError:
+                raise AccessError(
+                    _("No tienes permisos de %s para acceder al registro %s,%s")
+                    % (operation, record.res_model, record.res_id)
+                )
```
yajo: This can be much simpler. Besides, `check_access_rights` must be called always:

```suggestion
            # Check access to inherited model (and record)
            model = self.env[record.res_model]
            model.check_access_rights(operation)
            if record.res_id:
                model.browse(record.res_id).check_access_rule(operation)
```

dms/models/directory.py
```diff
@@ -194,6 +195,40 @@ class DmsDirectory(models.Model):
...
+        key = "perm_%s" % operation
+        for record in self.filtered(lambda x: x.complete_group_ids):
+            if not record.complete_group_ids.filtered(lambda x: x[key]):
+                raise AccessError(
+                    _("No tienes permisos de %s para acceder al registro") % operation
```
yajo: ¡Se te coló en español! 😋

dms/models/directory.py
```diff
@@ -194,6 +195,40 @@ class DmsDirectory(models.Model):
...
+                )
+        return
+
+    def check_access_complete_group_ids(self, operation):
+        key = "perm_%s" % operation
+        for record in self:
+            if not record.complete_group_ids.filtered(lambda x: x[key]):
+                return False
+        return True
```
yajo: This can be much simpler too:

```suggestion
        return bool(self.mapped('complete_group_ids').filtered(key))
```

dms/models/directory.py
```diff
@@ -365,25 +388,12 @@ def _search_panel_directory(self, **kwargs):
...
-                exclude_ids = self._get_ids_without_access_groups("read")
-                directory_ids -= set(exclude_ids)
-                return directory_ids
-        return result
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        res = super(DmsDirectory, self).search(args, offset, limit, order, False)
+        for item in res:
+            if not item.check_access("read"):
+                res -= item
+        return res
```
yajo: It seems like this implementation will introduce bugs.

For example, if `count=True`, you're not passing that to `super()`. Also, even if you pass it, it wouldn't return accurate results.

Also, if `offset` or `limit` are used, the returned recordset won't have the expected size.

So, you'd be safer extending `_apply_ir_rules()` directly, which filters search results at SQL level and doesn't have these problems. 🤷🏼‍♂️ 

If you don't want to go that deep, you'll have to do more searches, which would be much less performant, but would work too.

dms/models/dms_file.py
```diff
@@ -151,6 +152,10 @@ class File(models.Model):
         attachment=True, string="Content File", prefetch=False, invisible=True
     )
 
+    def check_access_rule(self, operation):
+        self.mapped("directory_id").check_access_rule(operation)
+        return
```
yajo:
```suggestion
        return super().check_access_rule(operation)
```

dms/models/dms_security_mixin.py
```diff
@@ -287,8 +306,41 @@ def filter_access_groups(self, operation):
             )
             self.env.cr.execute(sql_query, [self.env.user.id])
             ids_with_access += list(map(lambda val: val[0], self.env.cr.fetchall()))
+            # Add related to record_ref_mail_follower_ids
+            if self._table == "dms_directory" and not self.env.context.get(
```
yajo: This wasn't wrong, but it'd be the common way to check it:

```suggestion
            if self._name == "dms.directory" and not self.env.context.get(
```

dms/controllers/portal.py
```diff
@@ -60,9 +62,9 @@ def portal_my_dms(
             (
                 "id",
                 "in",
-                request.env["dms.directory"]._get_own_root_directories(
-                    request.env.user.id
-                ),
+                request.env["dms.directory"]
+                .sudo(request.env.user)
```
yajo: That's redundant:

```suggestion
```

dms/controllers/portal.py
```diff
@@ -31,8 +31,10 @@ def _dms_check_access(self, model, res_id, access_token=None):
 
     def _prepare_portal_layout_values(self):
         values = super()._prepare_portal_layout_values()
-        ids = request.env["dms.directory"]._get_own_root_directories(
-            request.env.user.id
+        ids = (
+            request.env["dms.directory"]
+            .sudo(request.env.user)
```
yajo: It's redundant: 

```suggestion
```

yajo (changes requested):

dms/models/directory.py
victoralmau: I understand that using this is not a good solution and needs to be improved.
To extend `_apply_ir_rules()` and include the accessible `directory_ids` in the query, you would need to do `.search()` of all records with `[('res_model', '! =', False)]` and get all that comply with `check_access("read")` BUT without going into a recursive loop, so a context variable would be used.
Do you think this change would be the correct one?

victoralmau: I have changed about it.

dms/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Document Management System",
     "summary": """Document Management System for Odoo""",
-    "version": "12.0.4.0.7",
+    "version": "12.0.4.1.0",
```
yajo: This changes some past behaviors, so I think it justifies a major bump.

```suggestion
    "version": "12.0.5.0.0",
```

dms/controllers/portal.py
```diff
@@ -16,7 +16,7 @@
 class CustomerPortal(CustomerPortal):
     def _dms_check_access(self, model, res_id, access_token=None):
         try:
-            item_sudo = request.env[model].sudo().browse([res_id])
+            item_sudo = request.env[model].browse([res_id])
```
yajo: The variable name `item_sudo` is now misleading:

```suggestion
            item = request.env[model].browse(res_id)
```

Change usages below too, of course.

---

Kudos for removing some `sudo`s!

dms/migrations/12.0.4.1.0/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    directories = (
+        env["dms.directory"]
+        .sudo()
+        .search(
+            [("storage_id.save_type", "=", "attachment"), ("group_ids", "!=", "[]")]
```
yajo: I'm pretty sure it should be like this:

```suggestion
            [("storage_id.save_type", "=", "attachment"), ("group_ids", "!=", False)]
```

Please test it, this is only from memory. 😅 

---

Remember to rename this folder to `12.0.5.0.0`.

dms/migrations/12.0.4.1.0/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+        )
+    )
+    if directories:
+        for directory in directories:
+            directory.group_ids = None
```
yajo: The `if` is not needed:

```suggestion
    for directory in directories:
        directory.group_ids = None
```

dms/models/directory.py
```diff
@@ -194,6 +195,27 @@ class DmsDirectory(models.Model):
...
+                continue
+            # Check access to inherited model (and record)
+            model = self.env[record.res_model]
+            model.check_access_rights(operation)
+            if record.res_id:
+                model.browse(record.res_id).check_access_rule(operation)
+        key = "perm_%s" % operation
+        records = self - self.filtered(lambda x: x.res_model)
+        if bool(records.mapped("complete_group_ids").filtered(key)) is False:
+            return
```
yajo: This seems to check the mere existence of a group that has `operation` access, but doesn't check if your user belongs to that group, right? Shouldn't that happen?

---

Also: I'm missing a `return super().check_access_rule(operation)` (or equivalent) at some point. Possibly that should be the 1st thing your do in this method.

dms/models/directory.py
```diff
@@ -510,6 +502,12 @@ def _onchange_category(self):
             )
             record.tag_ids = tags
 
+    @api.onchange("storage_id")
+    def _onchange_storage_id(self):
+        for record in self:
+            if record.storage_id.save_type == "attachment":
+                record.group_ids = None
```
yajo: In Odoo, the emptiness is represented as `False` usually:

```suggestion
                record.group_ids = False
```

dms/models/directory.py
```diff
@@ -520,6 +518,15 @@ def _check_directory_recursion(self):
...
         return True
 
+    @api.constrains("storage_id", "model_id")
+    def _check_storage_id_attachment_model_id(self):
+        for record in self:
+            if record.storage_id.save_type == "attachment" and not record.model_id:
+                raise ValidationError(
+                    _("A directory has to have model in attachment storage.")
+                )
+        return True
```
yajo: Constraints usually do not return. They just raise if something fails:

```suggestion
```

dms/models/directory.py
```diff
@@ -548,7 +555,9 @@ def _check_directory_access(self):
     @api.constrains("name")
     def _check_name(self):
         for record in self:
-            if not check_name(record.name):
+            if not self.env.context.get("bypass_check_name") and not check_name(
```
yajo: [I'm generally against negative booleans](https://www.recallstack.icu/en/2020/09/01/do-not-use-negative-booleans-please/). Better:

```suggestion
            if self.env.context.get("check_name", True) and not check_name(
```

dms/models/ir_attachment.py
```diff
@@ -7,26 +7,24 @@ class IrAttachment(models.Model):
...
         items = self._get_dms_directories(self.res_model, False)
         for item in items:
-            model_item = self.env[self.res_model].sudo().browse(self.res_id)
+            model_item = self.env[self.res_model].browse(self.res_id)
             ir_model_item = self.env["ir.model"].search(
                 [("model", "=", self.res_model)]
             )
-            self.env["dms.directory"].sudo().create(
+            self.env["dms.directory"].sudo().with_context(
+                bypass_check_name=True
```
yajo: Following the other comment above:

```suggestion
                check_name=False
```

dms/models/dms_file.py
```diff
@@ -588,6 +593,15 @@ def check_directory_access(self, operation, vals=False, raise_exception=False):
     # Constrains
     # ----------------------------------------------------------
 
+    @api.constrains("storage_id", "res_model")
+    def _check_storage_id_attachment_res_model(self):
+        for record in self:
+            if record.storage_id.save_type == "attachment" and not record.res_model:
```
yajo: If I'm not wrong, a file with this storage also requires a `res_id`, right?

```suggestion
            if record.storage_id.save_type == "attachment" and not (record.res_model and record.res_id):
```

dms/models/dms_file.py
```diff
@@ -588,6 +593,15 @@ def check_directory_access(self, operation, vals=False, raise_exception=False):
...
+                raise ValidationError(
+                    _("A file has to have model in attachment storage.")
```
yajo:
```suggestion
                    _("A file must have model and resource ID in attachment storage.")
```

dms/models/dms_file.py
```diff
@@ -588,6 +593,15 @@ def check_directory_access(self, operation, vals=False, raise_exception=False):
...
+                )
+        return True
```
yajo:
```suggestion
```

dms/models/dms_security_mixin.py
```diff
@@ -133,6 +134,23 @@ def _apply_access_groups(self, query, mode="read"):
             where_clause = "({groups_clause} OR {exists_clause})".format(
                 groups_clause=where_clause, exists_clause=exists_clause,
             )
+        # Add _get_directory_ids_with_res_model_without_access
+        if not self.env.context.get("bypass_ids_with_res_model_without_access"):
```
yajo:
```suggestion
        if not self.env.context.get("use_res_model_without_access", True):
```

dms/models/dms_security_mixin.py
```diff
@@ -287,8 +303,28 @@ def filter_access_groups(self, operation):
...
+                "bypass_ids_with_record_ref"
+            ):
+                ids_with_access += self.env["dms.directory"].search([])
         return self & self.browse(ids_with_access)
 
+    def _get_directory_ids_with_res_model_without_access(self, operation):
+        """
+        It's necessary to get all directories with res_model related and check
+        if access to related record.
+        Context bypass_ids_with_res_model_without_access is used to skip some
```
yajo:
```suggestion
        Context use_res_model_without_access=False is used to skip some
```

dms/models/dms_security_mixin.py
```diff
@@ -287,8 +303,28 @@ def filter_access_groups(self, operation):
...
+        part of functions that use these function and prevent recursion error.
+        """
+        return (
+            self.env["dms.directory"]
+            .with_context(bypass_ids_with_res_model_without_access=True)
```
yajo:
```suggestion
            .with_context(use_res_model_without_access=False)
```

dms/models/dms_security_mixin.py
```diff
@@ -133,6 +134,23 @@ def _apply_access_groups(self, query, mode="read"):
...
             )
+        # Add _get_directory_ids_with_res_model_without_access
+        if not self.env.context.get("bypass_ids_with_res_model_without_access"):
+            custom_ids = self._get_directory_ids_with_res_model_without_access(mode)
+            if len(custom_ids) > 0:
+                field = "id"
+                if self._name == "dms.file":
+                    field = "directory_id"
+                extra_clause = """
+                    "{table}".{field} NOT IN {ids}
```
yajo: This was previously expanding to (for example) `"dms_file".directory_id NOT IN (1), (2), (3)`, which is wrong SQL, right?

```suggestion
                    "{table}".{field} NOT IN ({ids})
```

dms/models/dms_security_mixin.py
```diff
@@ -133,6 +134,23 @@ def _apply_access_groups(self, query, mode="read"):
...
+                """.format(
+                    table=self._table,
+                    field=field,
+                    ids=", ".join(map(lambda id: "(%s)" % id, custom_ids)),
```
yajo: These 2 changes will make it expand to `"dms_file".directory_id NOT IN (%s,%s,%s)` instead.

```suggestion
                    ids=", ".join(["%s"] * len(custom_ids)),
```

dms/models/dms_security_mixin.py
```diff
@@ -133,6 +134,23 @@ def _apply_access_groups(self, query, mode="read"):
             where_clause = "({groups_clause} OR {exists_clause})".format(
                 groups_clause=where_clause, exists_clause=exists_clause,
             )
+        # Add _get_directory_ids_with_res_model_without_access
```
yajo, line 139: In this code block I'm gonna suggest some changes because I think you have a bug and because we should try to leverage `psycopg2` wherever possible to expand placehloder params.

```suggestion
        query.where_clause_params += [self.env.user.id]
        # Add _get_directory_ids_with_res_model_without_access
```

dms/models/dms_security_mixin.py
```diff
@@ -133,6 +134,23 @@ def _apply_access_groups(self, query, mode="read"):
...
+                """.format(
+                    table=self._table,
+                    field=field,
+                    ids=", ".join(map(lambda id: "(%s)" % id, custom_ids)),
+                )
+                where_clause = "({where_clause} AND {extra_clause})".format(
+                    where_clause=where_clause, extra_clause=extra_clause,
+                )
         query.where_clause += [where_clause]
         query.where_clause_params += [self.env.user.id]
```
yajo:
```suggestion
```

dms/models/dms_security_mixin.py
```diff
@@ -133,6 +134,23 @@ def _apply_access_groups(self, query, mode="read"):
...
+                extra_clause = """
+                    "{table}".{field} NOT IN {ids}
+                """.format(
+                    table=self._table,
+                    field=field,
+                    ids=", ".join(map(lambda id: "(%s)" % id, custom_ids)),
+                )
+                where_clause = "({where_clause} AND {extra_clause})".format(
+                    where_clause=where_clause, extra_clause=extra_clause,
+                )
```
yajo, line 152: And by adding this line, we tell `psycopg2` to use `custom_ids` to expand those `%s` above.

```suggestion
                )
                query.where_clause_params += custom_ids
```

yajo (changes requested):

yajo:
I've been doing functional review on runbot too, and found these issues:

imagen

imagen

imagen

dms/models/directory.py
victoralmau: I have changed to use `check_access_groups()`

victoralmau:
> I've been doing functional review on runbot too, and found these issues:
> 
> imagen
> 
> imagen
> 
> imagen

- Linked models duplicate in storage view is related to `dms_field` addon (i have added a commit to fix it).
- Model field in directory is required now if storage is attachment type.
- Download button work fine now.

dms/migrations/12.0.5.0.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+def migrate(env, version):
+    directories = (
+        env["dms.directory"]
+        .sudo()
+        .search(
+            [("storage_id.save_type", "=", "attachment"), ("group_ids", "!=", False)]
+        )
+    )
+    for directory in directories:
+        directory.group_ids = None
```
yajo: Batch this please:

```suggestion
    directories.write({"group_ids": False})
```

dms/models/dms_security_mixin.py
```diff
@@ -135,6 +136,22 @@ def _apply_access_groups(self, query, mode="read"):
...
+            if len(custom_ids) > 0:
+                field = "id"
+                if self._name == "dms.file":
+                    field = "directory_id"
+                extra_clause = """
+                    "{table}".{field} NOT IN ({ids})
+                """.format(
+                    table=self._table,
+                    field=field,
+                    ids=", ".join(["%s"] * len(custom_ids)) % tuple(custom_ids),
```
yajo: I still think we should leverage `psycopg2` placeholder expansion, which protects from injections:

```suggestion
                    ids=", ".join(["%s"] * len(custom_ids)),
```

I'm not sure if you did it this way due to something I'm missing, or if it was just a confusion... If I'm missing something please tell me 😅

dms/models/dms_security_mixin.py
```diff
@@ -135,6 +136,22 @@ def _apply_access_groups(self, query, mode="read"):
...
+                )
+                where_clause = "{extra_clause}".format(extra_clause=extra_clause)
+                query.where_clause += [where_clause]
```
yajo: This would complement the comment above (and fix one little nonsense you can see here BTW):

```suggestion
                query.where_clause += [extra_clause]
                query.where_clause_params += custom_ids
```

yajo (commented):
The code looks much better now. Just some little comments that I consider just recommendations.

Pending functional test to give final approval (runbot is rebuilding).

dms/models/dms_security_mixin.py
victoralmau: Sorry, your suggestion doesn't work because generate empty list.

victoralmau:
> The code looks much better now. Just some little comments that I consider just recommendations.
> 
> Pending functional test to give final approval (runbot is rebuilding).

Thanks a lot for your recommendations.

yajo:
Code looks very good now, although it's not working functionally. Tested on runbot:

1. Create this storage: 
screenshot-2021-04-16-09-19-31

1. Create this directory: 
imagen

1. Go to SO019 and send a message with an attachment. 
screenshot-2021-04-16-09-22-40

1. Log in as portal user, go to portal > Sales Orders. You can see SO019 with the message and attachment.
imagen

1. Go to portal > Documents > Sales: **it's empty**
imagen

yajo (approved):
It works perfectly now.

dms/migrations/12.0.5.0.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    directories = (
+        env["dms.directory"]
+        .sudo()
+        .search(
+            [("storage_id.save_type", "=", "attachment"), ("group_ids", "!=", False)]
+        )
+    )
+    directories.write({"group_ids": False})
```
yajo: This is failing when doing an actual migration:

```
2021-04-16 12:36:36,492 1 INFO prod OpenUpgrade: dms: post-migration script called with version 12.0.4.0.7 
2021-04-16 12:36:36,499 1 ERROR prod OpenUpgrade: dms: error in migration script dms/migrations/12.0.5.0.0/post-migration.py: 'project.task' 
2021-04-16 12:36:36,499 1 ERROR prod OpenUpgrade: 'project.task' 
  File "/usr/local/lib/python3.5/site-packages/openupgradelib/openupgrade.py", line 1875, in wrapped_function
    if use_env2 else cr, version)
  File "/opt/odoo/auto/addons/dms/migrations/12.0.5.0.0/post-migration.py", line 15, in migrate
    directories.write({"group_ids": False})
  File "/opt/odoo/auto/addons/dms/models/directory.py", line 662, in write
    res = super(DmsDirectory, self).write(vals)
  File "/opt/odoo/auto/addons/mail/models/mail_thread.py", line 321, in write
    result = super(MailThread, self).write(values)
  File "/opt/odoo/auto/addons/mail/models/mail_activity.py", line 621, in write
    return super(MailActivityMixin, self).write(vals)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3364, in write
    self._write(store_vals)
  File "/opt/odoo/auto/addons/dms/models/dms_security_mixin.py", line 329, in _write
    return super(DmsSecurityMixin, self)._write(vals)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3451, in _write
    self.check_access_rule('write')
  File "/opt/odoo/auto/addons/dms/models/directory.py", line 204, in check_access_rule
    model = self.env[record.res_model]
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 831, in __getitem__
    return self.registry[model_name]._browse((), self)
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 177, in __getitem__
    return self.models[model_name]
KeyError: 'project.task'
```

You must either move this to `end-migration` or avoid ORM usage.

yajo (changes requested):

dms/migrations/12.0.5.0.0/end-migration.py
```diff
@@ -0,0 +1,15 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    directories = (
+        env["dms.directory"]
+        .sudo()
+        .search(
+            [("storage_id.save_type", "=", "attachment"), ("group_ids", "!=", False)]
+        )
+    )
+    directories.write({"group_ids": False})
```
pedrobaeza: Why are you removing the content of this field for existing directories? Shouldn't they be only if the check is marked? And I think this should be done with SQL as well.

victoralmau: Finally not exists any check in storage because attachment storage access is according to access from record related or model related. Now, it's necessary to not exist groups in directories related and is not possible to assign groups in these directories from UI.

I have set these migration process to reset all fields related (`complete_groups_ids` and groups according to all files from these directories).

pedrobaeza: Talked directly and clarified.

victoralmau:
I have applied some changes according to @pedrobaeza comments, IMO it's important to review again and test it in runbot before merge.

dms/models/directory.py
```diff
@@ -194,6 +195,38 @@ class DmsDirectory(models.Model):
...
+                    and x.root_storage_id.inherit_access_from_parent_record
+                )
+                or (
+                    not x.is_root_directory
+                    and x.storage_id.inherit_access_from_parent_record
+                )
+            )
+        )
+        for record in records:
+            record.check_access_groups(operation)
```
yajo: Some comments about this code chunk:

1. When `directory.is_root_directory == True`, `storage_id` will be the same as `root_storage_id`: https://github.com/OCA/dms/blob/2fb661a5ef86afdf7eff51e6bdc1a46cb6591a7c/dms/models/directory.py#L381-L382
2. It seems you are entering a loop and discarding some elements, and then generating that loop again with the other elements. It would be easier to just loop everything and do a if-else.

So, this seems to be the same, but shorter:

```suggestion
        for record in self:
            if not record.storage_id.inherit_access_from_parent_record:
                record.check_access_groups(operation)
                continue
            # Check access to inherited model (and record)
            model = self.env[record.res_model]
            model.check_access_rights(operation)
            if record.res_id:
                model.browse(record.res_id).check_access_rule(operation)
```

dms/models/directory.py
```diff
@@ -510,6 +516,26 @@ def _onchange_category(self):
             )
             record.tag_ids = tags
 
+    @api.onchange("model_id")
+    def _onchange_model_id(self):
+        for record in self:
+            if record.model_id:
+                record.res_model = record.model_id.model
```
yajo: This is not something that should be done in an onchange (depends on the form), but in a compute (done always, at ORM level), right?. Related fields are computed fields behind the scenes. So, just make `res_model` a `fields.Char(related='model_id.model', readonly=True)`

dms/models/directory.py
```diff
@@ -510,6 +516,26 @@ def _onchange_category(self):
...
+        for record in self:
+            if (
+                not record.is_root_directory
+                and record.storage_id.save_type == "attachment"
+                and record.storage_id.inherit_access_from_parent_record
+            ) or (
+                record.is_root_directory
+                and record.root_storage_id.save_type == "attachment"
+                and record.root_storage_id.inherit_access_from_parent_record
+            ):
```
yajo, line 525: In harmony with other suggestions:

```suggestion
            if record.storage_id.inherit_access_from_parent_record:
```

dms/models/dms_security_mixin.py
```diff
@@ -289,6 +304,32 @@ def filter_access_groups(self, operation):
...
+            .search(
+                [
+                    ("res_model", "!=", False),
+                    "|",
+                    "&",
+                    ("is_root_directory", "=", True),
+                    ("root_storage_id.inherit_access_from_parent_record", "=", True),
+                    "&",
+                    ("is_root_directory", "=", False),
+                    ("storage_id.inherit_access_from_parent_record", "=", True),
```
yajo: Same stuff:

```suggestion
                    ("storage_id.inherit_access_from_parent_record", "=", True),
```

dms/models/storage.py
```diff
@@ -85,6 +85,12 @@ class Storage(models.Model):
     )
 
     model_ids = fields.Many2many("ir.model", string="Linked Models")
+    inherit_access_from_parent_record = fields.Boolean(
+        string="Inherit access from parent record", default=False
+    )
+    attachment_write_need_check = fields.Boolean(
+        string="Attachment write need check", default=False
+    )
```
yajo: Please add:

1. Help attributes on both fields.
2. An onchange that sets `inherit_access_from_parent_record = False` when `save_type != 'attachment'` (this should make all my other comments actually work).

Also, I think `attachment_write_need_check` is not a very helpful name or logic. What the user cares about is usually the fact that the attachment comes from a chatter message or that it was attached manually to the record, right? That's what the name should do (and it should be the actual implementation) IIUC.

A standard human being doesn't know that Odoo calls `write` the act of editing a record, and even less than every attachment coming from an email goes through that. Let's make this more friendly. 😃

dms/models/ir_attachment.py
```diff
@@ -54,21 +51,37 @@ def _dms_operations(self):
...
         records = super().create(vals_list)
         if not self.env.context.get("dms_file"):
             records._dms_operations()
         return records
+
+    @api.multi
+    def write(self, vals):
+        res = super().write(vals)
+        if not self.env.context.get("dms_file"):
+            self.with_context(attachment_action_write=True)._dms_operations()
```
yajo: If this is done to differentiate attachments that come from a message and get attached to a record, shouldn't we check first than the `res_model,res_id` combination has changed, and trigger the DMS stuff only on that case?

If somebody goes and just changes the attachment name, would we really want to trigger this? How about some module that edits an attachment in some other form? Another alternative: is there a context key that tells us this write is being issued from the mail system?

yajo (changes requested):

dms/models/directory.py
victoralmau: I'm not really sure if these change will have side effects. Something similar is commented here https://github.com/OCA/dms/pull/67#discussion_r608528935
IMO it's a good idea to change `root_storage_id` to `storage_id` and add  consistency to the fields BUT it's important to ask to @etobella (probablly kwnow more about it and if we can change it) and apply these change in other PR about to code re-factoring.

dms/models/directory.py
victoralmau: It's other important change that IMO it's necessary to apply but probably it's a better idea to apply in other PR but yes, will be related field, need to remove `_inverse_model_id` and simplify to `res_model` in `dms.file` (because it's other related about `directory_id.res_model`).
Now, i have added these function to prevent "inconsistent" data.

dms/views/directory.xml
```diff
@@ -552,7 +556,10 @@
                                 </group>
                             </group>
                         </page>
-                        <page string="Groups">
+                        <page
+                            string="Groups"
+                            attrs="{'invisible':[('storage_id_inherit_access_from_parent_record','=','attachment')]}"
```
yajo: AFAICS this is wrong:

```suggestion
                            attrs="{'invisible':[('storage_id_inherit_access_from_parent_record', '=', True)]}"
```

dms/views/directory.xml
```diff
@@ -563,7 +570,10 @@
                                 </tree>
                             </field>
                         </page>
-                        <page string="Complete Groups">
+                        <page
+                            string="Complete Groups"
+                            attrs="{'invisible':[('storage_id_inherit_access_from_parent_record','=','attachment')]}"
```
yajo:
```suggestion
                            attrs="{'invisible':[('storage_id_inherit_access_from_parent_record', '=', True)]}"
```

yajo (changes requested):

dms/models/ir_attachment.py
victoralmau: I have changed to use `mail_post_autofollow` context because exists in mail process and it's a better context.

dms/models/storage.py
```diff
@@ -85,6 +85,19 @@ class Storage(models.Model):
...
     model_ids = fields.Many2many("ir.model", string="Linked Models")
+    inherit_access_from_parent_record = fields.Boolean(
+        string="Inherit access from parent record",
+        default=False,
+        help="Indicate if directories and files access work only with "
+        "related model access (for example, if some directories are related "
+        "with any sale, only users with read access to these sale can acess)",
+    )
+    mail_post_autofollow = fields.Boolean(
+        string="Mail post autoflow",
```
yajo: Remember the string is for humans 😋:

```suggestion
        string="Create files from message attachments",
```

dms/models/storage.py
```diff
@@ -85,6 +85,19 @@ class Storage(models.Model):
     )
 
     model_ids = fields.Many2many("ir.model", string="Linked Models")
+    inherit_access_from_parent_record = fields.Boolean(
+        string="Inherit access from parent record",
```
yajo:
```suggestion
        string="Inherit permissions from related record",
```

yajo (changes requested):

dms/models/directory.py
yajo: As mentioned privately, this onchange can be removed because it overlaps with `_inverse_model_id()`:

https://github.com/OCA/dms/blob/d1ec8225c0a04de67fec79c2b2ed83dc75c37795/dms/models/directory.py#L315-L317

pedrobaeza: No, it doesn't. This applies the change immediately, not on save. It's a usual technique for acting like a "compute".

pedrobaeza: Anyway, the usual one is to call directly the inverse method, not to duplicate code.

dms/demo/storage.xml
```diff
@@ -10,4 +10,11 @@
         <field name="name">Documents Storage</field>
         <field name="save_type">database</field>
     </record>
+    <record id="storage_attachment_demo" model="dms.storage">
+        <field name="name">Attachment Storage</field>
+        <field name="save_type">attachment</field>
+        <field name="inherit_access_from_parent_record" eval="True" />
+        <field name="attachment_write_need_check" eval="True" />
```
yajo: This has changed.

yajo (changes requested):

dms/views/storage.xml
```diff
@@ -166,6 +166,14 @@
                     <group name="data_storage">
                         <group>
                             <field name="is_hidden" />
+                            <field
+                                name="inherit_access_from_parent_record"
+                                attrs="{'invisible': [('save_type', '!=', 'attachment')]}"
+                            />
+                            <field
+                                name="mail_post_autofollow"
```
yajo: Change

yajo (changes requested):

yajo (approved):

dms/models/dms_security_mixin.py
```diff
@@ -289,6 +304,32 @@ def filter_access_groups(self, operation):
...
+                ]
+            )
+            .filtered(lambda x: not x.check_access("read"))
+            .ids
```
pedrobaeza: Why returning `ids`, which always gives you less options in case you need any other file? Unless the method is orm_cached, return always recordsets

dms/models/directory.py
```diff
@@ -510,6 +502,24 @@ def _onchange_category(self):
             )
             record.tag_ids = tags
 
+    @api.onchange("storage_id")
+    def _onchange_storage_id(self):
+        for record in self:
+            if (
+                not record.is_root_directory
```
pedrobaeza, line 518: Why this duality between `storage_id` and `root_storage_id`. Can't you homogenize both?

dms/models/ir_attachment.py
```diff
@@ -54,21 +51,41 @@ def _dms_operations(self):
...
-                self.env["dms.file"].sudo().create(
-                    {
-                        "name": attachment.name,
-                        "directory_id": directory.id,
-                        "attachment_id": attachment.id,
-                        "res_model": attachment.res_model,
-                        "res_id": attachment.res_id,
-                    }
+                dms_file_model = self.env["dms.file"].sudo()
+                dms_file = dms_file_model.search(
```
pedrobaeza, line 57: Why is this called several times? The code this way is uglier... Maybe you can add a constraint and try to create always, although the ideal thing is to not call this several times.

dms/models/ir_attachment.py
```diff
@@ -54,21 +51,41 @@ def _dms_operations(self):
...
+                    )
 
     @api.model_create_multi
     def create(self, vals_list):
         records = super().create(vals_list)
         if not self.env.context.get("dms_file"):
             records._dms_operations()
         return records
+
+    @api.multi
```
pedrobaeza: Don't put this deprecated decorator

dms/security/ir.model.access.csv
```diff
@@ -3,18 +3,22 @@ id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
 access_dms_tag_user,dms_tag_user,model_dms_tag,group_dms_user,1,1,1,1
 access_dms_category_user,dms_category_user,model_dms_category,group_dms_user,1,1,1,1
 
+access_dms_storage_public,dms_storage_public,model_dms_storage,base.group_public,1,0,0,0
```
pedrobaeza: Why giving this permission (and others) to public/portal? I don't see that as a good default. You won't see this technique in any other hidden model.

dms/security/security.xml
```diff
@@ -67,4 +68,30 @@
         <field name="perm_unlink" eval="1" />
         <field name="domain_force">[(1 ,'=', 1)]</field>
     </record>
+    <record id="rule_security_dms_directory" model="ir.rule">
+        <field name="name">Dms Directories with dms group</field>
```
pedrobaeza: I'm not very comfortable exposing this via UI, so please put in the name of both that this will be filtered later by code.

dms/security/security.xml
```diff
@@ -67,4 +68,30 @@
...
+            eval="[(4, ref('base.group_user')),(4, ref('base.group_portal'))]"
+        />
+        <field name="perm_read" eval="1" />
+        <field name="perm_create" eval="0" />
+        <field name="perm_write" eval="0" />
+        <field name="perm_unlink" eval="0" />
+        <field name="domain_force">[(1,'=',1)]</field>
+    </record>
+    <record id="rule_security_dms_file" model="ir.rule">
+        <field name="name">Dms Files with dms group</field>
```
pedrobaeza: The same

dms/models/ir_attachment.py
pedrobaeza: Anyway, shouldn't it be enough to search by directory and attachment?

dms/models/directory.py
victoralmau: Both fields (`storage_id` and `root_storage_id`) exist yet in model. It's necessary 2 fields for same thing? IMO not, and something similar is https://github.com/OCA/dms/pull/67 to fix error.
IMO it's a good idea to homogenize it in only `storage_id` and convert to many2one only (without compute) BUT it's only one of a lot things that is need to improve, refactor and better code that will be a good idea to create after merge these PR create other PR about it.
What do you think about it?

dms/models/ir_attachment.py
victoralmau: Now it's possible to enter in these function when `create` and `write` attachment. It's very important to check if `dms.file` exists or will be create 2 times (create and write). Checking if directory exists (according to record, sale for example) is not enough to check if `dms.file` is need to create.

pedrobaeza: Searching for directory id + attachment id isn't enough? An attachment can be linked only to one record, so no duplicates.

dms/security/ir.model.access.csv
victoralmau: It's apply according to work fine some things:
- Prevent apply .`sudo() `in controller recordset.
- Prevent inconsistent data (if apply always `.sudo()` total elements in directories it's not correct because show total elements and not total elements according to request user can access).
- Prevent incorrect `check_access` when directories and files have storage with inherit permissions.

dms/models/directory.py
pedrobaeza: Let's do a POC in another PR to see if we merge it before this one and simplify it.

dms/security/ir.model.access.csv
pedrobaeza: But storages don't have permissions. The permissions are on directories. Storages should be still be accessed with `sudo`. If not, you are exposing a security hole. Anyone will be able to query via XML-RPC your storages definition. You can use `related` fields with `related_sudo=True` for avoiding the need of put `sudo` everywhere, and also `auto_join` when needed.

dms/models/directory.py
yajo: According to our tests, not doing it like this triggers an infinite loop because at this point `storage_id` is not yet computed. However, it'd be good to at least add a comment explaining that, for the next one that thinks the same (I did too 😆).

pedrobaeza: Yes, but we are talking about removing the duplicity of `storage_id` and `root_storage_id`, and thus, simplify all of this code. In that PoC, we will be able to handle that thing.

yajo: I'm not sure we can remove that duplicity.

In the case of a root folder, it needs to be specified a storage. In the case of a non-root folder, it needs to not specify that but inherit it from parent instead. You can understand that easily in https://github.com/OCA/dms/blob/d8fab70dd92222134b493b36272afe1ab606785c/dms/models/directory.py#L404-L410

What we should be able to remove is the usage of `root_storage_id` across the codebase, as `storage_id` will always have the correct data we need.

However, I agree with @victoralmau and this should be left for later. This PR is already in production and working fine. A refactor potentially will break things and I see no real need to do it in this same PR.

dms/security/ir.model.access.csv
victoralmau: I have checked to remove it (related to dms.storage) but maybe isn't possible because ```_get_directory_ids_with_res_model_without_access``` need to filter by (`"storage_id.inherit_access_from_parent_record", "=", True)`, and portal (for example) can't access or public user if we share directory link.
Any idea to solve it?

dms/models/directory.py
pedrobaeza: I think that can be done with a computed writable field, and setting readonly according that conditions. I wanted to clean this code before it lands production, but OK if you don't want more iterations. I also agree we have put a lot of time already.

dms/security/ir.model.access.csv
pedrobaeza: As said, a related field with `related_sudo` and `auto_join`.

dms/models/directory.py
yajo: Yes I agree with you @pedrobaeza. Actually you can see I suggested the same change yesterday: https://github.com/OCA/dms/pull/68#discussion_r615603500

But really we tried and it died, so I think we better leave it for later. 😕

pedrobaeza: Yes, it's been done in #67 for v13. We will leave this as is.

pedrobaeza (approved):

joao-p-marques:
@victoralmau can you fix the conflict here?

pedrobaeza:
This wasn't merged at all...

/ocabot merge nobump

PR #74, opened 2021-04-26
[13.0][IMP] dms: Inherit permissions from related records

victoralmau:
FWP from 12.0: https://github.com/OCA/dms/pull/68

Add Inherit access from parent record (in storage).

Locked by:
- [x] https://github.com/OCA/dms/pull/67

@Tecnativa TT29041

### DIFF
--- a/dms/controllers/portal.py
+++ b/dms/controllers/portal.py
@@ -16,24 +16,22 @@ _logger = logging.getLogger(__name__)
 class CustomerPortal(CustomerPortal):
     def _dms_check_access(self, model, res_id, access_token=None):
         try:
-            item_sudo = request.env[model].sudo().browse([res_id])
+            item = request.env[model].browse(res_id)
         except (AccessError, MissingError):
             return False
 
         if access_token:
-            if not item_sudo.check_access_token(access_token):
+            if not item.check_access_token(access_token):
                 return False
         else:
-            if not item_sudo.with_user(request.env.user.id).check_access("read", False):
+            if not item.with_user(request.env.user.id).check_access("read", False):
                 return False
 
-        return item_sudo
+        return item
 
     def _prepare_portal_layout_values(self):
         values = super()._prepare_portal_layout_values()
-        ids = request.env["dms.directory"]._get_own_root_directories(
-            request.env.user.id
-        )
+        ids = request.env["dms.directory"]._get_own_root_directories()
         values.update({"dms_directory_count": len(ids)})
         return values
 
@@ -57,13 +55,7 @@ class CustomerPortal(CustomerPortal):
             filterby = "name"
         # domain
         domain = [
-            (
-                "id",
-                "in",
-                request.env["dms.directory"]._get_own_root_directories(
-                    request.env.user.id
-                ),
-            )
+            ("id", "in", request.env["dms.directory"]._get_own_root_directories(),)
         ]
         # search
         if search and search_in:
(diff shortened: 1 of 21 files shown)


pedrobaeza:
I think this should wait until #67 is merged and the code is simplified.

victoralmau:
It's ready to merge now, right?

dms/controllers/portal.py
```diff
@@ -16,24 +16,22 @@
...
         except (AccessError, MissingError):
             return False
 
         if access_token:
-            if not item_sudo.check_access_token(access_token):
+            if not item.check_access_token(access_token):
                 return False
         else:
-            if not item_sudo.with_user(request.env.user.id).check_access("read", False):
+            if not item.sudo(request.env.user.id).check_access("read", False):
```
yajo: This should be `with_user`, isn't it?

dms/models/dms_file.py
```diff
@@ -131,6 +131,19 @@ class File(models.Model):
         attachment=True, string="Content File", prefetch=False, invisible=True
     )
 
+    def check_access_rule(self, operation):
+        self.mapped("directory_id").check_access_rule(operation)
+        return super().check_access_rule(operation)
+
+    def get_human_size(self):
+        return human_size(self.size)
```
yajo: It seems this diff is unnecessary if you keep the method where it was.

dms/models/storage.py
```diff
@@ -85,6 +85,19 @@ class Storage(models.Model):
     )
 
     model_ids = fields.Many2many("ir.model", string="Linked Models")
+    inherit_access_from_parent_record = fields.Boolean(
+        string="Inherit permissions from related record",
+        default=False,
```
yajo, line 91: IMHO this is the most sane default (see TT29500).

```suggestion
        default=True,
```

dms/models/storage.py
```diff
@@ -85,6 +85,19 @@ class Storage(models.Model):
...
+        help="Indicate if directories and files access work only with "
+        "related model access (for example, if some directories are related "
+        "with any sale, only users with read access to these sale can acess)",
+    )
+    include_message_attachments = fields.Boolean(
+        string="Create files from message attachments",
+        default=False,
```
yajo, line 98: Same here:

```suggestion
        default=True,
```

yajo (changes requested):
Some little details, but it looks good overall to me.

I'd appreciate to see more coverage, given the importance of this feature.

dms/models/storage.py
pedrobaeza: No, this shouldn't be enabled by default, as it's an opt-in feature, and even more, you will be changing current behavior on existing installations.

dms/models/storage.py
pedrobaeza: The same as above.

dms/models/storage.py
yajo: Setting `False` here means that you want to bypass Odoo's security model. It opens a door for data leaks, so IMHO it should be `True` by default.

The behavioural change is true, though. That can be fixed with a migration script.

pedrobaeza: As said, this should continue as it was before, with no change. If you want it, then you enable it.

yajo: Could you please elaborate why?

pedrobaeza: I have already told you. It's this way also in 12.0. And even more: this only makes sense if the storage is attachment. I would accept an onchange if you change the storage type to attachment.

yajo: Ah OK, @victoralmau could you please add the onchange?

pedrobaeza (approved):

yajo (approved):

PR #79, opened 2021-05-24
[14.0][FIX] dms: Remove `_get_share_url()` function because is not necessary to override nothing

victoralmau:
Remove `_get_share_url()` function because is not necessary to override nothing (default work fine according to share wizard and auto-add `access_token` in link).

It's need to apply in 13.0 too.

@Tecnativa TT29820

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -197,14 +197,6 @@ class DmsDirectory(models.Model):
             if record.res_id:
                 model.browse(record.res_id).check_access_rule(operation)
 
-    def _get_share_url(self, redirect=False, signup_partner=False, pid=None):
-        self.ensure_one()
-        return "/my/dms/directory/{}?access_token={}&db={}".format(
-            self.id,
-            self._portal_ensure_token(),
-            self.env.cr.dbname,
-        )
-
     def check_access_token(self, access_token=False):
         res = False
         if access_token:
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

pedrobaeza:
One question: shouldn't be the token then populated only if the new argument is true?

chienandalu (approved):

victoralmau:
> One question: shouldn't be the token then populated only if the new argument is true?

IMO is not necessary because actually code use always `access_token` (similar to https://github.com/odoo/odoo/commit/a55f125f5420797a9378925da5626e43d002b9b0#diff-b90ce1ef695264f5890e01458a99eb2b7ab9679df4cfdbec5a0d4c3f156f9e67R54)

pedrobaeza (changes requested):
That's not what I'm reading in the link you have passed me, so please follow the same steps as in sale:

- Gather params in the variables and then call to `encode_params`.
- Don't populate the access token if `share_token` is False.

victoralmau:
According to analysis related to `_get_share_url()` in https://github.com/OCA/contract/pull/680 and test share button, share link and button in mail, my conclusion is that `_get_share_url()` is not necessary (default functionality is correct in `dms` use case).

victoralmau:
It's not necessary now because these changes have apllied in https://github.com/OCA/dms/pull/93#event-4821233586

PR #81, opened 2021-05-25
[12.0][FIX] dms: Remove action column in portal (only used in files) and convert file name to link (download)

victoralmau:
Remove action column in portal (only used in files) and convert file name to link (download).
It's need to change in 13.0 (https://github.com/OCA/dms/pull/80) and 14.0 too.

@Tecnativa TT29848

### DIFF
--- a/dms/views/dms_portal_templates.xml
+++ b/dms/views/dms_portal_templates.xml
@@ -62,7 +62,6 @@
                         <th>Type</th>
                         <th>Size</th>
                         <th>Last update</th>
-                        <th>Actions</th>
                     </tr>
                 </thead>
                 <tbody>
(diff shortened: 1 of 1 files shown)


yajo (approved):

dms/views/dms_portal_templates.xml
```diff
@@ -91,18 +90,23 @@
...
                         <t t-foreach="dms_files" t-as="dms_file">
                             <tr class="tr_dms_file">
                                 <td>
-                                    <img
-                                        t-att-src="image_data_uri(dms_file.thumbnail)"
-                                        style="width:30px"
-                                    />
-                                    <span t-esc="dms_file.name" />
+                                    <a
+                                        t-attf-href="/my/dms/file/#{dms_file.id}/download?{{ keep_query() }}"
```
pedrobaeza, line 100: Side topic (is this download tokenized and thus not able to be dowloaded by anyone?)

pedrobaeza (approved):

dms/views/dms_portal_templates.xml
victoralmau: Link keep parameters (`access_token` for example) and controller decide who can download it (https://github.com/OCA/dms/blob/12.0/dms/controllers/portal.py#L179-L203)

pedrobaeza: Thanks for the info.

PR #84, opened 2021-05-26
[13.0][ADD] dms_version: DMS Version Control

victoralmau:
Superseed: https://github.com/OCA/dms/pull/28
DMS Version Control

imagen_1
imagen_2

@Tecnativa TT29907


LaxmanKalewar:
Unable to install module "dms_version" because an external dependency is not met: Python library not installed: bsdiff4
i install bsdiff4 lib.

please suggest
Thank you.

pedrobaeza:
You haven't installed the library in the correct environment.

dms/tests/test_storage.py
```diff
@@ -3,16 +3,17 @@
 # Copyright 2021 Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
 
-from .common import multi_users
-from .test_storage_database import StorageTestCase
+from .common import DocumentsBaseCase, multi_users
 
 
-class StorageLObjectTestCase(StorageTestCase):
+class TestStorageBase(DocumentsBaseCase):
```
joao-p-marques: Could you include this changes to the `dms` module in a different commit? Still in the same PR, just different commits to make it clearer and organize

dms_version/models/dms_file.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @api.model
+    def _check_context_no_versioning(self):
+        return self.env.context.get("migration") or self.env.context.get("reverting")
+
+    def action_revert_version(self):
+        for record in self:
+            versions = self.env["dms.version"].sudo()
+            content_tuple = versions.pop_version(record)
+            record.with_context(reverting=True).write(
```
joao-p-marques: Could you give a more specific name to this context variable? IMHO this one is too generic...

joao-p-marques (commented):
Code and functional review :+1: 

Just some small comments

victoralmau:
Thanks @joao-p-marques changes done.

dms_version/models/dms_file.py
```diff
@@ -0,0 +1,49 @@
...
+        string="Versions",
+        readonly=True,
+    )
+    count_versions = fields.Integer(
+        compute="_compute_count_versions", string="Count Versions"
+    )
+
+    @api.model
+    def _check_context_no_versioning(self):
+        return self.env.context.get("migration") or self.env.context.get(
```
joao-p-marques: Why is this variable used? 
If it is specific to this module, could you do the same for the other variable?
```suggestion
        return self.env.context.get("dms_migration") or self.env.context.get(
```

victoralmau:
Thanks @joao-p-marques changes done.

joao-p-marques (approved):

dms_version/models/dms_file.py
```diff
@@ -0,0 +1,45 @@
...
+    def action_revert_version(self):
+        for record in self:
+            versions = self.env["dms.version"].sudo()
+            content_tuple = versions.pop_version(record)
+            record.with_context(dms_versioning=True).write(
+                {"content": content_tuple[1], "name": content_tuple[0]}
+            )
+
+    def action_delete_versions(self):
+        self.mapped("version_ids").sudo().unlink()
```
yajo:
```suggestion
        self.mapped("version_ids").unlink()
```

I don't think we should give access to unlink versions to everybody. Just those with permissions.

dms_version/models/dms_file.py
```diff
@@ -0,0 +1,45 @@
...
+        string="Versions",
+        readonly=True,
+    )
+    count_versions = fields.Integer(
+        compute="_compute_count_versions", string="Count Versions"
+    )
+
+    def action_revert_version(self):
+        for record in self:
+            versions = self.env["dms.version"].sudo()
```
yajo: I don't think we should give access to unlink versions to everybody. Just those with permissions.

```suggestion
            versions = self.env["dms.version"]
```

dms_version/models/dms_file.py
```diff
@@ -0,0 +1,45 @@
...
+            content_tuple = versions.pop_version(record)
+            record.with_context(dms_versioning=True).write(
+                {"content": content_tuple[1], "name": content_tuple[0]}
+            )
+
+    def action_delete_versions(self):
+        self.mapped("version_ids").sudo().unlink()
+
+    @api.depends("version_ids")
+    def _compute_count_versions(self):
```
yajo: According to https://www.odoo.com/documentation/13.0/developer/misc/other/guidelines.html#symbols-and-conventions the order of methods should be:

1. Compute
2. CRUD (just write in this case)
3. Action methods

Please reorder the file accordingly.

dms_version/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_dms_version,access_dms_version,model_dms_version,dms.group_dms_user,1,0,0,0
```
yajo: This seems insufficient: nobody is able to write, create or unlink versions.

This maybe worked before because there were too many `sudo()` calls, but please use proper permissions instead.

dms_version/views/dms_file_view.xml
```diff
@@ -0,0 +1,78 @@
...
+    Copyright 2017-2020 MuK IT GmbH
+    License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+  -->
+<odoo>
+    <record id="action_dms_versions_file" model="ir.actions.act_window">
+        <field name="name">Versions</field>
+        <field name="res_model">dms.version</field>
+        <field name="view_mode">tree</field>
+        <field name="domain">[['file_id', '=', active_id]]
+    </field>
```
yajo: This has wrong indentation:

```suggestion
        <field name="domain">[['file_id', '=', active_id]]</field>
```

In any case, I remember problems in the past by relying on `active_id`. 🤔

dms_version/models/dms_file.py
```diff
@@ -0,0 +1,45 @@
...
+
+    version_ids = fields.One2many(
+        comodel_name="dms.version",
+        inverse_name="file_id",
+        string="Versions",
+        readonly=True,
+    )
+    count_versions = fields.Integer(
+        compute="_compute_count_versions", string="Count Versions"
+    )
```
yajo: This is English! 😃 

```suggestion
    versions_count = fields.Integer(
        compute="_compute_versions_count",
        store=True,
    )
```

☝🏼 Besides I added `store=True` because I can see that the result will always be the same, and you're using the field in multirecord views (kanban, tree), so storing it will improve performance.

dms_version/models/dms_file.py
```diff
@@ -0,0 +1,45 @@
...
+            record.count_versions = len(record.version_ids)
+
+    def write(self, vals):
+        if "content" in vals and not self.env.context.get("dms_versioning"):
+            for record in self.filtered("storage_id.has_versioning"):
+                if not record.require_migration:
+                    versions = self.env["dms.version"].sudo()
+                    versions.add_version(record, record.content)
+                    if record.storage_id.clean_versions == "immediately":
+                        versions.clean_versions(record.sudo())
```
yajo: I'm having a hard time understanding this piece of code... 😵 

1. Why do this on `write` and not on `create`?
2. Why are you adding versions *only* when `not self.env.context.get('dms_versioning')`?
3. Why are they always written with `sudo()`?
4. Why do we have a context for indicating versioning *and* a field in `dms.storage` for the same?
5. Under which contexts we really want to skip versioning?
6. What's the purpose of having versions that are cleared immediately?

Please add comments where needed, or fix where not.

dms_version/models/dms_storage.py
```diff
@@ -0,0 +1,65 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class DmsStorage(models.Model):
+    _inherit = "dms.storage"
+
+    has_versioning = fields.Boolean(
+        string="Active Versioning",
```
yajo: I think "Has versioning" is clerarer than "Active Versioning". Besides, it's the same as the field name, so you can remove this string:

```suggestion
```

dms_version/models/dms_storage.py
```diff
@@ -0,0 +1,65 @@
...
+        default=False,
+        help="Indicates if files have an active version control.",
+    )
+    stored_versions = fields.Integer(
+        string="Saved Versions",
+        default=3,
+        help="Indicates how many versions are saved per file.",
+    )
+    clean_versions = fields.Selection(
+        selection=[("immediately", _("On Write")), ("autovacuum", _("On Autovacuum"))],
```
yajo: Never use translations outside a context where there's a `self.env` available. Otherwise there's no place where to get `self.env.lang` from, and the translations do nothing.

OTOH, "On Autovacuum" is a bit too technical for users:

```suggestion
        selection=[("immediately", "Immediately"), ("autovacuum", "Scheduled")],
```

Finally... do we really need something like this? 🤔 Why should a user care about when versions are cleared? Can't we just do it on autovacuum always and add a note in the form explaining why they won't disappear immediately?

If that's not possible, wouldn't it be better a bool field? `clean_old_versions_immediately` -> yes/no

dms_version/models/dms_storage.py
```diff
@@ -0,0 +1,65 @@
...
+
+class DmsStorage(models.Model):
+    _inherit = "dms.storage"
+
+    has_versioning = fields.Boolean(
+        string="Active Versioning",
+        default=False,
+        help="Indicates if files have an active version control.",
+    )
+    stored_versions = fields.Integer(
```
yajo:
```suggestion
    max_versions = fields.Integer(
```

dms_version/models/dms_storage.py
```diff
@@ -0,0 +1,65 @@
...
+    clean_versions = fields.Selection(
+        selection=[("immediately", _("On Write")), ("autovacuum", _("On Autovacuum"))],
+        string="Clean Versions",
+        default="autovacuum",
+    )
+    compress_versions = fields.Boolean(
+        string="Compress Versions",
+        default=False,
+        help="Compress versions to minimize the used space.",
+    )
```
yajo: Wouldn't it be more useful to have a module that compresses *all* `ir.attachment` records?

dms_version/models/dms_storage.py
```diff
@@ -0,0 +1,65 @@
...
+    incremental_versions = fields.Boolean(
+        string="Only Save Changes",
+        default=False,
+        help="Only save the incremental changes between two versions.",
+    )
```
yajo: Are we sure we want to support this? 🤔 

1. What if somebody accidentally removes one middle version?
2. What will happen when your file reaches the versions max. and the oldest version is dropped? Will all the versions chain be recomputed, or will it be broken? none of both scenarios seems nice.

dms_version/models/dms_storage.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def action_clean_file_versions(self):
+        versions = self.env["dms.version"].sudo()
+        files = self.mapped("storage_file_ids").sudo()
+        versions.clean_versions(files)
```
yajo: No sudo please.

dms_version/models/dms_storage.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def action_delete_file_versions(self):
+        self.mapped("storage_file_ids.version_ids").sudo().unlink()
```
yajo: Again, no sudo please.

dms_version/models/dms_storage.py
```diff
@@ -0,0 +1,65 @@
...
+                    _("There amount of saved versions needs to be greater than 0.")
+                )
+
+    @api.constrains("has_versioning", "clean_versions")
+    def _check_clean_versions(self):
+        for record in self:
+            if record.has_versioning and not record.clean_versions:
+                raise ValidationError(
+                    _("A version cleaning setting has to be selected.")
+                )
```
yajo: Please respect recommended method order.
https://www.odoo.com/documentation/13.0/developer/misc/other/guidelines.html#symbols-and-conventions

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+
+import bsdiff4
+
+from odoo import _, api, fields, models, tools
+from odoo.exceptions import ValidationError
+
+
+class DmsVersion(models.Model):
+    _name = "dms.version"
+    _description = "Version"
```
yajo:
```suggestion
    _description = "DMS Version"
```

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    incremental_versions = fields.Boolean(
+        related="storage_id.incremental_versions", readonly=True
+    )
+    previous_version = fields.Many2one(
+        comodel_name="dms.version", ondelete="set null", string="Previous Version"
+    )
+    next_version = fields.Many2one(
+        comodel_name="dms.version", ondelete="set null", string="Next Version"
+    )
```
yajo, line 35: When you have a versions recordset, properly ordered, are these fields really needed?

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+    is_compress = fields.Boolean(string="Compressed", default=False)
+    is_incremental = fields.Boolean(string="Incremental", default=False)
```
yajo: Why do we need this? Can't it be obtained from the storage setting?

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+    size = fields.Integer(string="Size", readonly=True)
```
yajo, line 36: You can get size from attachments when you read the field with `bin_size=True` in context, so you don't need this field.

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+                "binary": base64.b64encode(binary),
+                "size": binary and len(binary) or 0,
+                "is_incremental": is_incremental,
+                "is_compress": is_compress,
+            }
+            updates[tools.frozendict(values)].add(record.id)
+        with self.env.norecompute():
+            for vals, ids in updates.items():
+                self.browse(ids).write(dict(vals))
+        self.recompute()
```
yajo: Please reorder methods.

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+        if self.is_compress:
+            binary = zlib.decompress(binary)
+        if self.is_incremental and patch:
+            prev_version = self.previous_version
+            prev_binary = prev_version.get_binary()
+            binary = bsdiff4.patch(prev_binary, binary)
+        return binary
+
+    @api.model
+    def clean_versions(self, dms_files, max_size=None):
```
yajo: Make methods private unless there's a button on UI to use them.

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+        for dms_file in dms_files.filtered(lambda rec: rec.count_versions > 0):
+            max_size = max_size or dms_file.storage_id.stored_versions
```
yajo: Is there really any situation where we want a different max_size from the one declared in the storage?

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+            if dms_file.count_versions > max_size:
+                remove_size = dms_file.count_versions - max_size
+                remove_versions = self.env["dms.version"]
+                current = dms_file.version_ids.filtered(
+                    lambda rec: not rec.previous_version
+                )
+                while len(remove_versions) < remove_size:
+                    remove_versions |= current
+                    current = current.next_version
```
yajo: If records are properly sorted, this is much simpler:

```suggestion
                remove_versions = dms_file.version_ids[remove_size:]
                current = dms_file.version_ids[remove_size - 1]
```

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+                    {
+                        "content": current.content,
+                        "previous_version": None,
+                        "is_incremental": False,
+                    }
+                )
+                remove_versions.unlink()
+
+    @api.model
+    def add_version(self, dms_file, content):
```
yajo: Make it private unless there's a need for it to be public.

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+        record = self.search(domain, limit=1)
+        if not record:
+            return self.create(vals)
+        vals["previous_version"] = record.id
+        new_version = self.create(vals)
+        record.write({"next_version": new_version.id})
+        return new_version
+
+    @api.model
+    def pop_version(self, dms_file):
```
yajo: Private unless needed.

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                )
+                remove_versions.unlink()
+
+    @api.model
+    def add_version(self, dms_file, content):
+        domain = [
+            ("file_id", "=", dms_file.id),
+            ("next_version", "=", False),
+        ]
```
yajo: It doesn't seem to make sense to use an `@api.model` method for something that expects a record. Instead:

```suggestion
    def add_version(self, content):
        domain = [
            ("file_id", "=", self.id),
            ("next_version", "=", False),
        ]
```

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+        new_version = self.create(vals)
+        record.write({"next_version": new_version.id})
+        return new_version
+
+    @api.model
+    def pop_version(self, dms_file):
+        domain = [
+            ("file_id", "=", dms_file.id),
+            ("next_version", "=", False),
+        ]
```
yajo: Use a multi method:

```suggestion
    def pop_version(self):
        domain = [
            ("file_id", "=", self.id),
            ("next_version", "=", False),
        ]
```

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+        record = self.search(domain, limit=1)
+        if not record:
+            return (None, None)
+        result = (record.name, record.content)
+        record.unlink()
+        return result
+
+    @api.depends("binary")
```
yajo, line 50: You need to add a `depends_context` too

dms_version/models/dms_version.py
```diff
@@ -0,0 +1,160 @@
...
+        attachment=False,
+        prefetch=False,
+        required=True,
+        store=False,
+    )
+    binary = fields.Binary(
+        string="Binary", prefetch=False, attachment=True, invisible=True
+    )
+
+    def get_binary(self, patch=True):
```
yajo: Private method please.

And remove the `patch` argument, as it seems there's no use case for it to be false.

dms_version/views/dms_version_view.xml
```diff
@@ -0,0 +1,21 @@
...
+<odoo>
+    <record id="view_dms_version_tree" model="ir.ui.view">
+        <field name="name">dms_version.tree</field>
+        <field name="model">dms.version</field>
+        <field name="arch" type="xml">
+            <tree default_order="create_date desc">
+                <field name="create_date" />
+                <field name="name" string="Filename" invisible="1" />
+                <field name="content" widget="binary" filename="name" />
+                <field name="size" widget="binary_size" />
```
yajo: Possibly it should be this:

```suggestion
                <field name="content" widget="binary_size" string="Size" />
```

yajo (changes requested):

dms_version/models/dms_version.py
victoralmau: I'm not really sure because in `dms.file` `size` field is necessary too.

victoralmau:
Still in progress, sorry.
I will continue soon.

yajo:
OK, no problem. Take it out of draft when ready to review again please 😊

dms_version/views/dms_file_view.xml
victoralmau: This domain has been removed and the active_id is used in the context.

dms_version/models/dms_file.py
victoralmau: All this code has been simplified and no special comments are needed now.

dms_version/models/dms_storage.py
victoralmau: OK, I agree, I think this complexity is not necessary.

dms_version/models/dms_storage.py
victoralmau: Ok, I find it better to remove this part and add it to ROADMAP as a separate future addon.

victoralmau:
All changes are made and I think it is ready for a new review.

joao-p-marques (approved):
Code review :+1:

victoralmau:
Sorry but I'm going to draft it again.
Currently you cannot update the contents of a file because you do not have permissions to create a record in `dms.version`.

In addition, I think another different approach would be much more optimal, and that is to use `dms.file` records with some new field (`is_version = True` or `parent_id`) to use the versions and thus be able to reuse the existing views.

victoralmau:
Superseed by: https://github.com/OCA/dms/pull/177

PR #85, opened 2021-05-26
[12.0][FIX] dms: Fix error in portal with access_token

victoralmau:
Fix error in portal with access_token

It's need to apply in 13.0 and 14.0

@Tecnativa TT29923

### DIFF
--- a/dms/controllers/portal.py
+++ b/dms/controllers/portal.py
@@ -1,32 +1,24 @@
-# Copyright 2020 Tecnativa - Víctor Martínez
+# Copyright 2020-2021 Tecnativa - Víctor Martínez
 import base64
-import logging
 
 from odoo import _, http
-from odoo.exceptions import AccessError, MissingError
 from odoo.http import request
 from odoo.osv.expression import OR
 
 from odoo.addons.portal.controllers.portal import CustomerPortal
 from odoo.addons.web.controllers.main import content_disposition, ensure_db
 
-_logger = logging.getLogger(__name__)
-
 
 class CustomerPortal(CustomerPortal):
     def _dms_check_access(self, model, res_id, access_token=None):
-        try:
-            item = request.env[model].browse(res_id)
-        except (AccessError, MissingError):
-            return False
-
         if access_token:
+            item = request.env[model].sudo().browse(res_id)
             if not item.check_access_token(access_token):
                 return False
         else:
+            item = request.env[model].browse(res_id)
             if not item.sudo(request.env.user.id).check_access("read", False):
                 return False
-
         return item
 
     def _prepare_portal_layout_values(self):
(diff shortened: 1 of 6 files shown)


dms/controllers/portal.py
```diff
@@ -1,32 +1,24 @@
...
         if access_token:
+            item = request.env[model].sudo().browse(res_id)
             if not item.check_access_token(access_token):
                 return False
         else:
+            item = request.env[model].browse(res_id)
             if not item.sudo(request.env.user.id).check_access("read", False):
                 return False
-
         return item
```
yajo, line 22: Why https://github.com/odoo/odoo/blob/8703ff1e3d9be6f2f5fce2e8c4e62589b05133fb/addons/portal/controllers/portal.py#L233-L244 is not enough?

dms/controllers/portal.py
```diff
@@ -124,11 +116,16 @@ def portal_my_dms_directory(
...
-            .search(domain, order=sort_br)
-        )
+        if access_token:
+            dms_directory_items = (
+                request.env["dms.directory"].sudo().search(domain, order=sort_br)
+            )
+        else:
+            dms_directory_items = (
+                request.env["dms.directory"]
+                .sudo(request.env.user.id)
```
yajo: This seems redundant, isn't it?

```suggestion
```

dms/controllers/portal.py
victoralmau: `_document_check_access` function is not enough for some reasons:
- Odoo allways return `document_sudo` and it's totally incorrect (in dms, if we use inherit permissions is important NOT apply `.sudo()` or we will show more records of which you should, all sales for example).
- `access_token` is checked allways with model but in dms is necessary to check with directory (if we try to access to `dms.file` model) or check parent directories (if we try to access to subdirectories).

dms/controllers/portal.py
victoralmau: It's very important and related to https://github.com/OCA/dms/pull/85#discussion_r640515228 (with `access_token` is necessary to use `.sudo()` but if not use access_token is very important to apply `request.env.user.id`).

dms/controllers/portal.py
yajo: > Odoo allways return `document_sudo` and it's totally incorrect (in dms, if we use inherit permissions is important NOT apply `.sudo()` or we will show more records of which you should, all sales for example).

but then, what happens if you share a tokenized URL of a directory that inherits permissions? 🤔 Is that a way to surpass permissions?



> `access_token` is checked allways with model but in dms is necessary to check with directory (if we try to access to `dms.file` model) or check parent directories (if we try to access to subdirectories).

OK that makes sense.

dms/controllers/portal.py
yajo: Yes, I mean that it's already applied. `request.env["dms.directory"]` already has `request.env.user.id`, so this line is redundant.

victoralmau: Ok thanks, changes done.

dms/controllers/portal.py
victoralmau: What happen if you share with somebody `share_link` in Google Drive for example? It will be possible to access all the files (and folders) that are inside that folder but never that are above it. Right? Any other idea about it?

yajo: I understand now, thanks 😄

yajo (approved):

Issue #86, opened 2021-05-27
Record Rules Bug


gdgellatly:
Hi,

I've spent a couple of hours and can't seem to track this down.

Have a setup of say
Root -> Dir A -> Subdir x
         -> Dir B -> Subdir y

So A and B are root directories and they have Security Group A and B with read access respectively as well as an admin group with CRUD. inherit groups is set on subdir x and y.

Security group A and B each contain 1 user - User A, and User B. 
Both users have Documents/User access rights.

Clicking the Documents Tile as one of those users - Record Rule error - read on Directory
Cut and paste a known good link to the files kanban - Record Rule error - read on Directory
Cut and paste a known good link to the directories kanban - works as expected.

victoralmau:
Hi,

Reviewing what you indicate it seems that there is an error following your steps.
I have created a PR (https://github.com/OCA/dms/pull/87) that will solve this problem.

victoralmau:
Now, it's solved in 13.0 and 14.0.

PR #87, opened 2021-05-27
[13.0][FIX] dms: Apply _apply_access_groups correctly (specially according to files) to get only records to get only allowed records according to directories groups

victoralmau:
Apply `_apply_access_groups` correctly (specially according to files) to get only records to get only allowed records according to directories groups.

Related to https://github.com/OCA/dms/issues/86

Steps to reproduce:

- Create user called: usera and set group: Documents > User
- Create Access group called "Group A" and set "Read access" and usera to explicit users.
- Create Directory called "Root directory A" and set groups: Admin + Group A
- Create Subdirectory X accordint to "Root directory a"
- Login as usera and go to "Documents"

Before get access error because try to get records (files) not allowed, now only get allowed records (files).

It's need to apply in 14.0

@Tecnativa TT30074

### DIFF
--- a/dms/models/dms_security_mixin.py
+++ b/dms/models/dms_security_mixin.py
@@ -104,24 +104,27 @@ class DmsSecurityMixin(models.AbstractModel):
         if self.env.user.has_group("base.group_public"):
             return None
 
+        field = "id"
+        if self._name == "dms.file":
+            field = "directory_id"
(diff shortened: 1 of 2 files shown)


etobella (approved):
Make sense to me.
technical review

pedrobaeza (approved):

PR #88, opened 2021-05-28
[FIX] dms: image thumbnails performance

yajo:
- Portal was too slow to load because every thumbnail came base64-encoded. Instead, now it loads resources externally, making the initial request much faster, and benefiting from caching.
- Fields `custom_thumbnail_medium` and `custom_thumbnail_small` are removed because they were mostly useless.
- Fields `thumbnail`, `thumbnail_medium` and `thumbnail_small` are now stored because otherwise the server had to recompute the resizing on each request, making it actually slower than having no thumbnails at all.
- Remove unused `size` parameter in thumbnail internal getters.
- All thumbnail fields were getting resized to sizes that were different from the sizes that, later, were actually used in UI, thus making the UI always re-resize them. Removed that nonsense. Thumbnails now use the same size that the UI will require, and then UI needs to ask for no resizes.
- `thumbnail` field is now actually resized, if needed.
- Typo fixed in unknown file default thumbnail.
- Do your customers really know what a mime type is? I bet not. Remove "type" column from portal. Thumbnails provide all the required information. If you're a geek, you can still hover over the thumbnail and get the mime type.

@Tecnativa TT29847 @victoralmau @pedrobaeza


victoralmau (approved):

victoralmau:
Sorry, travis is red.

dms/models/mixins_thumbnail.py
```diff
@@ -14,73 +14,53 @@ class Thumbnail(models.AbstractModel):
...
-    )
-    custom_thumbnail_small = fields.Binary(
-        "Small Custom Thumbnail", compute="_compute_custom_thumbnail", prefetch=False,
-    )
-
-    thumbnail = fields.Binary(
-        compute="_compute_thumbnail", string="Thumbnail", prefetch=False,
-    )
+    custom_thumbnail = fields.Binary(string="Custom Thumbnail", attachment=False)
+    thumbnail = fields.Binary(compute="_compute_thumbnail", store=True)
```
pedrobaeza: This change may be the hell on existing databases when updating the module. Why the `prefetch=False` is not working?

yajo: [`prefetch=False` is the default for binary fields](https://github.com/odoo/odoo/blob/18c6e059180367998f1c46efa0dc4697df67f0f3/odoo/fields.py#L1852), so it's not needed to declare it.

pedrobaeza: OK, but why is this not working then?

yajo:
Travis should get ✅ now.

dms/models/mixins_thumbnail.py
pedrobaeza: I insist that this change is no good, and I prefer to not do it.

yajo: The main change is to make it `store=True`. It is important to do it because thumbnails exist to speed up things, and having to go through batch image conversion on each request kills that purpose.

That's how odoo thumbnails work elsewhere AFAIK.

pedrobaeza: This is still no good. These fields shouldn't be computed unless they are used, so the problem is there. Please don't store it.

yajo: If I don't store them, then I have to remove them.

The only purpose for a thumbnail's existence is to generate a cache and speed things up, and that's why everywhere else Odoo stores computed thumbnails.

In fact, you can see that [in odoo 13.0 we have the `image.mixin` class that implements exactly this same behavior](https://github.com/odoo/odoo/blob/6a7db37d5cc6316cf51bf9b1eea5dcee0aebe97c/odoo/addons/base/models/image_mixin.py#L15-L19).

If Odoo needs to resize images on each request, things become slow. If there are 80 thumbnails in a screen, then it becomes a bottleneck.

So, I'll proceed to remove all these fields.

pedrobaeza: No, don't remove them, but as said, if they are not used, they shouldn't be computed. Why are they computed? You are not answering my question since the beginning, and that's the key.

yajo: Well, you didn't ask any questions. You just said "this is hell", "this is no good", etc. _That_ is the key.

Now regarding why they are computed: because they are used.

yajo: I can stop using them instead of removing them, if you prefer. However they'll still be garbage anyway. 🤷🏼‍♂️

pedrobaeza: You aren't still answering me where they are used.

yajo:
Well, the implementation changed a bit, so please update your reviews.

- The only fields that are now `store=True, attachment=True` are `custom_thumbnail_[medium,small]`, which were removed in the previous implementation. These fields, however, will only store thumbnails when `custom_thumbnail` is set by hand by the user. The purpose of having them stored is to produce a cache that speeds up things.
- A new `icon_url` field is added. This one is computed on the fly but very cheap, so I think it's not worth it to make it `store=True`. It just returns a URL pointing to the static file inside `/dms/static`, so all files/folders sharing the same icon will only ask once. This saves tons of requests and results in all views being really much more snappy. The counterpart is that it's not resized, but some analysis proved that the wins are much more than the losses in that regard.
- That new field is used to avoid consulting `thumbnail[,_medium,_small]` fields in list/kanban/portal views. This avoids costly recomputations (although they're much lighter now) and makes views render quickly.

TODO:


- [x] Migration script

dms/models/mixins_thumbnail.py
```diff
@@ -14,73 +14,82 @@ class Thumbnail(models.AbstractModel):
     _name = "dms.mixins.thumbnail"
     _description = "Thumbnail Mixin"
 
-    custom_thumbnail = fields.Binary(
-        string="Custom Thumbnail", attachment=False, prefetch=False,
-    )
+    custom_thumbnail = fields.Binary(string="Custom Thumbnail", attachment=False)
```
pedrobaeza, line 17: Why not being `attachment=True`?

dms/models/mixins_thumbnail.py
```diff
@@ -14,73 +14,82 @@ class Thumbnail(models.AbstractModel):
...
-    )
-
-    thumbnail_small = fields.Binary(
-        "SmallThumbnail", compute="_compute_thumbnail", prefetch=False,
+        "Small Custom Thumbnail",
+        attachment=True,
+        compute="_compute_custom_thumbnail",
+        store=True,
     )
+    thumbnail = fields.Binary(compute="_compute_thumbnail", string="Thumbnail",)
```
pedrobaeza, line 30: Does these fields makes sense to be preserved now?

dms/views/dms_portal_templates.xml
```diff
@@ -102,15 +100,12 @@
                                         t-att-title="dms_file.name"
                                     >
                                         <img
-                                            t-att-src="image_data_uri(dms_file.thumbnail)"
-                                            style="width:30px"
+                                            class="o_portal_contact_img"
+                                            t-att-src="dms_file.icon_url"
```
pedrobaeza, line 104: No `t-if` here?

dms/models/mixins_thumbnail.py
yajo: This makes no sense, but it's how it was before, and since it's not the root of the performance problem, I decided to not touch it.

dms/views/dms_portal_templates.xml
yajo: No, I think that the icon is just enough for all cases in the portal. The thumbnail doesn't seem very useful there.

dms/models/mixins_thumbnail.py
yajo: Not all are used, but the ones that are, are used now only in form views, where records are computed 1 by 1 and thus there's no performance bottleneck.

Since they are not a deep problem in this new fix, I decided to keep them, just to keep the model more stable.

etobella (approved):
Functional Review

victoralmau:
Please @Yajo can you apply these changes in 13.0 and 14.0 too?

PR #92, opened 2021-05-31
[13.0][FIX] dms: Improve tests (prevent error if installed other addons after dms install and new addons auto-create partners)

victoralmau:
Improve tests to prevent error if installed other addons after dms install and new addons auto-create partners..

Please @joao-p-marques and @pedrobaeza can you review it?

It's need to apply in 14.0

@Tecnativa TT30121


joao-p-marques (approved):

pedrobaeza (approved):

pedrobaeza:
Please next time document more the change. Not only "Improve test". Improve in what? Let's not block it for now anyways:

/ocabot merge patch

PR #97, opened 2021-06-02
[13.0][FIX] dms: image thumbnails performance

victoralmau:
FWP from 12.0: https://github.com/OCA/dms/pull/88

Please @pedrobaeza and @Yajo  can you review it?

@Tecnativa TT29847


etobella:
Something is wrong with the manifest, can you rebase?

victoralmau:
It's ready to merge now.

dms/migrations/13.0.6.0.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    for model in ("dms.file", "dms.directory"):
+        _logger.info("Computing stored custom thumbnails for model %s", model)
+        records = env[model].search([("custom_thumbnail", "!=", False)])
+        records._compute_custom_thumbnail()
```
pedrobaeza: I don't think this is needed, as previously it was not stored, so it will be stored on regular update mechanism.

dms/models/mixins_thumbnail.py
```diff
@@ -14,79 +14,86 @@ class Thumbnail(models.AbstractModel):
...
         folders = ["static", "lib", "img", "thumbnails"]
         path = get_resource_path("dms", *folders, name)
         if not os.path.isfile(path):
-            path = get_resource_path("dms", *folders, "file_unkown.svg")
+            path = get_resource_path("dms", *folders, "file_unknown.png")
         return path
 
     def _get_thumbnail_placeholder_name(self):
-        return "folder.svg"
+        return "folder.png"
```
pedrobaeza: Why changing this to PNG? SVG should be the standard.

victoralmau: Sorry, it's a mistake when apply commits from 12.0.

pedrobaeza (approved):

yajo (approved):

yajo (changes requested):
I don't see the migration script. Is it not needed in this version?

victoralmau:
> I don't see the migration script. Is it not needed in this version?

I deleted it according to this comment https://github.com/OCA/dms/pull/97#discussion_r656152821

yajo (approved):
OK, sorry I missed that part.

dms/models/mixins_thumbnail.py
```diff
@@ -14,79 +14,86 @@ class Thumbnail(models.AbstractModel):
...
+            record.update(
+                {
+                    "custom_thumbnail_medium": tools.image_process(
+                        base64_source, size=(128, 128)
+                    ),
+                    "custom_thumbnail_small": tools.image_process(
+                        base64_source, size=(64, 64)
+                    ),
+                }
+            )
```
yajo: Probably it would be better to remove this function and use the max_width and max_height field attributes as they were being used (just modifying values). That's a new feature that didn't exist in v12 and thus I wasn't able to do it that way.

~~Please push that separately so it can be easily analyzed apart from this cherry-pick.~~

yajo (changes requested):

yajo (approved):
Code & functional review.

PR #58, opened 2021-02-14
[12.0][ADD] URL Attachment

gabrielcardoso21:
This Pull Request make it possible to create URL attachments as documents in DMS.

PR #98, opened 2021-06-03
[12.0][IMP] dms: Add URL Attachment

victoralmau:
Add URL Attachment

Superseed: https://github.com/OCA/dms/pull/58

@Tecnativa TT30215

### DIFF
--- a/dms/demo/file.xml
+++ b/dms/demo/file.xml
@@ -238,4 +238,11 @@
         <field name="directory_id" ref="dms.directory_02_demo" />
         <field name="content" type="base64" file="dms/demo/data/text.rst" />
     </record>
+    <record id="file_37_demo" model="dms.file">
+        <field name="name">OCA</field>
+        <field name="color" eval="3" />
+        <field name="directory_id" ref="dms.directory_11_demo" />
+        <field name="type">url</field>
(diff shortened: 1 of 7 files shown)


pedrobaeza:
In my opinion we are mixing concepts: documents shouldn't be URLs. They are very different things. It's not the same as attachments, where you attach something, which can be a link or a document.

This shouldn't be merged IMO.

qtiplb:
Hi @pedrobaeza @victoralmau 

I found this PR while looking exactly for this functionality. We also want to add this to v12 and v14.

The goal is to be able to organize documents (more specifically ir.attachments) in the DMS, regardless of their storage location. I feel this a must for anyone not storing all their documents in Odoo (directly in the filestore or through an object storage).

I would appreciate if you could reconsider so it can stay under the OCA umbrella.

Thanks!

pedrobaeza:
You can extract it to a new module, but not in the main module.

qtiplb:
Alright we'll extract it, do you want us to submit it to OCA/dms ?

pedrobaeza:
Yes, it can be hosted here. The main issue is to maintain such extra code in the main module.

PR #99, opened 2021-06-04
[FIX] dms: missing migration parameter

yajo:
Bug:

    TypeError: migrate() takes 1 positional argument but 2 were given

@Tecnativa TT29847


pedrobaeza (approved):

pedrobaeza:
Direct merging as no need of extra CIs

PR #102, opened 2021-06-18
[REF] dms: modify permissions computation

yajo:
The intention of this PR was to implement permissions in a different way that avoids the constant security problems, bottlenecks, hacks and recursiveness problems we face constantly in `dms`.

However, I haven't been able to actually make it work. It must not be too far from a working status, but this is what I have so far.

Right now, the main problem is that, when you create a new `dms.directory` with proper permissions, it fails because, when checking the related groups, probably the m2m field that links directories with `dms.access.groups` has not been created yet.

Details follow:

@Tecnativa TT29847

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -651,7 +651,14 @@ class DmsDirectory(models.Model):
                 parent = self.browse([vals["parent_id"]])
                 data = next(iter(parent.sudo().read(["storage_id"])), {})
                 vals["storage_id"] = self._convert_to_write(data).get("storage_id")
-        return super().create(vals_list)
+        # Create as sudo to avoid testing creation permissions before DMS security
+        # groups are attached (otherwise nobody would be able to create)
+        res = super(DmsDirectory, self.sudo()).create(vals_list)
+        # Go back to original user and check we really had creation permission
+        res = res.sudo(self.env.user)
+        res.check_access_rights("create")
+        res.check_access_rule("create")
+        return res
 
     def write(self, vals):
         # Groups part
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -4,13 +4,13 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import base64
-import functools
 import logging
-import operator
 from collections import defaultdict
 
 from odoo import _, api, fields, models, tools
-from odoo.exceptions import AccessError, ValidationError
+from odoo.exceptions import ValidationError
+from odoo.osv.expression import OR
+from odoo.tools import consteq
 
 from odoo.addons.http_routing.models.ir_http import slugify
 
@@ -38,6 +38,7 @@ class DmsDirectory(models.Model):
 
     _parent_store = True
     _parent_name = "parent_id"
+    _directory_field = _parent_name
 
     name = fields.Char(string="Name", required=True, index=True)
 
@@ -64,18 +65,37 @@ class DmsDirectory(models.Model):
         ondelete="restrict",
         auto_join=True,
         store=True,
+        compute_sudo=True,
     )
-
     parent_id = fields.Many2one(
         comodel_name="dms.directory",
-        domain="[('permission_create', '=', True)]",
         string="Parent Directory",
+        domain="[('permission_create', '=', True)]",
         ondelete="restrict",
-        auto_join=True,
         index=True,
         copy=True,
+        # Access to a directory doesn't necessarily mean access its parent, so
+        # prefetching this field could lead to misleading access errors
+        prefetch=False,
     )
-
+    group_ids = fields.Many2many(
+        comodel_name="dms.access.group",
+        relation="dms_directory_groups_rel",
+        column1="aid",
+        column2="gid",
+        string="Groups",
+    )
+    complete_group_ids = fields.Many2many(
+        comodel_name="dms.access.group",
+        relation="dms_directory_complete_groups_rel",
+        column1="aid",
+        column2="gid",
+        string="Complete Groups",
+        compute="_compute_groups",
+        readonly=True,
+        store=True,
+        compute_sudo=True,
+    )
     complete_name = fields.Char(
         "Complete Name", compute="_compute_complete_name", store=True
     )
@@ -87,10 +107,7 @@ class DmsDirectory(models.Model):
         copy=False,
     )
     is_hidden = fields.Boolean(
-        string="Storage is Hidden",
-        related="storage_id.is_hidden",
-        readonly=True,
-        search="_search_is_hidden",
+        string="Storage is Hidden", related="storage_id.is_hidden", readonly=True,
     )
     company_id = fields.Many2one(
         related="storage_id.company_id",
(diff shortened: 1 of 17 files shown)


victoralmau (changes requested):
Please, check Travis (IMO some errors are related to need to add groups to directories in some tests or demo data).

victoralmau (approved):

dms/models/directory.py
```diff
@@ -277,10 +277,11 @@ def _get_parent_categories(self, access_token):
...
 
     def _get_own_root_directories(self):
-        items = self.env["dms.directory"].search(
-            [("is_hidden", "=", False), ("parent_id", "=", False)]
+        return (
+            self.env["dms.directory"]
+            .search([("is_hidden", "=", False), ("parent_id", "=", False)])
+            .ids
         )
-        return items.ids
```
yajo: AFAICS this is the same as before, right?

Then, please, undo the change. Try to change only relevant stuff, so there's less to review and less chances to introduce more bugs. (This is a general rule of thumb).

dms/models/directory.py
```diff
@@ -691,7 +650,9 @@ def create(self, vals_list):
                 parent = self.browse([vals["parent_id"]])
                 data = next(iter(parent.sudo().read(["storage_id"])), {})
                 vals["storage_id"] = self._convert_to_write(data).get("storage_id")
-        return super().create(vals_list)
+        res = super(DmsDirectory, self.sudo()).create(vals_list)
+        res.invalidate_cache()
+        return res
```
yajo: This has some bugs, and also makes me wonder some things.

1. You're returning a sudoed recordset. That's *very* dangerous.
2. When you resudo it back to the current user, probably the cache will be invalidated automatically (because the cache is based on its env, and sudo creates a new env). However, don't trust me and make sure that's how it works.
3. If I'm wrong in the previous point and you really need to invalidate cache, then you also need to put a comment explaining why that's necessary. Also, you should pass a list of fields to invalidate, because only the permission fields are relevant for this use case.
4. You can't just use sudo and go on without putting a comment explaining why you're doing what you're doing. Be generous on comments, please (actually lack of comments made this refactor very hard 😅).
5. At some point, you really need to check the user has creation access.

So, probably the fix would be close to this, but please double-check it:

```suggestion
        # Create as sudo to avoid testing creation permissions before DMS security
        # groups are attached (otherwise nobody would be able to create)
        res = super(DmsDirectory, self.sudo()).create(vals_list)
        # Go back to original user and check we really had creation permission
        res = res.sudo(self.env.user)
        res.check_access_rights("create")
        res.check_access_rule("create")
        return res
```

yajo (commented):
👍🏼 Tested functionally and code review.

dms/models/directory.py
```diff
@@ -64,18 +65,35 @@ class DmsDirectory(models.Model):
...
     )
-
     parent_id = fields.Many2one(
         comodel_name="dms.directory",
-        domain="[('permission_create', '=', True)]",
         string="Parent Directory",
+        domain="[('permission_create', '=', True)]",
         ondelete="restrict",
-        auto_join=True,
+        prefetch=False,
```
pedrobaeza: Why this? I think it requires an explanation as comment on code.

dms/security/ir.model.access.csv
```diff
@@ -3,6 +3,7 @@ id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
 access_dms_tag_user,dms_tag_user,model_dms_tag,group_dms_user,1,1,1,1
 access_dms_category_user,dms_category_user,model_dms_category,group_dms_user,1,1,1,1
 
+access_dms_storage_portal,dms_storage_portal,model_dms_storage,base.group_portal,1,0,0,0
```
pedrobaeza, line 6: Can we avoid this permission for not exposing possible credentials stored on this model to everyone?

dms/security/security.xml
```diff
@@ -48,6 +48,19 @@
...
+        <field name="model_id" ref="model_dms_file" />
+        <field name="groups" eval="[(4, ref('base.group_user'))]" />
+        <field name="global" eval="True" />
+        <field name="perm_read" eval="0" />
+        <field name="perm_create" eval="1" />
+        <field name="perm_write" eval="1" />
+        <field name="perm_unlink" eval="1" />
+        <field
+            name="domain_force"
+        >['|', ('locked_by', '=', False), ('locked_by', '=', user.id)]</field>
```
pedrobaeza, line 62: What about using the newly discovered?

```suggestion
        >[('locked_by', '=?', user.id)]</field>
```

dms/models/directory.py
yajo: Yes, it's because having access to a directory doesn't necessarily grant access to its parent, so while prefetching this you could get an unnecessary access error. I'll add the comment, it's really not obvious.

dms/security/ir.model.access.csv
yajo: No, sorry. To be able to filter record access using `ir.rule`, before you need model access from an `ir.model.access`. This line is essential for making permissions on this module behave more like Odoo permissions, which is an essential part of the patch.

However, you can see that in `security.xml` global rules are added that imply searching the corresponding computed fields, which produce the correct domain for restricting record access.

pedrobaeza: Imagine a new module `dms_google_drive`, the usual move is to add to `dms.storage` the fields `Client secret` and `API key`. With this permission, you are giving access to any portal user to this secret information, and record rules doesn't serve, as they give whole access to the record. I think we can access to this via `sudo`, no matter other done tricks.

yajo: Ah I see, sorry, I was confused in my previous answer and thought you were referring to access to `dms.file` or `dms.directory`. I see now that it refers to `dms.storage`.

In the case you are proposing, that hypothetical module could restrict field access [using the `groups` argument](https://www.odoo.com/documentation/12.0/developer/reference/security.html#field-access).

I'll try to remove it and report anyways, as it seems a good idea in any case.

dms/security/security.xml
yajo: After testing, I'll have to remove this one, because it doesn't actually mean the same.

The original domain means "locked_by can be false or user.id", but the new one means "if user.id is False, we don't filter by locked_by; but if it has a value, we have to filter by it".

dms/security/ir.model.access.csv
yajo: I pushed https://github.com/OCA/dms/pull/102/commits/dd68406f7149fffcfca33c120ae01f3aa3112968 to forbid portal and basic users reading hidden `dms.storage` records. That, combined with the suggestion of filtering fields by groups, should be enough for that use case.

I've spent 2h trying to remove this line, but this is the only solution that worked without breaking all tests... Is it OK or should I spend more time on it?

pedrobaeza: OK, just include a comment in the roadmap and it's good to go.

pedrobaeza:
Travis is red right now.

dms/security/security.xml
pedrobaeza: I see. Indeed it's not the same. The operator servers for other purpose.

yajo:
Thanks, it should be all good now.

pedrobaeza (approved):

PR #100, opened 2021-06-08
[FIX] dms: cascade removal of related files/directories

yajo:
When some dms.directory or dms.file was related to a real DB record (a.k.a. attachment storage), if the record got deleted, the DMS mirror was still alive.

If the dms.storage had inherit_access_from_parent_record=True, the situation got worse because the dms.directory became visible to everyone.

@Tecnativa TT30223

PR #103, opened 2021-06-21
[13.0][FIX] dms: cascade removal of related files/directories

victoralmau:
FWP from 12.0: https://github.com/OCA/dms/pull/100

When some dms.directory or dms.file was related to a real DB record (a.k.a. attachment storage), if the record got deleted, the DMS mirror was still alive.

If the dms.storage had inherit_access_from_parent_record=True, the situation got worse because the dms.directory became visible to everyone.

@Tecnativa TT30223

### DIFF
--- a/dms/models/__init__.py
+++ b/dms/models/__init__.py
@@ -1,4 +1,5 @@
 from . import access_groups
+from . import base
 from . import mixins_thumbnail
 from . import dms_security_mixin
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as it's a fw-port:

/ocabot merge patch

Issue #105, opened 2021-06-22
[13.0] dms.file tree order


rbellanova:
Hi dear, I have found that sorting of columns in the file tree view does not work

![Peek 22-06-2021 11-47](https://user-images.githubusercontent.com/42999048/122903350-9cb64480-d34f-11eb-9157-3705b725a5f5.gif)

rbellanova:
If it could be useful... problably the problem is on "_search" function

rbellanova:
I solved in this way, removing the set operations in favor of recordset operations
Hope it could be useful

`
        file_ids = self.browse(result)
        directories = self._get_directories_from_database(result)
        for directory in directories - directories._filter_access("read"):
            file_ids -= directory.sudo().mapped("file_ids").ids
        return len(file_ids) if count else file_ids.ids`

victoralmau:
> I solved in this way, removing the set operations in favor of recordset operations
> Hope it could be useful
> 
> ` file_ids = self.browse(result) directories = self._get_directories_from_database(result) for directory in directories - directories._filter_access("read"): file_ids -= directory.sudo().mapped("file_ids").ids return len(file_ids) if count else file_ids.ids`

Thanks, i have opened some change (https://github.com/OCA/dms/pull/106) to solve it.

victoralmau:
Hi, it's solved on 13.0 and 14.0

PR #106, opened 2021-06-22
[13.0][FIX] dms: Remove _search() function in files

victoralmau:
Related to: https://github.com/OCA/dms/issues/105

Keep order in search function.

Is necessary to change it in 14.0 too.

Locked by:
- [x] https://github.com/OCA/dms/pull/107

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa

### DIFF
--- a/dms/i18n/es.po
+++ b/dms/i18n/es.po
@@ -252,12 +252,6 @@ msgid "A directory with the same name already exists."
 msgstr "Una carpeta con el mismo nombre ya existe."
 
 #. module: dms
-#: code:addons/dms/models/directory.py:0
-#, python-format
-msgid "A file is locked, the folder cannot be deleted."
-msgstr "El archivo está bloqueado, la carpeta no se puede eliminar."
-
-#. module: dms
 #: code:addons/dms/models/dms_file.py:0
 #, python-format
 msgid "A file must have model and resource ID in attachment storage."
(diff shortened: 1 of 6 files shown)


dms/models/dms_file.py
```diff
@@ -508,11 +508,10 @@ def _search(
         # operations
         if not result:
             return 0 if count else []
-        file_ids = set(result)
         directories = self._get_directories_from_database(result)
         for directory in directories - directories._filter_access("read"):
-            file_ids -= set(directory.sudo().mapped("file_ids").ids)
-        return len(file_ids) if count else list(file_ids)
+            result -= set(directory.sudo().mapped("file_ids").ids)
+        return len(result) if count else list(result)
```
joao-p-marques: I don't think this is going to work... `result` is a list (as returned from the original Odoo function https://github.com/odoo/odoo/blob/13.0/odoo/models.py) and you are trying to subtract a `set`, which is incompatible:
```
>>> [1, 2, 3] - {1}
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for -: 'list' and 'set'
```
Maybe the subtracted item also needs to be a list, but I wonder if the conversion to `set` (that avoids duplicates) is necessary for this logic...

joao-p-marques (changes requested):

dms/models/dms_file.py
victoralmau: IMO `_search()` function is not necessary now because only returned records that user can read. 
What do you think @etobella

joao-p-marques: Well, that would be even better :sweat_smile: If it is possible, that option LGTM

yajo: > IMO `_search()` function is not necessary now because only returned records that user can read.

IIRC That will only be true after #107, so we could consider adding the patch there, or keep this one as draft until then.

yajo:
@victoralmau What happened? Did you close this on purpose?

victoralmau:
> @victoralmau What happened? Did you close this on purpose?

Sorry, these changes merged in https://github.com/OCA/dms/pull/107

PR #107, opened 2021-06-23
[13.0][REF] dms: modify permissions computation

victoralmau:
Related to https://github.com/OCA/dms/pull/102 + Adapt 13.0

- [NEW] Fix prevent error related to mail_message parent not exists in some cases (directory)

Locked: 
- [ ] https://github.com/OCA/dms/pull/106

@Tecnativa TT29847

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -4,13 +4,13 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import base64
-import functools
 import logging
-import operator
 from collections import defaultdict
 
 from odoo import _, api, fields, models, tools
-from odoo.exceptions import AccessError, UserError, ValidationError
+from odoo.exceptions import UserError, ValidationError
+from odoo.osv.expression import OR
+from odoo.tools import consteq
 
 from odoo.addons.http_routing.models.ir_http import slugify
 
@@ -39,6 +39,7 @@ class DmsDirectory(models.Model):
 
     _parent_store = True
     _parent_name = "parent_id"
+    _directory_field = _parent_name
 
     parent_path = fields.Char(index=True)
     is_root_directory = fields.Boolean(
@@ -59,16 +60,15 @@ class DmsDirectory(models.Model):
         ondelete="restrict",
         auto_join=True,
         store=True,
-        copy=True,
-        default=False,
     )
-
     parent_id = fields.Many2one(
         comodel_name="dms.directory",
-        domain="[('permission_create', '=', True)]",
         string="Parent Directory",
+        domain="[('permission_create', '=', True)]",
         ondelete="restrict",
-        auto_join=True,
+        # Access to a directory doesn't necessarily mean access its parent, so
+        # prefetching this field could lead to misleading access errors
+        prefetch=False,
         index=True,
         store=True,
         readonly=False,
@@ -84,6 +84,24 @@ class DmsDirectory(models.Model):
         else:
             return False
 
+    group_ids = fields.Many2many(
+        comodel_name="dms.access.group",
+        relation="dms_directory_groups_rel",
+        column1="aid",
+        column2="gid",
+        string="Groups",
+    )
+    complete_group_ids = fields.Many2many(
+        comodel_name="dms.access.group",
+        relation="dms_directory_complete_groups_rel",
+        column1="aid",
+        column2="gid",
+        string="Complete Groups",
+        compute="_compute_groups",
+        readonly=True,
+        store=True,
+        compute_sudo=True,
+    )
     complete_name = fields.Char(
         "Complete Name", compute="_compute_complete_name", store=True
     )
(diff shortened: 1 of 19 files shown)


victoralmau:
Please @Yajo can you review it?

dms/models/directory.py
```diff
@@ -94,6 +112,25 @@ def _default_parent_id(self):
...
+        index=True,
+    )
+
+    color = fields.Integer(string="Color", default=0)
+
+    category_id = fields.Many2one(
+        comodel_name="dms.category",
+        context="{'dms_category_show_path': True}",
+        string="Category",
+    )
```
yajo: These changes do not exist in #102. What happened?

It seems these should come from `abstract.dms.mixin` in v13, including `is_hidden` (which is in v12 but probably not needed here).

Could you check that please?

dms/models/directory.py
```diff
@@ -638,7 +647,11 @@ def create(self, vals_list):
                 parent = self.browse([vals["parent_id"]])
                 data = next(iter(parent.sudo().read(["storage_id"])), {})
                 vals["storage_id"] = self._convert_to_write(data).get("storage_id")
-        return super().create(vals_list)
+        # Hack to prevent error related to mail_message parent not exists in some cases
+        ctx = dict(self.env.context).copy()
+        ctx.update({"default_parent_id": False})
+        res = super(DmsDirectory, self.with_context(ctx)).create(vals_list)
```
yajo, line 634: This code is easy to simplify:

```suggestion
        # Hack to prevent error related to mail_message parent not exists in some cases
        res = super(DmsDirectory, self.with_context(default_parent_id=False)).create(vals_list)
```

I see that you were able to remove this HACK due to the new v13 ORM: https://github.com/OCA/dms/blob/db9cb933ba4f29cc1a610190c32e2ff4bb16b1ce/dms/models/directory.py#L654-L660

Nice!

dms/models/dms_file.py
```diff
@@ -179,6 +173,7 @@ def check_access_token(self, access_token=False):
         invisible=True,
         ondelete="cascade",
     )
+    storage_id_save_type = fields.Selection(related="storage_id.save_type", store=False)
```
yajo: This change does not exist in #102. Is it really needed?

dms/models/dms_security_mixin.py
```diff
@@ -2,336 +2,216 @@
...
-        if operator == "=" and operand:
-            return [("id", "in", self.search([])._filter_access_ids("create"))]
-        return [("id", "not in", self.search([])._filter_access_ids("create"))]
+    def _get_permission_domain(self, operator, value, operation):
+        """Abstract logic for searching computed permission fields."""
+        # Tricky one, to know if you want to search
+        # positive or negative access
+        positive = (operator not in NEGATIVE_TERM_OPERATORS) == bool(value)
+        if self.env.uid == SUPERUSER_ID:
+            return TRUE_DOMAIN if positive else FALSE_DOMAIN
```
yajo: Another HACK removed from v12: https://github.com/OCA/dms/blob/db9cb933ba4f29cc1a610190c32e2ff4bb16b1ce/dms/models/dms_security_mixin.py#L176-L184

Nice!

dms/security/security.xml
```diff
@@ -68,34 +81,109 @@
...
+    <!-- These rules leverage computed permission management -->
+    <record id="rule_directory_computed_create" model="ir.rule">
+        <field name="name">Apply computed create permissions.</field>
+        <field name="model_id" ref="model_dms_directory" />
+        <field name="global" eval="True" />
+        <field name="perm_read" eval="0" />
+        <field name="perm_create" eval="1" />
+        <field name="perm_write" eval="0" />
+        <field name="perm_unlink" eval="0" />
+        <field name="domain_force">[('permission_create', '=', True)]</field>
```
yajo: And the last HACK from v12 also removed: https://github.com/OCA/dms/blob/db9cb933ba4f29cc1a610190c32e2ff4bb16b1ce/dms/security/security.xml#L119

Nice!

yajo (changes requested):
Some comments just point out the removed HACK, and that's nice!

However there are some other things to fix.

Also I see that you squashed all commits from #102. No problem on that, but it'd be nice to include in your commit message the contents of all those 3.

Add also `Co-authored-by: Victor Martínez <your-email>` at the end, [to keep co-authorship](https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors), as you totally deserve it.

dms/models/dms_file.py
victoralmau: Sorry, it's not necessary because exists in `abstract.dms.mixin`. Thanks!

dms/models/directory.py
victoralmau: Sorry, it's not necessary because exists in `abstract.dms.mixin`. Thanks!

yajo (changes requested):
Now the code looks good. However, I found an usability issue.

In v12, when you create a directory assigning no groups (so you're creating an obvious permissions problem), you get this error, telling you you cannot create it:

image

However, now in this v13 runbot, I tested the same and the error is about problems to _read_ the directory:

screenshot-2021-07-02-11-34-09


All this means that the directory was effectively created, but then, after creating, when trying to read it, there's the access error. I can double-check that by becoming superuser and asserting that it effectively exists:

screenshot-2021-07-02-11-35-34

This is not what should happen. Instead, just like in v12, when creating the record, the creation rules should be checked and should block (or roll back) the record creation, to avoid allowing these senseless conditions.

yajo (approved):

yajo (approved):

Issue #40, opened 2020-12-08
Subdirectories bug


thopv133802:
when I try to access subdirectories !
![image](https://user-images.githubusercontent.com/44016234/101469093-dbb61700-3976-11eb-9c9d-8a84ab18726e.png)
![image](https://user-images.githubusercontent.com/44016234/101469130-e83a6f80-3976-11eb-9815-ef6870264088.png)
![image](https://user-images.githubusercontent.com/44016234/101469155-f1c3d780-3976-11eb-9956-61a08d09d5ad.png)
![image](https://user-images.githubusercontent.com/44016234/101469167-f5eff500-3976-11eb-8b8d-5ee91ee9358f.png)
![image](https://user-images.githubusercontent.com/44016234/101469225-0c964c00-3977-11eb-93b8-ec2f52730fa9.png)

pedrobaeza:
Can you explain a bit the problem? Maybe a video? You should also try on runbot, not on local, as maybe your problem is something of the deployment.

gaikaz:
I think what he is trying to show here is that the currently open directory is always shown as a sub-directory of itself, which leads to this "inception" sort of behavior (look at the breadcrumbs of each screenshot). 
I also agree that this is kind of misleading for the end-user as it is very unusual for any kind of "File explorer" programs.

victoralmau:
IMO it's solved with https://github.com/OCA/dms/pull/117 in 13.0, right?

yajo:
Yes, it should. Reopen otherwise please.

PR #109, opened 2021-07-06
[13.0][FIX] dms: Show only real subdirectories in view and when we use searchpanel directory filter

victoralmau:
Show only real subdirectories in view and when we use searchpanel directory filter.
Related to https://github.com/OCA/dms/issues/40

It's necessary to apply 14.0 too.

@Tecnativa TT30826

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -402,7 +402,7 @@ class DmsDirectory(models.Model):
 
     def _compute_count_total_directories(self):
         for record in self:
-            count = self.search_count([("id", "child_of", record.id)])
+            count = self.search_count([("parent_id", "=", record.id)])
             count = count - 1 if count > 0 else 0
             record.count_total_directories = count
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@keshrath @etobella what do you think about this change of behavior, based on the feedback given by our users, saying that it's very confusing to see all subfolders each time.

dms/models/directory.py
```diff
@@ -682,3 +682,31 @@ def unlink(self):
...
+                    parent_id_override
+                    and not bool(
+                        item[0] == "parent_id" and item[1] == "=" and not item[2]
+                    )
+                ):
+                    domain2.append(item)
+            domain = domain2
+        return super().search_read(
+            domain=domain, fields=fields, offset=offset, limit=limit, order=order
+        )
```
yajo, line 712: Why do you need to override this method? Doesn't `[('parent_id', '=', False)]` return the top level directories already? Can't you just use that domain in the action?

pedrobaeza: But shouldn't folders be shown according current subdirectory level? Thus it should be something like: `parent_id = <current_subdirectory>`, being `False` or other.

victoralmau:
I close this because this idea is included in https://github.com/OCA/dms/pull/117

PR #112, opened 2021-07-09
[12.0][FIX] dms: me cases admin and demo users have previously in 'Show private access' group and causes some tests to fail.

victoralmau:
In some cases admin and demo users have previously in 'Show private access' group and causes some tests to fail. (Standarize according to 13.0) (It's fixed directly in https://github.com/OCA/dms/pull/107)

@Tecnativa

### DIFF
--- a/dms/tests/test_storage_attachment.py
+++ b/dms/tests/test_storage_attachment.py
@@ -17,6 +17,11 @@ class StorageAttachmentTestCase(DocumentsBaseCase):
                 "groups_id": [(6, 0, [self.env.ref("base.group_user").id])],
             }
         )
+        user_admin = self.browse_ref("base.user_admin")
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I think you should use other user for tests instead of that 2.

victoralmau:
> I think you should use other user for tests instead of that 2.

Yes, totally agree but that would make it necessary to change all tests now. In the future it would be a good idea to redo the tests with users created in the tests (not admin and demo).

pedrobaeza:
Then instead of this, just remove both users from the group on the setup of the test instead of doing this ugly `if`, as you are leaving without testing certain things.

joao-p-marques (approved):

pedrobaeza (approved):

yajo (approved):

PR #113, opened 2021-07-09
[14.0][REF] dms: modify permissions computation

victoralmau:
Related to https://github.com/OCA/dms/pull/107 + Adapt 14.0

- [NEW] Fix prevent error related to mail_message parent not exists in some cases (directory)
- Fix tests for environments that add other groups to base users https://github.com/OCA/dms/pull/116

Please @Yajo can you review it?

@Tecnativa TT29847

### DIFF
--- a/dms/tests/test_portal.py
+++ b/dms/tests/test_portal.py
@@ -6,6 +6,7 @@ import base64
 import odoo.tests
 
 
+@odoo.tests.tagged("post_install", "-at_install")
 class TestDmsPortal(odoo.tests.HttpCase):
     def setUp(self):
         super().setUp()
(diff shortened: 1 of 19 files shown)


dms/tests/test_portal.py
```diff
@@ -6,6 +6,7 @@
 import odoo.tests
 
 
+@odoo.tests.tagged("post_install", "-at_install")
```
yajo, line 9: Please don't do this. It's a big source for problems.

yajo (changes requested):

dms/tests/test_portal.py
victoralmau: Sorry, odoo suggested these change when is http test or warning show always.

yajo: Weird... do you have more details about what warning it logs?

victoralmau: Log message:
`odoo.addons.dms.tests.test_portal.TestDmsPortal: HttpCase test should be in post_install only`

https://github.com/odoo/odoo/blob/14.0/odoo/tests/common.py#L1377

yajo: After internal discussion with @pedrobaeza, we're gonna obey that warning.

The problems introduced with post install mode were unavoidable in lower versions, but the framework has evolved and it seems like today it should be more reliable.

We'll revert this in the future if it gives those problems again.

yajo (approved):

pedrobaeza (approved):

PR #115, opened 2021-07-09
[12.0][FIX] dms: Fix tests for environments that add other groups to base users

joao-p-marques:
Instead of relying on base users, create the ones we need for the test. This helps to avoid problems when other modules modify the groups for base users.

Missing changes from https://github.com/OCA/dms/pull/112

ping @victoralmau @Yajo


pedrobaeza (approved):

victoralmau (approved):

yajo (approved):

pedrobaeza:
Please move it to upper versions.

PR #117, opened 2021-07-13
[13.0][IMP+FIX] dms: Changes in file tree view + Fix file preview.

victoralmau:
[IMP] Changes in file tree view (It's necessary to change in 14.0)
[FIX] File preview.
[IMP ]Show in directory kanban view only root directories (Change **All** text to **Root**) + Change to searchpanel _directory_id_ (in files) or _parent_id_ (in directories) to filter equal and not _child_of_ (It's related to https://github.com/OCA/dms/issues/40 )

Locked by:
- [x] https://github.com/OCA/social/pull/744

@Tecnativa TT30992

### DIFF
--- a/dms/static/src/js/views/search_panel.js
+++ b/dms/static/src/js/views/search_panel.js
@@ -1,0 +1,45 @@
+/* Copyright 2021 Tecnativa - Víctor Martínez
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+odoo.define("dms.SearchPanel", function(require) {
+    "use strict";
+
+    const core = require("web.core");
+    const SearchPanel = require("web.SearchPanel");
+    const _t = core._t;
+    SearchPanel.include({
+        _renderCategory: function() {
+            let res = this._super.apply(this, arguments);
+            if (this.model === "dms.directory") {
+                // Replace label on JS because QWeb has not enough context
+                const res_dom = $(res);
+                res_dom.find(".o_search_panel_label_title b").html(_t("Root"));
+                res = res_dom.html();
+            }
+            return res;
+        },
+        _getCategoryDomain: function() {
+            var domain = this._super.apply(this, arguments);
+            var field_name_need_check = false;
+            if (this.model === "dms.file") {
+                field_name_need_check = "directory_id";
+            } else if (this.model === "dms.directory") {
+                field_name_need_check = "parent_id";
+            }
+            if (field_name_need_check) {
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):

dms/static/src/js/views/search_panel.js
```diff
@@ -0,0 +1,45 @@
...
+    SearchPanel.include({
+        _renderCategory: function() {
+            var res = this._super.apply(this, arguments);
+            if (this.model === "dms.directory") {
+                var str_search = '<span class="o_search_panel_label_title"><b>';
+                var pos_start = res.search(str_search) + str_search.length;
+                var pos_end = pos_start + res.substring(pos_start).search("</b>");
+                res = res.substring(0, pos_start) + _t("Root") + res.substring(pos_end);
+            }
+            return res;
```
yajo:
```suggestion
            let res = this._super.apply(this, arguments);
            if (this.model === "dms.directory") {
                // Replace label on JS because QWeb has not enough context
                let res_dom = $(res);
                res_dom.find('.o_search_panel_label_title').html(_t('<b>Root</b>'));
                res = res_dom.html();
            }
            return res;
```

dms/static/src/js/views/search_panel.js
```diff
@@ -0,0 +1,45 @@
...
+            if (this.model === "dms.file") {
+                field_name_need_check = "directory_id";
+            } else if (this.model === "dms.directory") {
+                field_name_need_check = "parent_id";
+            }
+            if (field_name_need_check) {
+                domain.forEach(function(item, key) {
+                    if (item[0] === field_name_need_check && item[1] === "child_of") {
+                        domain[key] = [item[0], "=", item[2]];
+                    }
```
yajo, line 32: This will fail if a domain leaf is `|`, `&` or `!`.

yajo (changes requested):
Just some little comments to make it more resilient.

dms/static/src/js/views/search_panel.js
victoralmau: It is not necessary to change anything, right?
We only want to modify domain when it is `child_of`.

yajo: Ah sorry I was on my python mode, 🤦🏼‍♂️  and though that `"|"[1]` would produce an `IndexError`, while in ES it returns `undefined` and just skips the if.

yajo (approved):

PR #121, opened 2021-07-20
[IMP] dms: automatic image thumbnails

yajo:
If a dms.file contains an image, use it as thumbnail automatically, and hide controls for customizing that record's thumbnail.

**Note for reviewers: please merge as nobump**

@Tecnativa TT31004


frankofno:
i guess the "Could not display the selected image" message/notification will be resolved with this?

victoralmau (approved):

dms/models/dms_file.py
```diff
@@ -133,6 +133,27 @@ class File(models.Model):
...
+    @api.depends(
+        "custom_thumbnail_medium", "custom_thumbnail_small", "res_mimetype", "content"
+    )
+    def _compute_thumbnail(self):
+        """Provide image thumbnail automatically if possible."""
+        result = super()._compute_thumbnail()
+        for one in self.filtered("res_mimetype"):
+            if one.res_mimetype.startswith("image/"):
+                for field in ("thumbnail", "thumbnail_medium", "thumbnail_small"):
+                    one[field] = one.content
```
pedrobaeza: This strategy is not adequate, as it will provide the full image to all thumbnail sizes. You should use image resize methods for providing the proper one. I'm also worried about the performance of this not being stored. What about putting the image as custom thumbnail on content changed and let the rest to work as is?

dms/models/dms_file.py
```diff
@@ -133,6 +133,27 @@ class File(models.Model):
         attachment=True, string="Content File", prefetch=False, invisible=True
     )
 
+    @api.depends("res_mimetype")
+    def _compute_custom_thumbnail_editable(self):
```
pedrobaeza: This seems a bit convoluted/duplicated. We have the custom thumbnail field for adding any kind of preview according mimetypes. Why adding other mechanism?

dms/models/dms_file.py
yajo: This was implemented in #88 on purpose, for performance. When loading multiple records, it turns out to be more heavy having to resize twice each image than just setting all thumbnails the same size and letting the browser adjust them. Keep in mind that this is computed and non-stored, so it'd spend lots of CPU cycles resizing once and again the same images.

In reality, most chances are today we could just remove all those 3 fields, and rely on using a custom thumbnail if available, or the image URL otherwise. But these were kept to ease the instability caused by so many patches on standard versions.

One alternative would be to split this into 3 methods, so the ORM only lazy-computes the ones really needed, but that'd turn out to have them never used for the reason just explained, so this seemed like the simplest solution.

I can add the removal of those fields if you think it's worth it at this point.

dms/models/dms_file.py
yajo: In line with https://github.com/OCA/dms/pull/121#discussion_r673703104, the purpose is to avoid using those computed thumbnails at all.

The fact of using a icon URL instead of an icon thumbnail improves performance a lot everywhere. For instance, your browser will only download one "folder.svg" icon when you're browsing your folders, instead of one icon per record. Besides, it'll cache it, so actually using different sizes is more slow & expensive (more resources to cache) than always using the big image.

OTOH I considered using the custom thumbnail fields as computed writable, `custom_thumbnail` is stored with `attachment=False`, so I couldn't just rely on Odoo's built-in filestore deduplication system if I applied the same image to the `custom_thumbnail` field. So, if I stored in the `custom_thumbnail` the same image as the file content, I'd be occupying 2x disk space: once on db and once on filestore. Well, actually more than 2x because of the 2 smaller thumbnail versions, but that's how thumbnails work so it's no the biggest problem.

Maybe a better fix would be to add `attachment=True` and write the corresponding migration script.

I preferred to keep it as it was because I didn't understand why it was `attachment=False` (so I assumed it was for a reason). Also, those custom smaller thumbnails would produce smaller versions, but then there's the caching and optimizations explained above...

So... well, things aren't so simple. I think we could rethink this on migration for v15 and remove garbage stuff in this module.

I'd *really* love to just get rid of all the custom thumbnails stuff. People can put tags and categories to files, so as not to need having those custom thumbnails. Besides, I doubt anybody in the world will ever takes screenshots of their PDFs to put them as thumbnails here. The use case is too hypothetical IMHO. Maybe we can remove them in v15 and, if there's a real need, let somebody else rescue code from v14 in a separate module?

dms/models/dms_file.py
pedrobaeza: One of the purposes of extra fields is to avoid sending a lot of big files through Internet for only previewing something. This is what `image.mixin` does. And for me, only 2 main fields should be exists: icon and custom_image/custom_preview, but with the automatic extension to lower resolutions provided by `image.mixin`. Can you do this switch?

yajo: Let me check.

dms/models/dms_file.py
pedrobaeza: The requirement of custom preview is required for the purposes we have already talked about (other extensions providing such preview). The way is implemented is ours. There's no reason for `attachment=False` IMO, just to not remember such feature. Just think on how to apply KISS and DRY principles here.

yajo: OK, do you have any example around about how to migrate from `attachment=False` to `attachment=True`? That'd help a lot.

pedrobaeza: https://github.com/OCA/bank-payment/blob/10.0/account_banking_mandate/migrations/10.0.2.0.0/post-migration.py

victoralmau (approved):
Code review OK.

pedrobaeza:
Travis is red

yajo:
It should go green now.

pedrobaeza (approved):
Thanks for the excellent work

pedrobaeza:
/ocabot merge nobump

Now to v14

yajo:
> Now to v14

Done in #126

PR #122, opened 2021-07-21
[14.0][IMP+FIX] dms: Changes in file tree view + Fix file preview.

victoralmau:
FWP from 13.0: https://github.com/OCA/dms/pull/117

Changes done:
- [x] [IMP] Changes in file tree view
- [x] [FIX] File preview.
- [x] [IMP]Show in directory kanban view only root directories (Change **All** text to **Root**) + Change to searchpanel _directory_id_ (in files) or _parent_id_ (in directories) to filter equal and not _child_of_ (It's related to https://github.com/OCA/dms/issues/40 )
- [x] [FIX] Display the short name of directories in searchpanel.

@Tecnativa TT30992

### DIFF
--- a/dms/views/dms_file.xml
+++ b/dms/views/dms_file.xml
@@ -228,10 +228,21 @@
                                             class="o_kanban_dms_file_preview"
                                             t-att-data-id="widget.db_id"
                                         >
-                                            <img
-                                                t-att-src="record.icon_url.raw_value"
-                                                alt="Icon"
-                                            />
+                                            <t
+                                                t-if="record.custom_thumbnail_small.raw_value"
+                                            >
+                                                <img
+                                                    t-att-src="kanban_image('dms.file', 'custom_thumbnail_small', record.id.raw_value)"
+                                                    t-att-class="o_kanban_dms_file_preview"
+                                                    alt="Custom thumbnail"
+                                                />
+                                            </t>
+                                            <t t-else="">
+                                                <img
+                                                    t-att-src="record.icon_url.raw_value"
+                                                    alt="Icon"
+                                                />
+                                            </t>
                                         </a>
                                     </div>
                                 </div>
@@ -240,11 +251,6 @@
                                         <div
                                             class="o_kanban_record_title o_text_overflow"
                                         >
-                                            <a
-                                                class="o_kanban_dms_file_preview fa fa-search"
-                                                title="Search"
-                                                t-att-data-id="widget.db_id"
-                                            />
                                             <field name="name" />
                                         </div>
                                         <div class="o_kanban_record_body">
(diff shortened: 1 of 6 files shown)


yajo (dismissed):

dms/views/dms_file.xml
```diff
@@ -311,6 +313,11 @@
                 <field name="size" />
                 <field name="mimetype" />
                 <field name="path_names" widget="path_names" string="Path" />
```
pedrobaeza: Shouldn't this be removed?

yajo (changes requested):
I'd say this patch simply doesn't work. I tested it on runbot and I see no effect:

image

image

victoralmau:
Can you check @Tardo to try to fix it? Thanks!

pedrobaeza:
What is blocking this?

pedrobaeza:
@victoralmau what is missing?

victoralmau:
Can you check @Tardo to try to fix it? Thanks!

andreampiovesana (approved):
LGTM

pedrobaeza:
@victoralmau can you resume this?

victoralmau:
All these changes are already merged in v13.

There is a pending issue regarding https://github.com/OCA/dms/pull/122#pullrequestreview-712435850 (I will try to resolve it).

pedrobaeza:
Please ping when done.

victoralmau:
Changes done. Could you review it in runboat?

pedrobaeza (approved):

pedrobaeza (changes requested):
Sorry, there's still a problem on directories part: on the runboat, on the directories part, if you unfold the Media" one, only one of the four subfolders is shown. Third level folders are not shown, and the path is not stripped (it's shown as "Media / Photos" instead of just "Photos")

victoralmau:
You are right, the name of the folders shown in searchpanel was not correct. Although it is a change that currently exists in v14 I have added a commit to resolve it.

The list of folders shown in the searchpanel is also not correct (I have pending to review it to resolve it) although it is something that happens now in v14.

pedrobaeza:
Should I wait then for that fix or do we have it for other PR?

victoralmau:
IMO it is better to merge this already and then, create another PR to solve the other issue.

pedrobaeza (approved):

pedrobaeza:
Ok, merging now.

victoralmau:
Checking regarding the error "Show all folders in the search panel" it is actually not an error and is working correctly. The purpose of searchpanel is to show records if they have child records. Only folders that have subfolders are shown (this behavior in files does not happen because if a file exists it is because it is in a directory and for that reason, that directory is shown in searchpanel).

pedrobaeza:
But that seems very weird for user perspective... What we want on the search folder is a complete folder structure, not only those with childs. If there are not subfolders, clicking on them we will see no elements on the right part, but that can be best accepted.

PR #127, opened 2021-07-28
[13.0][IMP] dms: Forgotten migration script related to https://github.com/OCA/dms/commit/0ae10d1714d806bb806c16796204e610b6c1a777 is added

victoralmau:
Forgotten migration script related to https://github.com/OCA/dms/commit/0ae10d1714d806bb806c16796204e610b6c1a777 is added.

It's need to apply yo 14.0 too.

Please @pedrobaeza and @Yajo  can you review it?

@Tecnativa TT30992

### DIFF
--- a/dms/__manifest__.py
+++ b/dms/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Document Management System",
     "summary": """Document Management System for Odoo""",
-    "version": "13.0.8.0.0",
+    "version": "13.0.8.1.0",
     "category": "Document Management",
     "license": "LGPL-3",
     "website": "http://github.com/OCA/dms",
--- a/dms/migrations/13.0.8.1.0/post-migration.py
+++ b/dms/migrations/13.0.8.1.0/post-migration.py
@@ -1,0 +1,9 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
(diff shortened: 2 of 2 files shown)


yajo (approved):

dms/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Document Management System",
     "summary": """Document Management System for Odoo""",
-    "version": "13.0.8.0.0",
+    "version": "13.0.8.1.0",
```
dreispt, line 7: No need to bump versions - the Bot can do that when merging

dreispt (approved):

pedrobaeza:
Should we do it for all the existing ones? Why is it needed? Isn't this computed when the field was introduced?

victoralmau:
IMO yes, is necessary to do for all records (old records have incorrect value).
`res_mimetype` field is computed in creation and is important to preview files.

pedrobaeza:
But why is it not recomputed when the contents change? Maybe a missing depends? Anyways, I don't see why this migration script is needed now.

victoralmau:
Records that were created before this change have the wrong `res_mimetype` value. The field have `content` depends but it is a stored field so it will still have the wrong value.

pedrobaeza:
OK, let's go with it although I don't fully understand it:

/ocabot merge nobump

I suppose we need it as well for v14

victoralmau:
Finally, it's not necessary in 14.0.

dms/__manifest__.py
yajo: You need it if you want to test a migration script.

PR #131, opened 2021-08-20
[13.0][FIX] dms: Set res_mimetype correctly in files

victoralmau:
Set `res_mimetype` correctly in files.

@Tecnativa TT31461

### DIFF
--- a/dms/readme/INSTALL.rst
+++ b/dms/readme/INSTALL.rst
@@ -4,3 +4,6 @@ Preview
 `mail_preview_base` is required for DMS but it is recommended to install all
 the other `mail_preview` modules from `social` OCA repository
 in order to improve the preview of files.
+
+`python-magic` library is recommended to be installed for having whole support
+to get proper file types and file preview.
--- a/dms/tests/test_file.py
+++ b/dms/tests/test_file.py
@@ -1,11 +1,17 @@
 # Copyright 2017-2019 MuK IT GmbH.
 # Copyright 2020 Creu Blanca
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import base64
 
 from .common import multi_users
 from .test_file_database import FileTestCase
+
+try:
+    import magic
+except ImportError:
+    magic = None
 
 
 class FileFilestoreTestCase(FileTestCase):
@@ -84,3 +90,9 @@ class FileFilestoreTestCase(FileTestCase):
         self.assertEqual(file_svg.res_mimetype, "image/svg+xml")
         file_logo = self.env.ref("dms.file_02_demo")
         self.assertEqual(file_logo.res_mimetype, "image/jpeg")
+
+    def test_content_file_res_mimetype_magic_library(self):
+        if not magic:
+            self.skipTest("Without python-magic library installed")
+        file_video = self.env.ref("dms.file_10_demo")
+        self.assertEqual(file_video.res_mimetype, "video/mp4")
--- a/dms/README.rst
+++ b/dms/README.rst
@@ -47,6 +47,9 @@ Preview
 the other `mail_preview` modules from `social` OCA repository
 in order to improve the preview of files.
 
+`python-magic` library is recommended to be installed for having whole support
+to get proper file types and file preview.
+
 Configuration
 =============
 
--- a/dms/static/description/index.html
+++ b/dms/static/description/index.html
@@ -3,7 +3,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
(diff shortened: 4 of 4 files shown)


pedrobaeza (changes requested):
I'm not sure we really need this, as this is very partial, but deploy properly a library that detects all the possible mimetypes instead of the limited implementation inside Odoo. Please check this code:

https://github.com/odoo/odoo/blob/c5a22a122eccd18247280a941efd2c96935c43d6/odoo/tools/mimetypes.py#L165-L179

Try in your local to put such library to see if it's detected properly, and if so we should:

- Add in base Doodba such library
- Put in the README in the installation section the recommendation to install the library.

dms/models/dms_file.py
```diff
@@ -377,10 +378,15 @@ def _compute_extension(self):
         for record in self:
             record.extension = file.guess_extension(record.name)
 
-    @api.depends("content")
+    @api.depends("content", "name")
     def _compute_mimetype(self):
+        """ Similar to addons/base/models/ir_attachment.py#L232 """
```
yajo: This comment says nothing. And the line number can change. Please add a proper docstring and, if you need to refer to another resource, add a permalink.

dms/models/dms_file.py
```diff
@@ -377,10 +378,15 @@ def _compute_extension(self):
...
         for record in self:
-            record.res_mimetype = guess_mimetype(base64.b64decode(record.content or ""))
+            record.res_mimetype = mimetypes.guess_type(record.name or "")[0]
+            if record.res_mimetype == "application/octet-stream":
+                record.res_mimetype = guess_mimetype(
+                    base64.b64decode(record.content or "")
+                )
```
yajo: IIUC, you're trying to get the mimetype from the file name and, later, from the file contents.

IMHO it should be the opposite. The contents are usually more reliable than the file extension to know the kind of file you have.

yajo (changes requested):
As @pedrobaeza said, the first thing to try is to install those libraries and see if the problem goes away automatically. If that happens, close this.

Otherwise, a couple of code comments below.

victoralmau:
I tested locally to add the `python-magic` library and this changes will not be necessary because the `res_mimetype` field will be defined correctly.

Now the question is, do we add it as a dependency in the manifest or do we suggest its installation in the readme file and skip the parts of the test that are necessary if the library is not added?

Let's not forget that this happened because the `res_mimetype` field for `.mp4` files was not defined correctly.

pedrobaeza:
See my previous comment in https://github.com/OCA/dms/pull/131#pullrequestreview-735078892

victoralmau:
Changes done.
If everything is correct, do we add the `python-magic` library to https://github.com/Tecnativa/doodba or is it optional?

pedrobaeza:
> If everything is correct, do we add the `python-magic` library to https://github.com/Tecnativa/doodba or is it optional?

IMO we should add it for all (as it affects also to non DMS users that attach a video in chatter), but meanwhile we can add it to current instances through the regular way. cc @Yajo

dms/readme/INSTALL.rst
```diff
@@ -4,3 +4,6 @@ Preview
 `mail_preview_base` is required for DMS but it is recommended to install all
 the other `mail_preview` modules from `social` OCA repository
 in order to improve the preview of files.
+
+`python-magic` library is recommended to install to correctly define file
+types and file preview.
```
pedrobaeza:
```suggestion
`python-magic` library is recommended to be installed for having whole support
to get proper file types and file preview.
```

pedrobaeza (approved):

dms/tests/test_file.py
```diff
@@ -84,3 +90,9 @@ def test_content_file_res_mimetype(self):
         self.assertEqual(file_svg.res_mimetype, "image/svg+xml")
         file_logo = self.env.ref("dms.file_02_demo")
         self.assertEqual(file_logo.res_mimetype, "image/jpeg")
+
+    def test_content_file_res_mimetype_magic_library(self):
+        if not magic:
+            self.skipTest("Without python-magic library installed")
```
yajo, line 96: FTR there's a decorator for this: https://docs.python.org/3/library/unittest.html#unittest.skipIf

However it's OK as it is 😊

yajo (approved):

yajo:
Failed due to https://github.com/odoo/odoo/pull/75387.

rvalyi:
Hello @Yajo, you could use https://github.com/OCA/l10n-brazil/pull/1587/commits/d43e0eb3329a8116628259a1b6f5ec556e59efb5 
(see https://odoo-community.org/groups/contributors-15/contributors-206982?mode=date&date_begin=2021-08-01&date_end=2021-09-01 )
to workarround the issue meanwhile if you want. That being said it seems Odoo SA is moving forward with this issue so may be waiting they fix it is an option.

pedrobaeza:
/ocabot merge patch

It can be fw-port it to v14 as well

PR #135, opened 2021-09-13
Explicit access group information

majenkotech:
Add explicit access group creation and usage instructions in the brief getting started instructions.

### DIFF
--- a/dms/README.rst
+++ b/dms/README.rst
@@ -56,13 +56,14 @@ To configure this module, you need to:
 #. Create a new document storage. You can choose between two options on `Save Type`:
     * `Database`: Store the files on the database as a field
     * `Attachment`: Store the files as attachments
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
Thanks for the changes. You have to perform them on readme/CONFIGURE.rst file instead, and they will be dumped to this file.

Please don't remove the access group part, as it's still relevant.

pedrobaeza:
Thanks. Only one missing thing: can you please squash both commits into one?

majenkotech:
Yeah, I'm just going to nuke the fork and start afresh GH sucks sometimes.

pedrobaeza:
Yes, I never worked directly with GH interface, but use local git, but I admit it complicates a bit the things.

PR #146, opened 2021-11-05
[13.0][FIX] dms: Define the file extension correctly (even if there is no extension in the file name).

### LINKED ISSUE #125: extension is needed in filename
the field name contains also the extension. If you remove the .pdf for example, the extension is no longer set, the preview thumb is gone and a notification is showing up.

there should be seperate field, where you can set or edit the real filename with extension. the name of the record should be whatever the user wants to type in without extension.

https://user-images.githubusercontent.com/35775329/126568574-670923e1-34d2-40ef-a874-3b2b4a38d102.mov

victoralmau:
Define the file extension correctly (even if there is no extension in the file name). 
Related to https://github.com/OCA/dms/issues/125.

Please @joao-p-marques and @Yajo can you review it?

@Tecnativa

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -373,10 +373,12 @@ class File(models.Model):
                 }
             )
 
-    @api.depends("name")
+    @api.depends("name", "res_mimetype", "content")
     def _compute_extension(self):
         for record in self:
-            record.extension = file.guess_extension(record.name)
+            record.extension = file.guess_extension(
+                record.name, record.res_mimetype, record.content
+            )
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Don't mix 2 unrelated patches in the same PR. This does it more difficult to be reviewed/merged. For example, I have cleared one patch, but not the other about the total number of files. It's very opinionated to prefer the total or the partial one.

pedrobaeza (approved):

yajo (approved):

PR #156, opened 2022-01-24
[13.0][FIX] dms: Change size field type from integer to float to prevent 2147483647 limit from postgresql.

victoralmau:
Change `size` field type from integer to float to prevent 2147483647 limit from postgresql.

It's need to apply too to 14.0

@Tecnativa TT33935

### DIFF
--- a/dms/views/directory.xml
+++ b/dms/views/directory.xml
@@ -451,7 +451,7 @@
                     </group>
                     <group name="data">
                         <group>
-                            <field name="size" />
+                            <field name="size" widget="integer" />
                             <field name="count_elements" string="Elements" />
                         </group>
                         <group>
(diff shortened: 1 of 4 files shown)


dms/views/directory.xml
```diff
@@ -451,7 +451,7 @@
                     </group>
                     <group name="data">
                         <group>
-                            <field name="size" />
+                            <field name="size" widget="integer" />
```
pedrobaeza, line 454: Isn't here a widget `dms_size`? Is interesting to be rescued? cc @etobella

pedrobaeza: Any answer?

victoralmau: Finally, do we merge this with `widget="integer"` or try to rescue `widget="dms_size"`?

pedrobaeza: Let's go for now with the integer, although it would be interesting to rescue it.

pedrobaeza (approved):

pedrobaeza:
Fast-tracking for solving the problem:

/ocabot merge patch

pedrobaeza:
Cherry-picked for v14 in acfe367534bd768d44acffec52159539dc5ed901

Issue #137, opened 2021-09-16
[14.0] Create Directory | TypeError: 'NewId' object is not iterable


EricNLecuyer:
On a fresh install, I am able to create Storage and Files, but whenever I try to create a Directory (either from the Storage page or the "Create" button on the Directory page), I get the following error:

`Error:
Odoo Server Error

Traceback (most recent call last):
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 237, in _dispatch
    result = request.dispatch()
  File "/opt/odoo/odoo/http.py", line 682, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/odoo/http.py", line 358, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/odoo/http.py", line 346, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/odoo/http.py", line 911, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/odoo/http.py", line 530, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/addons/web/controllers/main.py", line 1359, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/main.py", line 1351, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 396, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 383, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/odoo/models.py", line 6128, in onchange
    value = record[name]
  File "/opt/odoo/odoo/models.py", line 5617, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/opt/odoo/odoo/fields.py", line 973, in __get__
    self.compute_value(recs)
  File "/opt/odoo/odoo/fields.py", line 1112, in compute_value
    records._compute_field_value(self)
  File "/opt/odoo/addons/mail/models/mail_thread.py", line 410, in _compute_field_value
    return super()._compute_field_value(field)
  File "/opt/odoo/odoo/models.py", line 4020, in _compute_field_value
    getattr(self, field.compute)()
[...]

PR #157, opened 2022-01-24
[13.0][FIX] dms: Prevent CacheMis error from count_total_files in some NewId use cases.

victoralmau:
Prevent CacheMis error from `count_total_files` in some NewId use cases.

Related to: https://github.com/OCA/dms/issues/137

It's need to apply too to 14.0

@Tecnativa TT34095

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -410,15 +410,19 @@ class DmsDirectory(models.Model):
 
     def _compute_count_total_directories(self):
         for record in self:
-            count = self.search_count([("id", "child_of", record.id)])
-            count = count - 1 if count > 0 else 0
-            record.count_total_directories = count
+            count = (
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
The same may happen on the directory count, isn't it? This can be avoided if you do `len(o2m_field)` if existing, although not sure which one is more performant.

chienandalu (approved):

victoralmau:
> The same may happen on the directory count, isn't it? This can be avoided if you do `len(o2m_field)` if existing, although not sure which one is more performant.



> 

The error is related to https://github.com/OCA/dms/blob/13.0/dms/views/directory.xml#L501-L509, there was really no problem with `count_total_directories` field but I changed it too to avoid inconsistencies if decide to add.

pedrobaeza (approved):

PR #167, opened 2022-03-04
[13.0][FIX] dms: Prevent 200 records limit in search panel.

victoralmau:
[FIX] Prevent 200 records limit in search panel.
[FIX] Show a short name in directory kanban view (similar to file view) in search panel.
[IMP] Add title to elements in search panel if text is some large.

@Tecnativa TT31492

### DIFF
--- a/dms/models/abstract_dms_mixin.py
+++ b/dms/models/abstract_dms_mixin.py
@@ -1,9 +1,12 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, models
+import odoo
+from odoo import api, fields
+
+from odoo.addons.web.models import models
 
 
-class AbstractDmsMixin(models.AbstractModel):
+class AbstractDmsMixin(odoo.models.AbstractModel):
     _name = "abstract.dms.mixin"
     _description = "Abstract Dms Mixin"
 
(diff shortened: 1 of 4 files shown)


dms/models/base.py
```diff
@@ -17,3 +18,17 @@ def unlink(self):
             [("res_model", "=", self._name), ("res_id", "in", self.ids)]
         ).unlink()
         return result
+
+    @api.model
+    def search_panel_select_range(self, field_name):
```
pedrobaeza: Why not overriding this in the specific models?

pedrobaeza: And I prefer that you overwrite by monkey-patching the variable `SEARCH_PANEL_LIMIT` before calling super and restore after than verbatim copying code.

pedrobaeza: Please read my second comment to see if feasible.

pedrobaeza:
Why not putting the override in `abstract.dms.mixin`?

dms/models/base.py
victoralmau: Yes, it is feasible and much more optimal and readable.

pedrobaeza (approved):

victoralmau:
@Tardo , can you review it?

etobella (approved):

PR #175, opened 2022-05-12
column "custom_thumbnail" does not exist

gtorresemmanuel:
bad query: ALTER TABLE "dms_file" RENAME "custom_thumbnail" TO "openupgrade_legacy_14_0_custom_thumbnail"
ERROR: column "custom_thumbnail" does not exist

### DIFF
--- a/dms/migrations/14.0.4.0.0/post-migration.py
+++ b/dms/migrations/14.0.4.0.0/post-migration.py
@@ -5,12 +5,12 @@ from openupgradelib import openupgrade, openupgrade_90
 
 def convert_binary_fields_to_attachment(env):
     """Convert old user-provided thumbnails to attachments."""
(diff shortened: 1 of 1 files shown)


dms/migrations/14.0.4.0.0/post-migration.py
```diff
@@ -5,12 +5,12 @@
 
 def convert_binary_fields_to_attachment(env):
     """Convert old user-provided thumbnails to attachments."""
-    column = openupgrade.get_legacy_name("custom_thumbnail")
```
etobella, line 8: wouldn't be better to do `if openupgrade.column_exists(env.cr, "dms_file", column)` and not removing some migration commits?

etobella (changes requested):

victoralmau:
Fixed in https://github.com/OCA/dms/pull/199

PR #176, opened 2022-05-12
column "custom_thumbnail" does not exist

gtorresemmanuel:
column "custom_thumbnail" does not exist


pedrobaeza:
This is not the correct fix, but check the column existence before executing that.

victoralmau:
Fixed in https://github.com/OCA/dms/pull/199

PR #179, opened 2022-05-17
[ADD] dms_action

etobella:
New module to handle actions on documents.
To do:

- [x] Tests
- [x] Readme
- [x] Demo Data


marcelsavegnago:
ping @augustodinizl

victoralmau (commented):
Thanks, very good suggestion but I don't understand why it is necessary to add .js code if a button in the file (view kanban for example) open the wizard.

etobella:
The JS is used in order to avoid two clicks. When you open the wizard you need to select the option and then select go without the JS. Right now, you just select the action and it is executed :smile:

Eduardolimr:
DId some functional test with server actions and files, everything seems OK to me.

eilimrib (commented):
Did some functional tests, everything seems OK

eilimrib (approved):

dms_action/models/dms_file.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import models
+
+
+class DmsFile(models.Model):
+    _inherit = "dms.file"
+
+    def action_execute_action(self):
+        self.ensure_one()
+        action = self.env.ref("dms_action.dms_action_file_act_window").read()[0]
```
victoralmau:
```suggestion
        action = self.env.ref("dms_action.dms_action_file_act_window").sudo().read()[0]
```

dms_action/views/dms_file.xml
```diff
@@ -0,0 +1,45 @@
...
+            <xpath expr="//div[hasclass('o_kanban_image')]" position="before">
+
+                <div class="mk_file_kanban_actions">
+                    <div class="mk_file_kanban_actions_wrapper">
+                        <a
+                            type="object"
+                            name="action_execute_action"
+                            role="button"
+                            class="btn btn-sm btn-outline-primary mk_directory_kanban_directories"
+                            title="Play action"
```
victoralmau: Some possibilities about it:
- A: Hide the button when there are no actions available (it is strange to press and that the wizard is empty).
- B: Place the button just before the block button (in this way that block would not be added to the left of the file image).
- C: Define an icon for each action, and place them directly in the kanban view before the lock button (this would avoid the wizard).

IMO option C is the most optimal.

victoralmau (changes requested):
Good job, Code review OK, some comments

elvise:
@etobella can you check the suggestion from @victoralmau ?

dms_action/views/dms_file.xml
etobella: Did the change on the kanban (really hard, by the way :rofl: ). If you like it, I will do something similar on the form with the header and remove the wizard. WDYT? @victoralmau

etobella:
@victoralmau This is ready for review if you want :smile:

pedrobaeza:
He's on holidays now, so not sure if it's going to be very responsive.

etobella:
No need to hurry, I just wanted to clarify that it is ready for review :laughing:

dms_action/static/src/js/dms_action_button_widget.js
```diff
@@ -0,0 +1,23 @@
...
+    License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+ **********************************************************************************/
+
+odoo.define("dms_action.button_widget", function (require) {
+    "use strict";
+
+    var AbstractField = require("web.AbstractField");
+    var field_utils = require("web.field_utils");
+    var registry = require("web.field_registry");
+    console.log("HI");
```
victoralmau:
```suggestion
```

dms_action/views/dms_file.xml
victoralmau: I think it's better this way in the kanban view, if you can change it in the form view to remove wizard it would be perfect.

victoralmau (changes requested):

victoralmau:
Fix tests and it would be done.

victoralmau (approved):

PR #181, opened 2022-05-18
Revert "[14.0][FIX] dms: Prevent 200 records limit in search panel."

etobella:
Reverts OCA/dms#180

It is failing, as you can see on the travis action. I think it is probably no longer needed


etobella:
Fast tracking, as dms is failing

/ocabot merge patch

pedrobaeza:
@victoralmau please check if it's still needed and the patch should be other way

victoralmau:
Sorry, after reviewing this in v14, it is no longer necessary to apply this change.

victoralmau:
Sorry, after reviewing this in v14, it is no longer necessary to apply this change.

PR #182, opened 2022-05-18
[ADD] dms_storage

etobella:
Allows to storage DMS Files on external storages. 
TO DO:
- [ ] Update the external directory when changing the drectory from DMS


marcelsavegnago:
ping @augustodinizl

dms_storage/models/dms_storage.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class DmsStorage(models.Model):
+    _inherit = "dms.storage"
+
+    save_type = fields.Selection(
+        selection_add=[("storage", "Storage")], ondelete={"storage": "cascade"}
```
victoralmau, line 11: This is really necessary? With this approach, it could not be used as an attachment type storage actually.

etobella: Well, the idea is to store files on a different place. If you use attachment you will store it on you attachment system. With storage you use the functionality of storage in order to store it somewhere different

elvise:
@etobella really great PR! 
Do you have a plan to finish it ?

etobella:
Well, it was something I did on a sprint code. If you are imterested you can finish it 😉

dms_storage/models/dms_storage.py
len-foss: You can setup the attachments to be stored on a storage backend, so it doesn't seem to restrict any functionality.

pedrobaeza:
@etobella can you rebase and fix pre-commit and let's merge this for not losing it, being a new module, it doesn't hurt.

pedrobaeza (approved):
/ocabot merge patch

PR #183, opened 2022-05-22
[13.0][IMP] dms: Don't restrict model_id as it can be False

rousseldenis:
As model_id can be False (directory can be not linked to any model):

https://github.com/OCA/dms/blob/13.0/dms/models/directory.py#L307

This had been introduced there:

https://github.com/OCA/dms/commit/7cc0d07dcd04e3c7dc245cbd0d4a6b4795b90b8e

@victoralmau

### DIFF
--- a/dms/views/directory.xml
+++ b/dms/views/directory.xml
@@ -478,7 +478,7 @@
                         <field name="allowed_model_ids" invisible="True" />
                         <field
                             name="model_id"
-                            attrs="{'invisible': [('storage_id_save_type', '!=', 'attachment')], 'readonly': [('count_total_files', '>', 0)], 'required': [('storage_id_save_type','=','attachment')]}"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But if storage is `attachment`, then forcefully it should be filled.

rousseldenis:
> But if storage is `attachment`, then forcefully it should be filled.

Mmmh is it possible filestore storage was not available in first DMS versions?

pedrobaeza:
Storage types have always been there, but the specific `attachment` one was introduced later. Anyways, there can be the chance that the storage type was changed in the middle of the lifecycle of such storage?

rousseldenis:
My point is in previous versions you could add DMS files to attachment storage without model linked to.

And in Odoo standard, you can create attachments without model.

pedrobaeza:
> And in Odoo standard, you can create attachments without model.

There's no sense for that here in the DMS context, as its usage is that it gathers the attachments of a record in the DMS folder. It's not for saving the contents of the file as orphan attachment. It's always linked to any record.

PR #185, opened 2022-05-23
[13.0][FIX] dms: Apply sudo() when downloading the files (portal users do not have access to ir.attachment for example).

victoralmau:
Apply `sudo()` when downloading the files (portal users do not have access to ir.attachment for example).

@Tecnativa TT36723

### DIFF
--- a/dms/controllers/portal.py
+++ b/dms/controllers/portal.py
@@ -182,7 +182,7 @@ class CustomerPortal(CustomerPortal):
             else:
                 return request.redirect("/my")
 
-        dms_file_sudo = res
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't this granting access to everyone?

victoralmau:
> Isn't this granting access to everyone?

No, the files are only displayed in the portal if we (the user) have access to the folder that contains them (or with access_token to the folder).

chienandalu (approved):
I prefer this approach as it relies on the record rules and not in the public link for everyone.

pedrobaeza (approved):

PR #190, opened 2022-07-15
[FIX] dms: do_warn method no longer available in v15

gdrius:
When file is dragged&dropped into the DMS, it should show warning if no directory is selected. Instead of warning, an error is thrown, because of do_warn, which was not replace with displayNotification when migrating to v15.


etobella (approved):
Thanks! do_warn was removed and we didn't notice that :wink:

etobella:
Fast tracking, as it fixes and issue

/ocabot merge patch

PR #193, opened 2022-08-31
[12.0][IMP] DMS: Change Access Rights for dms.access.group

HekkiMelody:
Proposing a change in the permissions for DMS Access Groups:

- Not every user can create new access groups, but only dms_users.
- DMS managers (not admin users) can manage every access group
- Add tests

After this change:
- the base user has read only access
- the group_dms_user can create, and can write and unlink their own groups
- the group_dms_manager can do everything on every group.


HekkiMelody:
How does it actually have less code coverage than before :sob:

etobella:
Don't worry.

As you can see the branch 12.0 has the following coverage https://app.codecov.io/gh/OCA/dms/branch/12.0 and it is lower than yours :smile:

PR #199, opened 2022-09-03
[14.0][OU-FIX] dms: Don't crash when migrating from 13.0.8.0.0

pedrobaeza:
When coming from 13.0 latest version, the conversion of the thumbnails has already been done, so this is failing saying:

psycopg2.errors.UndefinedColumn: column "custom_thumbnail" does not exist

@Tecnativa


pedrobaeza:
Tested on a real migration and working, so merging.

PR #200, opened 2022-09-05
[14.0][IMP] dms: Refactor tests. 

victoralmau:
Refactor tests. 

Please @pedrobaeza could you review it?

@Tecnativa


dms/tests/test_portal.py
```diff
@@ -14,7 +14,19 @@ def setUp(self):
         self.user_admin = self.env.ref("base.user_admin")
         self.file_text = self.env.ref("dms.file_13_demo")
         self.model_partner = self.env.ref("base.model_res_partner")
-        storage = self.env.ref("dms.storage_attachment_demo")
+        # We need to create a new storage and directory because records may already
+        # exist (e.g. if l10n_en_aeat has been installed).
+        storage = self.env.ref("dms.storage_attachment_demo").copy(
```
pedrobaeza: Don't use demo data

dms/tests/test_storage_attachment.py
```diff
@@ -57,6 +57,9 @@ def setUp(self):
                 ],
             }
         )
+        self.env.ref("base.group_private_addresses").write(
```
pedrobaeza: Why this? I don't think it's needed.

dms/tests/test_portal.py
```diff
@@ -14,7 +14,19 @@ def setUp(self):
...
+        )
+        self.env["dms.directory"].create(
+            {
+                "name": "Test partners",
+                "is_root_directory": True,
+                "storage_id": storage.id,
+                "res_model": self.model_partner.model,
+            }
+        )
         self.partner = self.env.ref("base.partner_demo_portal")
```
pedrobaeza, line 14: Don't use demo data

pedrobaeza (changes requested):

dms/tests/test_portal.py
victoralmau: Almost all tests are based on demo data and users with specific permissions, I think this should be changed, what do you think?

pedrobaeza: Ok

victoralmau:
Changes done and code refactor.

pedrobaeza (approved):

chienandalu (approved):

PR #205, opened 2022-09-08
[14.0][FIX] dms: Searchpanel from directories

victoralmau:
We need to overwrite function from directories because odoo only return records with childs (very weird for user perspective). All records are returned now.

Related to: https://github.com/OCA/dms/pull/122#issuecomment-1240850412

**Before**
dms-antes

**After**
dms-despues

@Tecnativa

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -682,3 +682,26 @@ class DmsDirectory(models.Model):
         if self.child_directory_ids:
             self.child_directory_ids.unlink()
         return super().unlink()
+
+    @api.model
+    def _search_panel_domain_image(
+        self, field_name, domain, set_count=False, limit=False
+    ):
+        """We need to overwrite function from directories because odoo only return
+        records with childs (very weird for user perspective).
+        All records are returned now.
+        """
+        if field_name == "parent_id":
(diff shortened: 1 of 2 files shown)


dms/views/dms_file.xml
```diff
@@ -71,8 +71,8 @@
                     <field
                         name="directory_id"
                         icon="fa-folder"
-                        context="{'directory_short_name': True}"
```
pedrobaeza, line 74: Why removing this?

dms/models/directory.py
```diff
@@ -682,3 +682,26 @@ def unlink(self):
         if self.child_directory_ids:
             self.child_directory_ids.unlink()
         return super().unlink()
+
+    @api.model
+    def _search_panel_domain_image(
```
pedrobaeza, line 687: Isn't any other method to override for this?

dms/views/dms_file.xml
victoralmau: I have separated in several commits the changes to understand it.

pedrobaeza (approved):

chienandalu (approved):

PR #206, opened 2022-09-12
[15.0][FIX] dms: Searchpanel from directories

victoralmau:
FWP from 14.0: https://github.com/OCA/dms/pull/205

We need to overwrite function from directories because odoo only return records with childs (very weird for user perspective). All records are returned now.

Related to: https://github.com/OCA/dms/pull/122#issuecomment-1240850412

**Before**
dms-antes

**After**
dms-despues

@Tecnativa

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -698,3 +698,26 @@ class DmsDirectory(models.Model):
         if self.child_directory_ids:
             self.child_directory_ids.unlink()
         return super().unlink()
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

chienandalu (approved):

pedrobaeza:
GitHub is having problems:

/ocabot merge minor

PR #208, opened 2022-09-19
[15.0][FIX] dms: skip attachment creation when storage type is different from attachment

marcelsavegnago:
cc @pedrobaeza @victoralmau


pedrobaeza (approved):

pedrobaeza:
We can fast track this one:

/ocabot merge patch

PR #209, opened 2022-09-19
[14.0][IMP] dms_field: adds user guidance on storage selection wizard

marcelsavegnago:
I don't know if this PR makes sense for everyone but it would help the end user.


pedrobaeza:
Isn't this something to be put on the help of the field `storage`?

Issue #214, opened 2022-10-11
[12.0] dms directory: Unlink method fails for recordsets with length > 1


HekkiMelody:
## Module
dms

## Describe the bug
The unlink method is not safe to unlink more than one directory at the same time. Additionally, even if one deletes one directory at a time, the method 

https://github.com/OCA/dms/blob/21754ced4f6156846717c5a54076cf9d17983b68/dms/models/directory.py#L675-L684

## To Reproduce

1. Install with demo data. 
2a. Try to unlink any directory with more than one child directory
2b. Try to unlink more than one directory at the same time

Checked on Odoo 12

PR #215, opened 2022-10-11
[12.0][FIX]directory: fix unlink with multiple records

HekkiMelody:
#214


pedrobaeza (approved):
On v12 this is needed

PR #216, opened 2022-10-11
[14.0][FIX]directory: fix unlink with multiple records

HekkiMelody:
#214 
Forward port of #215


pedrobaeza:
This is not needed, as relational fields admits more than one record in the source, not needing mapped

Issue #212, opened 2022-09-26
[14.0] dms_field: Error trying to open directory


marcelsavegnago:
Error trying to open directory:

![image](https://user-images.githubusercontent.com/3595132/192323063-62dada8b-aa8b-4cda-afc6-55ab345c8456.png)

![image](https://user-images.githubusercontent.com/3595132/192323146-9d068867-d8d4-4eaf-a53c-daa0857aa9a6.png)

PR #217, opened 2022-10-23
[FIX] Fix opening a directory using the right click open action #212

Zodk2:

Impacted versions:

    12.0 and above

Steps to reproduce:

    Create a new root directory
    Right-click and choose `Open`

Current behavior:

Expected behavior:

### DIFF
--- a/dms_field/static/src/js/base/dms_tree_renderer.js
+++ b/dms_field/static/src/js/base/dms_tree_renderer.js
@@ -288,7 +288,7 @@ odoo.define("dms.DmsTreeRenderer", function(require) {
                 icon: "fa fa-external-link",
                 label: _t("Open"),
                 action: function() {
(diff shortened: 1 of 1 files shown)


Zodk2:
> The root directory can't be opened using the contextual action menu (`open`)
> 
> Impacted versions:
> 
> ```
> 12.0 and above
> ```
> 
> Steps to reproduce:
> 
> ```
> Go to "DMS directories" in the partner form view
> Create a new root directory
> Right-click and choose `Open`
> ```
> 
> Current behavior:
> 
> ```
> Odoo raises an error `TypeError: record is null`
> ```
> 
> Expected behavior:
> 
> ```
> Open the directory like if we click on the `Open` button in the right section (Document preview)
> ```
Please apply the "no stale" label.

victoralmau:
This is related to `dms_field` module, and happens from version 12.0 up to and including 15.0.

PR #224, opened 2023-01-25
[13.0][FIX] dms: Avoid error when accessing a subdirectory without access token

victoralmau:
Misc changes:
- [x] Avoid error when accessing a subdirectory without access token
- [x] Use the correct domain with inherit_access_from_parent_record field

@Tecnativa

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -260,7 +260,11 @@ class DmsDirectory(models.Model):
         current_directory = self
         while current_directory:
             directories.insert(0, current_directory)
-            if access_token and consteq(current_directory.access_token, access_token):
+            if (
+                access_token
+                and current_directory.access_token
+                and consteq(current_directory.access_token, access_token)
+            ):
                 return directories
             current_directory = current_directory.parent_id
         if access_token:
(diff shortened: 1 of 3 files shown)


dms/models/directory.py
```diff
@@ -260,7 +260,11 @@ def _get_parent_categories(self, access_token):
         current_directory = self
         while current_directory:
             directories.insert(0, current_directory)
-            if access_token and consteq(current_directory.access_token, access_token):
+            if (
```
pedrobaeza, line 263: But is this taking into account subdirectories of a shared directory?

victoralmau: Of course, if you share a directory, you will have access to subdirectories (and the breadcrumb will only show the directories from which you have access).

pedrobaeza: I mean the following scenario:

- You share via URL with token a directory.
- That directory has subdirectories that are seen on the portal.
- You click on a subdirectory.

Do you see the contents of such subdirectory?

victoralmau: Yes, a small example (similar to any other system, if you share something you share EVERYTHING in it).

dms

pedrobaeza (approved):

chienandalu (approved):

PR #226, opened 2023-02-07
[13.0][FIX] dms: Misc changes.

victoralmau:
Use case:
- Create a "Documents" directory with access to Admin user.
- Create a "Code" directory with read-only access to the Test (new user) user.
- Create a file in "Code" folder.

Remove `parent_id` defined if you do not have access to it search panel directory.
Before:
antes-1

After:
despues-1

Get the "parent" directories (those of the top level to which you have access) in the portal view.
Before:
antes-2

After:
despues-2

Limit breadcrumb from portal to directories you have access to.

@Tecnativa TT41585

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -273,11 +273,15 @@ class DmsDirectory(models.Model):
         return directories
 
     def _get_own_root_directories(self):
-        return (
-            self.env["dms.directory"]
-            .search([("is_hidden", "=", False), ("parent_id", "=", False)])
-            .ids
+        res = self.env["dms.directory"].search_read(
+            [("is_hidden", "=", False)], ["parent_id"]
         )
+        all_ids = [value["id"] for value in res]
+        res_ids = []
+        for item in res:
+            if not item["parent_id"] or item["parent_id"][0] not in all_ids:
+                res_ids.append(item["id"])
+        return res_ids
 
     allowed_model_ids = fields.Many2many(
         related="storage_id.model_ids",
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -261,9 +261,13 @@ class DmsDirectory(models.Model):
         while current_directory:
             directories.insert(0, current_directory)
             if (
-                access_token
-                and current_directory.access_token
-                and consteq(current_directory.access_token, access_token)
+                (
+                    access_token
+                    and current_directory.access_token
+                    and consteq(current_directory.access_token, access_token)
+                )
+                or not access_token
+                and current_directory.check_access_rights("read")
             ):
                 return directories
             current_directory = current_directory.parent_id
--- a/dms/models/abstract_dms_mixin.py
+++ b/dms/models/abstract_dms_mixin.py
@@ -44,5 +44,10 @@ class AbstractDmsMixin(odoo.models.AbstractModel):
         models.SEARCH_PANEL_LIMIT = False
         _self = self.with_context(directory_short_name=True)
         res = super(AbstractDmsMixin, _self).search_panel_select_range(field_name)
+        all_ids = [value["id"] for value in res["values"]]
+        # Prevent error if user not access to parent record
+        for value in res["values"]:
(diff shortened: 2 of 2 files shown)


dms/models/directory.py
```diff
@@ -273,10 +280,13 @@ def _get_parent_categories(self, access_token):
...
 
     def _get_own_root_directories(self):
+        domain = [("is_hidden", "=", False)]
+        model = self.env["dms.directory"]
+        item = model.search(domain, limit=1, order="level_depth asc")
         return (
-            self.env["dms.directory"]
-            .search([("is_hidden", "=", False), ("parent_id", "=", False)])
-            .ids
+            model.search(domain + [("level_depth", "=", item.level_depth)]).ids
```
pedrobaeza: Why do you need the level_depth?

victoralmau: This is necessary for an easy search.

Example:
- Documents
- Documents > Code

If the user only has access to Code, we need to list "top" level directories to which the user has access.

pedrobaeza: Yes, but I don't see the relation... This is something dynamic, not static, depending on your security grants.

victoralmau: But with the search method the records are already being filtered according to your permissions.

In this case it is not a permissions problem, we just need to list the correct records, nothing else.

pedrobaeza: That approach is not correct, as you can have access to level 0 directory, but then in another tree, to a level 2 directory. Both should be shown as top level directories for that user.

victoralmau: Ok, in that case, how do you suggest to do it in an optimal way and with the search method?

pedrobaeza: Wasn't this solved in portal view to see all the possible directories? How was this done in v11 by MuK?

victoralmau: Portal view did not exist in Muk v11, I have changed the approach and solved it in a different way, achieving what you indicated.

dms/models/directory.py
```diff
@@ -273,11 +273,14 @@ def _get_parent_categories(self, access_token):
         return directories
 
     def _get_own_root_directories(self):
-        return (
-            self.env["dms.directory"]
-            .search([("is_hidden", "=", False), ("parent_id", "=", False)])
-            .ids
+        res = self.env["dms.directory"].search_read(
+            [("is_hidden", "=", False)], ["display_name", "parent_id"]
```
pedrobaeza: Why are you reading `display_name`?

dms/models/directory.py
```diff
@@ -273,11 +273,14 @@ def _get_parent_categories(self, access_token):
...
         )
+        all_ids = [value["id"] for value in res]
+        for index, item in enumerate(res):
+            if item["parent_id"] and item["parent_id"][0] in all_ids:
```
pedrobaeza: Aren't this keeping subdirectories of third-level?

dms/models/directory.py
victoralmau: Right, not really necessary, deleted.

dms/models/directory.py
victoralmau: This will delete subdirectories if we have access to the parent directory.

Example:
ejemplo-1
ejemplo-2

pedrobaeza: But you can see the problem in your screenshot: "2023 (level 3)" shouldn't be displayed there. Add the previously removed folders in the list / another list for being taking into account in successive removals.

victoralmau: Sorry, but what the screenshot shows is what you indicated at https://github.com/OCA/dms/pull/226#discussion_r1098947571, right?

pedrobaeza: True, as "Slides" is not allowed, but please put "2023 (level 3)" inside "Code (level 2)". Only code should be shown, but if I read your code, there may be the possibility of both being shown.

pedrobaeza (approved):

pedrobaeza:
Patching this:

/ocabot merge patch

chienandalu (approved):

PR #228, opened 2023-02-10
[13.0][IMP] dms_attachment_link: Add searchpanel to wizard to improve file selection.

victoralmau:
Add searchpanel to wizard to improve file selection.
ejemplo

@Tecnativa TT41512


pedrobaeza:
This is a better selection picker indeed! Do you think this "selector" can be part of `dms` base module for beeing a tool for anyone needing to use it?

victoralmau:
Yes, it would probably be interesting to add it to `dms`. Besides simplifying this module, this tool can be used by anyone for other use cases.

chienandalu (approved):
Works like charm :)

pedrobaeza:
Are you going to extract it into `dms` finally?

victoralmau:
Yes, let me have a few minutes, I'm working on it now.

pedrobaeza:
OK, the idea is to have the basic JS + view, and put in the README the instructions to override the view to plug the method/action into the button + call it in dependent modules.

victoralmau:
Changes done (added to `dms` part of the code that can be extended by other modules, including this one).

chienandalu (approved):

dms_attachment_link/views/dms_file_views.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="view_dms_file_kanban_wizard_selection" model="ir.ui.view">
+        <field name="name">dms_file.kanban.wizard.selection.attachment.link</field>
+        <field name="model">dms.file</field>
+        <field name="inherit_id" ref="dms.view_dms_file_kanban_wizard_selection" />
+        <field name="arch" type="xml">
```
pedrobaeza, line 9: This view should be primary for having multiple inheritance.

victoralmau: I think it is not necessary, the template that is primary is `dms.view_dms_file_kanban_wizard_selection`, and in `dms` there is also `action_dms_file_wizard_selector`. The `dms_attachment_link` module just adds the button to the already existing `dms` view.

pedrobaeza: Exactly, but if you add a new module using this file selector, this button "Add" will appear also, and we don't want that.

victoralmau: Ok, changes done.

pedrobaeza (approved):
/ocabot merge minor

PR #225, opened 2023-01-26
[14.0][FIX] dms: Use the correct domain with inherit_access_from_parent_record field

victoralmau:
FWP from 13.0: https://github.com/OCA/dms/pull/224

Misc changes:
- [x] Avoid error when accessing a subdirectory without access token (already merged at https://github.com/OCA/dms/commit/e576bb587969c119df02d6fa0edd0554e366e6a8)
- [x] Use the correct domain with inherit_access_from_parent_record field

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa

PR #231, opened 2023-02-23
[15.0][FIX] dms: Use the correct domain with inherit_access_from_parent_record field

victoralmau:
FWP from 14.0: https://github.com/OCA/dms/pull/225

Use the correct domain with inherit_access_from_parent_record field

@Tecnativa


pedrobaeza (approved):
Let's fast-track it for finishing the fw-port chain:

/ocabot merge minor

PR #236, opened 2023-04-03
[13.0][FIX] dms: dms.file thumbnail generation checks for supported mimetypes 

HekkiMelody:
The method _compute_image_1920 of dms_file checks if the file is of mimetype image and tries to use it as a thumbnail. However, there are several types of image files that Odoo/PIL cannot work with, leading to the following error:

```
  File "/opt/odoo/custom/src/odoo/odoo/tools/image.py", line 404, in base64_to_image
    return Image.open(io.BytesIO(base64.b64decode(base64_source)))
  File "/usr/local/lib/python3.6/site-packages/PIL/Image.py", line 2687, in open
    % (filename if filename else fp))
OSError: cannot identify image file <_io.BytesIO object at 0x7f0a6ecb1780>
```

and also further down

```
odoo.exceptions.UserError: ('This file could not be decoded as an image file. Please try with a different file.', '')
```

Examples of such files are CAD files (mimetype: image/vnd.dwg) For a list of all mimetypes starting with "image/" see: https://www.iana.org/assignments/media-types/media-types.xhtml#image

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -9,6 +9,8 @@ import json
 import logging
 from collections import defaultdict
 
+from PIL import Image
+
 from odoo import _, api, fields, models, tools
 from odoo.exceptions import ValidationError
 from odoo.osv import expression
@@ -143,7 +145,10 @@ class File(models.Model):
     def _compute_image_1920(self):
         """Provide thumbnail automatically if possible."""
         for one in self.filtered("res_mimetype"):
-            if one.res_mimetype.startswith("image/"):
+            # Image.MIME provides a dict of mimetypes supported by Pillow,
+            # SVG is not present in the dict but is also a supported image format
+            # lacking a better solution, it's being added manually
+            if one.res_mimetype in (*Image.MIME.values(), "image/svg+xml"):
                 one.image_1920 = one.content
 
     def check_access_rule(self, operation):


HekkiMelody:
The pre-commit error seems unrelated to my PR. Unsure how to fix it...

pedrobaeza:
I think we should instead restrict the mimetypes to the ones supported.

pedrobaeza:
And next time, it's better to only propose the PR for one branch, and wait until the feedback and being merged for not having to modify 3 PRs at the same time (although thank you very much for caring about all the versions :smiley:)

HekkiMelody:
> I think we should instead restrict the mimetypes to the ones supported.

I'll try to find that out and implement it. Thanks for the feedback

HekkiMelody:
Having a fair amount of trouble finding any sort of exhaustive list of mimetypes pillow supports.

I think the best I can do is to start with a conservative list of image types that are for sure supported and then maybe it can be expanded in the future. (I also don't necessarily have all these files using different extensions at hand to easily test the PR with)

pedrobaeza:
Ok, then maybe the try/except approach is better having that uncertainty. My only fear was the memory consumption, as for example DWG files are usually very heavy.

pedrobaeza:
I'm seeing that you can query the supported mimetypes seeing if included in `Image.MIME`. That can be other way.

HekkiMelody:
> Ok, then maybe the try/except approach is better having that uncertainty. My only fear was the memory consumption, as for example DWG files are usually very heavy.

I agree. Maybe a size limit could work instead? Don't try to generate a thumbnail if the image is bigger than 20MB?

HekkiMelody:
> I'm seeing that you can query the supported mimetypes seeing if included in `Image.MIME`. That can be other way.

I'll check that out.

HekkiMelody:
The list of mimetypes in the registry is surprisingly small. In Odoo shell on v13, python 3.6 and Pillow 5.4.1:

```
>>> from PIL import Image
>>> Image.MIME
{'BMP': 'image/bmp', 'PNG': 'image/png', 'GIF': 'image/gif', 'TIFF': 'image/tiff', 'JPEG': 'image/jpeg'}
```

Thanks a lot for the pointers, by the way!

pedrobaeza (approved):
I see... The advantage of this method is that if PIL evolves for further formats, we will be covered without doing any change in our code.

I would bet that latest versions have support for WEBP and other formats.

victoralmau (approved):

pedrobaeza:
@victoralmau can you check about pre-commit in other PR?

pedrobaeza:
I'm wondering now what happens with SVG files. Have you checked them?

HekkiMelody:
(Sorry, didn't have time to look at this this week, will pick it back up asap)

pedrobaeza:
You can rebase now to have green CI. Please check the SVG part.

HekkiMelody:
> I'm wondering now what happens with SVG files. Have you checked them?

I can confirm that if you upload an SVG file, the preview is not automatically generated. However, SVG files do work if they're used manually on the image_1920 field.

Whereas, with a .png file, the preview is automatically set when uploading the file itself.

Clearly, `Image.MIME` doesn't actually have all the possible mimetypes that Pillow can handle.

pedrobaeza:
Well, I think we can start this with such list + SVG manually added to the supported ones.

HekkiMelody:
> Well, I think we can start this with such list + SVG manually added to the supported ones.

Done. Thank you again for the support with this issue.

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to upper versions.

PR #239, opened 2023-04-04
[13.0][FIX] Update copier template

pedrobaeza:
To avoid interpreter problem.

@Tecnativa


pedrobaeza:
It doesn't seem to be this...

rvalyi (approved):

pedrobaeza:
There's seems to be something more.

rvalyi:
last time I got this problem, that was isort that needed an update.

pedrobaeza:
In theory, the latest copier template includes that patches, but the error continues the same.

PR #240, opened 2023-04-10
[13.0][IMP] Update copier template

victoralmau:
Update copier template

@Tecnativa

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.14.0
+_commit: v1.14.2
 _src_path: https://github.com/OCA/oca-addons-repo-template.git
 ci: GitHub
 dependency_installation_mode: PIP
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I tried this in #239 and it didn't seem enough. Let's see what happens now.

HekkiMelody:
isn't this the infamous isort issue?

https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co

https://github.com/OCA/oca-addons-repo-template/pull/188

Think it was only fixed for v14 and onwards

pedrobaeza:
The isort version change is already reflected in the template for v13: https://github.com/OCA/oca-addons-repo-template/blob/fe5b6f5c35c3d13822d080a9c9e20df209f3a6c5/src/.pre-commit-config.yaml.jinja#L11, but copier is not updating such change. We have already opened an issue in the repository for the same problem: https://github.com/copier-org/copier/issues/1100. Please support it for getting more attention in it.

pedrobaeza:
Manually changed the isort version, so merging.

PR #241, opened 2023-04-24
[14.0] [IMP] dms: more restrictive access rights for access groups

HekkiMelody:
Change in the permissions for DMS Access Groups:

- Not every user can create new access groups, but only dms_users.
- DMS managers (not admin users) can manage every access group
- Add tests

After this change:

- the base user has read only access
- the group_dms_user can create, and can write and unlink their own groups
- the group_dms_manager can do everything on every group.

### DIFF
--- a/dms/security/ir.model.access.csv
+++ b/dms/security/ir.model.access.csv
@@ -19,4 +19,5 @@ access_dms_file_user,dms_file_user,model_dms_file,group_dms_user,1,1,1,1
 
 access_dms_access_group_public,access_dms_access_group_public,model_dms_access_group,base.group_public,1,0,0,0
 access_dms_access_group_portal,access_dms_access_group_portal,model_dms_access_group,base.group_portal,1,0,0,0
-access_security_access_groups_user,access_security_access_groups_user,model_dms_access_group,base.group_user,1,1,1,1
+access_security_access_groups_user,access_security_access_groups_user,model_dms_access_group,base.group_user,1,0,0,0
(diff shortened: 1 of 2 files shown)


victoralmau (commented):
The change in ACLs makes sense but I think the tests are not necessary (specially in `multi_users()`).

HekkiMelody:
> The change in ACLs makes sense but I think the tests are not necessary (specially in `multi_users()`).

Could you explain why not?

victoralmau:
> > The change in ACLs makes sense but I think the tests are not necessary (specially in `multi_users()`).
> 
> Could you explain why not?

I think it is not necessary because we are reviewing that Odoo allows or does not allow to do what we have defined in the ACLs (something redudant) and it does not add too much, but it is just my point of view.

HekkiMelody:
> > > The change in ACLs makes sense but I think the tests are not necessary (specially in `multi_users()`).
> > 
> > 
> > Could you explain why not?
> 
> I think it is not necessary because we are reviewing that Odoo allows or does not allow to do what we have defined in the ACLs (something redudant) and it does not add too much, but it is just my point of view.

No, that's fair. Always interested to see what people think should and should not be tested, is all.

victoralmau (approved):

pedrobaeza (approved):
@aleuffre take into account that at the end, tests are also code to maintain, so you have to keep it focused on what your module does and keep it minimal, relying that the framework tools behind are already checking that they work.

/ocabot merge minor

pedrobaeza:
Please forward port it to upper versions.

PR #245, opened 2023-05-28
[FIX] Opening a directory using the right click open action

Zodk2:

Impacted versions:

    12.0 and above

Steps to reproduce:

    Create a new root directory
    Right-click and choose `Open`

Current behavior:

Expected behavior:

### DIFF
--- a/dms_field/static/src/js/base/dms_tree_controller.js
+++ b/dms_field/static/src/js/base/dms_tree_controller.js
@@ -408,7 +408,7 @@ odoo.define("dms.DmsTreeController", function(require) {
                 ? "directory_" + directory.parent_id[0]
                 : "#";
             var directoryNode = {
-                id: "directory_" + directory.id,
+                id: dt.id,
                 text: directory.name,
                 icon: "fa fa-folder-o",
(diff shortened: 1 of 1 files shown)


Zodk2:
@etobella Please, can you verify why the build was not successful

etobella:
Probably we need to update the template of dms...

Zodk2:
@etobella Please can you check why **_runboat/build_** is pending?

etobella:
Just rebase :wink:

Zodk2:
Why has this pull request never been verified unless I ask (7 months)?

etobella (approved):

etobella:
Because DMS is a repo with not  too many maintainers. In my case, I am already following a lot of repos and cannot manage everything. Obviously, this is an OpenSource community and it depends on the community work, if there is not a lot of people involved to a repo, something like this might happen.

Anyway, you are right, and this needs to be merged, I see other problems on 12 version, but are not targetted by this change

/ocabot merge patch

Zodk2:
@etobella Thank you so much

PR #248, opened 2023-06-15
[13.0][FIX] dms: Explicitly exclude the pdf type in the image_1920 field of the files.

victoralmau:
Explicitly exclude the pdf type in the `image_1920` field of the files.

Please @pedrobaeza can you review it?

@Tecnativa TT43563

PR #250, opened 2023-06-19
[14.0][FIX] dms: Explicitly exclude the pdf type in the image_1920 field of the files

victoralmau:
FWP from 13.0: https://github.com/OCA/dms/pull/248

Explicitly exclude the pdf type in the `image_1920` field of the files.

@Tecnativa TT43563


pedrobaeza:
I think we are not going for now to put this patch, as the side effect in v13 may not be applicable here.

Issue #255, opened 2023-07-17
error message: This file could not be decoded as an image file.


ccandito:
Hi, 
in uploading a pdf file the following error has occurred:
This file could not be decoded as an image file.

Regards
Claudio Candito

pedrobaeza:
Please reproduce the problem in runboat and put the steps to reproduce and the image. Make sure you have latest code, and include the Odoo version.

ccandito:
Odoo version 15
I've created a file and tried to change it by assigning another file.
Regards
Claudio Candito

ccandito:
the image of the error
Thanks in advance
[error.docx](https://github.com/OCA/dms/files/12072937/error.docx)

pedrobaeza:
@victoralmau it seems c0179b2e630db7d37504091f3487803455932004 is also needed on v14/v15

ccandito:
By applying the above correction the issue is solved
Regards
Claudio Candito

pedrobaeza:
Applying them on PRs now

ccandito:
Before implementing it, it is necessary to change the correction as follows:

if one.mimetype != "application/pdf" and one.res_mimetype in
(*Image.MIME.values(), "image/svg+xml"):


regards

Claudio

Il giorno mar 18 lug 2023 alle ore 10:22 Pedro M. Baeza <
***@***.***> ha scritto:

victoralmau:
Fixed already in 15.0 with https://github.com/OCA/dms/pull/258

PR #258, opened 2023-07-18
[15.0][FIX] dms: Explicitly exclude the pdf type in the image_1920 field of the files

victoralmau:
Related to https://github.com/OCA/dms/issues/255

FWP from 14.0: https://github.com/OCA/dms/pull/256

Explicitly exclude the pdf type in the `image_1920` field of the files.

@Tecnativa TT43563


pedrobaeza (approved):
/ocabot merge patch

PR #259, opened 2023-09-01
[14.0][IMP]dms: check for same storage

PicchiSeba:
Check if the storage is changed when both `storage_id` and `parent_id` are passed to the write method

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -78,6 +78,10 @@ class DmsDirectory(models.Model):
         default=lambda self: self._default_parent_id(),
     )
 
+    root_directory_id = fields.Many2one(
+        "dms.directory", "Root Directory", compute="_compute_root_id", store=True
+    )
+
     def _default_parent_id(self):
         context = self.env.context
         if context.get("active_model") == self._name and context.get("active_id"):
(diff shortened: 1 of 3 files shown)


dms/models/directory.py
```diff
@@ -661,17 +661,27 @@ def create(self, vals_list):
...
+                new_storage_id = (
+                    vals["storage_id"]
+                    if "storage_id" in vals.keys()
+                    else item.storage_id.id
+                )
+                new_parent_id = (
+                    vals["parent_id"]
+                    if "parent_id" in vals.keys()
+                    else item.parent_id.id
+                )
```
```suggestion
                new_storage_id = vals.get("storage_id", item.storage_id.id)
                new_parent_id = vals.get("parent_id", item.parent_id.id)
```

dms/models/directory.py
```diff
@@ -661,17 +661,27 @@ def create(self, vals_list):
...
+                old_storage_id = (item.storage_id or item.parent_id.storage_id).id
```
pedrobaeza: What happens if the directory is a third, fourth or so in depth level? I think there's a root_directory_id or so that should be used.

dms/models/directory.py
```diff
@@ -661,17 +661,27 @@ def create(self, vals_list):
...
+                if new_parent_id:
+                    if old_storage_id != self.browse(new_parent_id).storage_id.id:
+                        raise UserError(
+                            _("It is not possible to change parent to other storage.")
+                        )
+                else:
```
pedrobaeza: No need of the else.

pedrobaeza (commented):
Thanks for the patch.

pedrobaeza (approved):

HekkiMelody (approved):
Code review, LGTM

pedrobaeza:
/ocabot merge minor

You can fw-port it to upper versions.

PR #262, opened 2023-09-18
[16.0][MIG] dms

vancouver29:
[16.0][MIG] dms


pedrobaeza:
Please check CI

len-foss:
Works great!

For the CI tests it's only minor issues, the fixes are in commit https://github.com/len-foss/dms/commit/fefc351c0e186fd5f6aab7c16203fc8aad780c98 and https://github.com/len-foss/dms/commit/59b97732c0ac9157e7d73eb5c1afc4eeb569ad72 

The thing that would be more interesting would be to split the commit to keep track of actual changes (in particular since there are still some TODOs left, that would make finishing up the work easier, and localize fixups to these TODO commits). Let me know if you want help with that.

vancouver29:
> Works great!
> 
> For the CI tests it's only minor issues, the fixes are in commit [len-foss@fefc351](https://github.com/len-foss/dms/commit/fefc351c0e186fd5f6aab7c16203fc8aad780c98) and [len-foss@59b9773](https://github.com/len-foss/dms/commit/59b97732c0ac9157e7d73eb5c1afc4eeb569ad72)
> 
> The thing that would be more interesting would be to split the commit to keep track of actual changes (in particular since there are still some TODOs left, that would make finishing up the work easier, and localize fixups to these TODO commits). Let me know if you want help with that.

@len-foss Great :smiley: ! I will add your patch for the Tests and start it again.  

I used to have many commits in a PR, but it does not conform to the OCA Mig. convention :smiling_face_with_tear: that all changes should be in one commit e.g., `[MIG] dms: Migration from 15.0 to 16.0`. Should OCA consider that in the future? @pedrobaeza  

The TODO points in the manifest are actually not needed here, since we don't need the other 2 OCA modules, `web_drop_target` and `mail_preview_base`. But maybe it would be nicer if we had these modules in the future :)

vancouver29:
@len-foss, thank you so much for your contribution :) The tests have passed successfully!

len-foss:
I don't really get this rule, since I often ended up doing a lot of forward and backporting, so maximizing commit portability is a great help. But let's not derail the conversation here ;-)

The TODO I was referring are the ones in the view_dms_directory_manager_form, and it's such an odd pattern that I wasn't sure how to migrate.

dms/views/directory.xml
```diff
@@ -0,0 +1,736 @@
...
+            <field name="is_root_directory" position="attributes">
+                <attribute name="invisible">0</attribute>
+                <!-- TODO Fix Update Error
+                <attribute name="groups">dms.group_dms_manager</attribute>
+                -->
+            </field>
+            <field name="storage_id" position="attributes">
+                <attribute name="options">{}</attribute>
+                <!-- TODO Fix Update Error
+                <attribute name="groups">dms.group_dms_manager</attribute>
```
len-foss: These should be fixed before merge

len-foss: For the first one, it can be fixed the following way:
```
<field name="is_root_directory" position="after">
    <field name="is_root_directory" groups="dms.group_dms_manager"/>
</field>
```
The original stays invisible, the second one is displayed only for managers so it's the same as before.

For the second one, I don't see any solution that isn't an overkill hack, so unless somebody has a better idea I would just remove it altogether. There isn't a huge usability impact without it.

len-foss (approved):
Besides a quick clean up of the todos everything looks good.
Great work!

dms/views/directory.xml
vancouver29: Thanks for noticing that, and thanks for your solution. With your Solution, we can uncomment the second TODO :+1: Great Work! :) I have just found another bug in the GUI. The patches will be adjusted in a moment.

pedrobaeza:
Please don't misunderstand my indications. The migration commit should be unique, as there's not too much sense to split the needed changes to make this work into several commits. Anyway, it's not a mandatory thing, just a guideline, and it depends on the contributors/reviewers/PSCs. I always prefer the less possible commits though. If the migration is very huge, others have decided to split the changes into several commits, but it's not the usual thing, as that changes separately doesn't bring too much added value from seeing the diff in each file.

Other thing is that doing the migration a patch that applies to other versions or imp is done, and then, it can be interesting to split it in a separate commit for giving a proper explanation on the commit message and being able to cherry-pick it on other versions.

agent-z28:
@vancouver29 i found this two littlt things as of yet

image


image

agent-z28:
@vancouver29  

Related to web_drop_target: 

I use chrome browser Version 105.0.5195.125 (Offizieller Build) for Linux Mint (64-Bit)

When i drag&drop files, basically it works propper. 
But when it drop the file not directly in the center, while it's flickering this happens:

https://github.com/OCA/dms/assets/7473428/af9c3286-2f62-4f44-80cd-804fd7e4f756

Since i did quite good testing, as of yet, i have to say great work from functional side :)

vancouver29:
> @vancouver29 i found this two littlt things as of yet
> 
> image
> 
> image

@agent-z28 That is a good point :smiley: Thank you very much for letting me know. A patch for the described issues is adjusted. I also added a small fix for the GUI.

vancouver29:
> @vancouver29
> 
> Related to web_drop_target:
> 
> I use chrome browser Version 105.0.5195.125 (Offizieller Build) for Linux Mint (64-Bit)
> 
> When i drag&drop files, basically it works propper. But when it drop the file not directly in the center, while it's flickering this happens:
> 
>  drop_error.mp4 
> Since i did quite good testing, as of yet, i have to say great work from functional side :)

I tested it in both Google Chrome and Firefox, but I could not reproduce your bug.  Have you tried to delete the browser cache?

dms/views/directory.xml
```diff
@@ -0,0 +1,731 @@
...
+        <field name="model">dms.directory</field>
+        <field name="inherit_id" ref="view_dms_directory_form" />
+        <field name="arch" type="xml">
+            <field name="is_root_directory" position="after">
+                <field name="is_root_directory" groups="dms.group_dms_manager" />
+            </field>
+            <field name="storage_id" position="attributes">
+                <attribute name="options">{}</attribute>
+                <attribute name="groups">dms.group_dms_manager</attribute>
+            </field>
```
len-foss: That's actually a mistake, because with this change the field is removed from the view altogether if the user isn't in the manager group. 
I think the best is simply to remove these lines.

The impact of removing these lines is that a manager would need to go through the menu to modify storage settings, instead of directly using the internal link (because of the no_open). It's relatively minor.

dms/__manifest__.py
```diff
@@ -4,17 +4,18 @@
...
     "author": "MuK IT, Tecnativa, Odoo Community Association (OCA)",
     "depends": [
-        "web_drop_target",
+        # "web_drop_target", #TODO only exist in v15
         "mail",
         "http_routing",
         "portal",
-        "mail_preview_base",
+        # "mail_preview_base", #TODO only exist in v15
+        "base",
```
victoralmau: If `web_drop_target` and `mail_preview_base` are not needed dependencies can be removed (do not mark them as TODO. 

After testing it functionally in runboat, it seems that they are no longer needed and that the necessary code has been added to `dms`, good job!!!!

**Note**: How would the compatibility be if for example you want to preview an audio with `mail_preview_audio`?

dms/__manifest__.py
```diff
@@ -32,20 +33,18 @@
...
-            "dms/static/src/js/views/search_panel.esm.js",
-            "dms/static/src/js/views/file_list_controller.js",
-            "dms/static/src/js/views/file_list_view.js",
-            "dms/static/src/js/views/file_kanban_controller.js",
-            "dms/static/src/js/views/file_kanban_renderer.js",
-            "dms/static/src/js/views/file_kanban_view.js",
+            "dms/static/src/scss/*",
+            "dms/static/src/js/fields/*",
+            "dms/static/src/js/views/*.esm.js",
+            "dms/static/src/js/views/*.xml",
```
victoralmau, line 42: Is it correct to include it here?

dms/controllers/main.py
```diff
@@ -12,15 +12,15 @@ def document_onboarding_directory(self):
...
+                "html": request.env["ir.qweb"]._render(
+                    request.env.ref("dms.document_onboarding_directory_panel").id,
                     {
                         "state": company.get_and_update_documents_onboarding_state(),
                         "company": company,
-                    }
+                    },
                 )
             }
+
```
victoralmau: Please, don't add extra lines.

dms/models/dms_file.py
```diff
@@ -84,7 +84,7 @@ class File(models.Model):
         relation="dms_file_tag_rel",
         column1="fid",
         column2="tid",
-        domain="['|', ('category_id', '=', False),('category_id', '=?', category_id)]",
+        # domain="['|', ('category_id', '=', False),('category_id', '=?', category_id)]",
```
victoralmau: This would not be correct, domain must be kept or removed by adding a comment in the code explaining why it is not necessary now.

dms/models/dms_file.py
```diff
@@ -132,6 +132,20 @@ class File(models.Model):
...
+        readonly=True,
+        invisible=True,
+    )
+
+    # ----------------------------------------------------------
+    # Compute
+
+    def _compute_default_model_name(self):
+        for record in self:
+            record.model_name = "dms.file"
```
victoralmau: Is this really necessary? I see it is used for .js but maybe you could avoid adding this field by using model.

dms/models/dms_file.py
```diff
@@ -0,0 +1,644 @@
...
+            record.extension = file.guess_extension(
+                record.name, record.mimetype, record.content
+            )
+
+    @api.depends("content")
+    def _compute_mimetype(self):
+        for record in self:
+            binary = base64.b64decode(record.content or "")
+            guess_mimetype(binary)
+            record.mimetype = guess_mimetype(binary)
```
victoralmau:
```suggestion
            record.mimetype = guess_mimetype(binary)
```

dms/tests/common.py
```diff
@@ -227,11 +227,14 @@ def create_storage(self, save_type="database"):
             {"name": "Test Storage", "save_type": save_type}
         )
 
-    def create_directory(self, storage=False, directory=False, res_model=False):
+    def create_directory(self, storage=False, directory=False, model_id=False):
         record = Form(self.directory_model)
         record.name = uuid.uuid4().hex
         record.is_root_directory = True
-        record.res_model = res_model
```
victoralmau: Keep this assignment to confirm that `model_id` is defined correctly and there are no permissions issues.

dms/views/directory.xml
```diff
@@ -0,0 +1,731 @@
...
+            <xpath expr="//sheet//group[@name='settings']" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
+            <xpath expr="//sheet//group[@name='data']" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
+            <xpath expr="//sheet//group[@name='data']" position="after">
+                <group>
+                    <group>
+                        <field name="is_root_directory" readonly="1" />
```
victoralmau:
```suggestion
                        <field name="is_root_directory" readonly="1" groups="!dms.group_dms_manager" />
                        <field name="is_root_directory" groups="dms.group_dms_manager" />
```
The correct technique would be to add the field 2 times (each with the corresponding permissions).

dms/views/directory.xml
```diff
@@ -0,0 +1,731 @@
...
+        <field name="arch" type="xml">
+            <field name="is_root_directory" position="after">
+                <field name="is_root_directory" groups="dms.group_dms_manager" />
+            </field>
+            <field name="storage_id" position="attributes">
+                <attribute name="options">{}</attribute>
+                <attribute name="groups">dms.group_dms_manager</attribute>
+            </field>
+        </field>
+    </record>
```
victoralmau:
```suggestion
```
This view should be deleted according to the previous comment.

dms/views/dms_file.xml
```diff
@@ -0,0 +1,676 @@
...
+                            <field name="mimetype" />
+                        </group>
+                    </group>
+                    <group name="settings">
+                        <group>
+                            <field
+                                name="directory_id"
+                                options="{'no_quick_create': True}"
+                            />
+                            <field name="storage_id" options="{'no_open': True}" />
```
victoralmau:
```suggestion
                            <field name="storage_id" options="{'no_open': True}" groups="!dms.group_dms_manager" />
                            <field name="storage_id" groups="dms.group_dms_manager" />
```

dms/views/dms_file.xml
```diff
@@ -0,0 +1,676 @@
...
+        <field name="name">dms_file.form</field>
+        <field name="model">dms.file</field>
+        <field name="inherit_id" ref="view_dms_file_form" />
+        <field name="arch" type="xml">
+            <field name="storage_id" position="attributes">
+                <attribute name="options">{}</attribute>
+                <attribute name="groups">dms.group_dms_manager</attribute>
+            </field>
+        </field>
+    </record>
```
victoralmau:
```suggestion
```

victoralmau (changes requested):
Very good job, thank you very much for the work done.

Some comments:
- Can you add as a summary all the changes done in the PR description or commit (it would be useful to know for reviewers).
- Can you do Squash administrative commits? (https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate)

Use cases not covered:

1- If we go to the form view of a directory and click on the Files smart-button and drag a file, the file is not added to the directory.
error-1

**Tip**: I think the fix will be to select the correct directory in search_panel.

2- It should not list the selected directory in search_panel.
**v15**:
ejemplo-v15

**v16**:
ejemplo-v16

TT44199

dms/views/directory.xml
pedrobaeza: The technique for this is usually to insert twice the field, once with the group, and another as invisible without group. Check https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

dms/__manifest__.py
len-foss: Your note is an interesting question; if I remember well, the previous version had support for more formats than just pdf, which would include most 'office' documents (docx and the likes), so there's already a loss of interesting functionality with this change. So there may be the need of a module to plug this feature back (which might not be a dependency of dms proper though).

dms/tests/common.py
len-foss: Please check https://github.com/len-foss/dms/commit/fefc351c0e186fd5f6aab7c16203fc8aad780c98
The commit message mentioned the change in Form that ensures that this code snippet, which is wrong, cannot work in v16 and upwards.

dms/__manifest__.py
pedrobaeza: Unfortunately, there has never been support for Office/LibreOffice formats. The module `mail_preview_base` served for clicking on the icon of the kanban view of the attachments/dms files and being able to preview them, but not expanding the supported files (PDF, images, videos and text). `mail_preview_base_audio` expanded the support for the audio files, although I think now there's support for them in core as well.

vancouver29:
> Very good job, thank you very much for the work done.
> 
> Some comments:
> 
> * Can you add as a summary all the changes done in the PR description or commit (it would be useful to know for reviewers).
> * Can you do Squash administrative commits? (https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate)
> 
> Use cases not covered:
> 
> 1- If we go to the form view of a directory and click on the Files smart-button and drag a file, the file is not added to the directory. error-1 
> 
> **Tip**: I think the fix will be to select the correct directory in search_panel.
> 
> 2- It should not list the selected directory in search_panel. **v15**: ejemplo-v15
> 
> **v16**: ejemplo-v16
> 
> TT44199

@victoralmau, thank you very much for your feedback. The issue of showing the correct directory is fixed now. Please take a look and give it a try. I will take time to write the summary of the changes in the next days.

dms/__manifest__.py
vancouver29: If we want to preview more data types, I suggest using the module `attachment_preview` instead. If anyone wants this module in v16, please open an issue. I could migrate it if I have time for that. @pedrobaeza, do you know if anyone can review this PR [attachment_preview PR 361](https://github.com/OCA/knowledge/pull/361) and merge it if everything is ok?

dms/__manifest__.py
vancouver29: Yeah, I think so, or do you have any suggestions?

dms/controllers/main.py
vancouver29: done!

dms/models/dms_file.py
vancouver29: It is a good point. I fixed it. Thanks!

dms/models/dms_file.py
vancouver29: I removed it since it is not necessary anymore!

dms/models/dms_file.py
vancouver29: Is fixed!

dms/__manifest__.py
pedrobaeza: Well, adding a glue module between 2 can be a good option for viewing them, but I would like to go further with a direct edition feature using some of the available server options (LibreOffice online or OnlyOffice integration with Odoo - https://www.onlyoffice.com/es/office-for-odoo.aspx -).

dms/tests/common.py
vancouver29: I agree with @len-foss. With his solution, the tests could go through. Here are the arguments for his solution: 

* in the form view of the directory: you can only set the **model_id**, not the **res_model**, because **res_model** is invisible. We get this error message if we keep the line. 

```bash

  File "/home/vancouver/*/*/*/*dms/tests/test_portal.py", line 13, in setUp
    super().setUp()
  File "/opt/odoo/lib/odoo/odoo/tests/common.py", line 1605, in setUp
    super().setUp()
  File "/home/vancouver/*/*/*/*dms/tests/common.py", line 277, in setUp
    self.create_directory(storage=self.storage, res_model="res.partner")
  File "/home/vancouver/*/*/*/*dms/tests/common.py", line 237, in create_directory
    record.res_model = res_model
    ^^^^^^^^^^^^^^^^
  File "/opt/odoo/lib/odoo/odoo/tests/common.py", line 2254, in __setattr__
    assert not self._get_modifier(field, 'invisible'), \
AssertionError: can't write on invisible field res_model
```

* Either we make **res_model** field editable in form view or keep those changes for the test

dms/views/directory.xml
vancouver29: It does not work if I remove this view. 

* Before removing this view (User as Manager): 

dms_with_view_dms_directory_manager_form_edited

* After removing this view (User as normal User and Manager): 
dms_without_view_dms_directory_manager_form

Do you have any other suggestions here?

len-foss: I think you must have forgotten to duplicate the field in the parent view; one with the `no_open` option and groups `!dms.group_dms_manager`, the other with the `groups dms.group_dms_manager` and without the option. If you do so you should be good.

dms/__manifest__.py
len-foss: Oh that's right I confused that with the `attachment_preview` module. 
It'd be a nice thing to have for sure!

OnlyOffice could be nice I guess, but it seems it would be some order(s) of magnitude larger in effort. And it doesn't invalidate that a direct preview embedded would probably still be useful, unless OnlyOffice could instantly load in an iframe without much disruption of the UX, which doesn't seem likely.

dms/views/directory.xml
vancouver29: `directory.xml`
image

No, I did not forget it.

dms/views/directory.xml
```diff
@@ -0,0 +1,731 @@
...
+                    </div>
+                    <group name="settings">
+                        <group>
+                            <field name="is_root_directory" invisible="1" />
+                            <field
+                                name="storage_id"
+                                options="{'no_open': True}"
+                                attrs="{'readonly':[('is_root_directory','=',False)], 'required':[('is_root_directory','=',True)]}"
+                                force_save="1"
+                            />
```
len-foss:
```suggestion
                            <field
                                name="storage_id"
                                options="{'no_open': True}"
                                attrs="{'readonly':[('is_root_directory','=',False)], 'required':[('is_root_directory','=',True)]}"
                                force_save="1"
                                groups="!dms.group_dms_manager"
                            />
                            <field
                                name="storage_id"
                                options="{}"
                                attrs="{'readonly':[('is_root_directory','=',False)], 'required':[('is_root_directory','=',True)]}"
                                force_save="1"
                                groups="dms.group_dms_manager"
                            />
```

dms/views/directory.xml
len-foss: I added a new code suggestion below. Basically, if you duplicate the fields in the parent view, you don't need the override altogether. But the comments are a bit confused with all that is going on.

My tip with that is generally to address reviewer comments in specific commits, and squash things as necessary at the very last moment before merge, so that it's easy to keep track of everything during back and forth.

If you need I'll do another branch so that you can cherry-pick fixes to that.

dms/views/dms_file.xml
vancouver29: Done!

dms/views/dms_file.xml
vancouver29: Done!

dms/views/directory.xml
vancouver29: Done!

dms/views/directory.xml
vancouver29: Done!

dms/__manifest__.py
```diff
@@ -0,0 +1,59 @@
...
+            "dms/static/src/models/*.js",
+        ],
+        "web.assets_backend": [
+            "dms/static/src/scss/*",
+            "dms/static/src/js/fields/*",
+            "dms/static/src/js/views/*.esm.js",
+            "dms/static/src/js/views/*.xml",
+            "dms/static/src/js/views/fields/binary/*",
+        ],
+        #    "web.assets_qweb": ["dms/static/src/xml/views.xml"],
```
victoralmau:
```suggestion
```
If this is finally not necessary, you can delete it.

dms/models/dms_file.py
```diff
@@ -0,0 +1,643 @@
...
+    #     readonly=True,
+    #     invisible=True,
+    # )
+
+    # # ----------------------------------------------------------
+    # # Compute
+
+    # def _compute_default_model_name(self):
+    #     for record in self:
+    #         record.model_name = "dms.file"
```
victoralmau:
```suggestion
```

dms/views/dms_file.xml
```diff
@@ -0,0 +1,682 @@
...
+                <field name="icon_url" />
+                <field name="create_uid" />
+                <field name="write_date" />
+                <field name="locked_by" />
+                <field name="is_locked" />
+                <field name="is_lock_editor" />
+                <field name="permission_write" />
+                <field name="permission_unlink" />
+                <field name="__last_update" />
+                <!-- <field name="model_name" /> -->
```
victoralmau, line 112:
```suggestion
```

dms/views/dms_file.xml
```diff
@@ -0,0 +1,682 @@
...
+                                            t-att-data-id="record.id"
+                                            t-att-id="record.id.raw_value"
+                                            t-att-name="record.name.raw_value"
+                                        >
+                                            <img
+                                                t-att-src="record.icon_url.raw_value"
+                                                alt="Icon"
+                                            />
+                                        </a>
+
```
victoralmau:
```suggestion
```

dms/views/dms_file.xml
```diff
@@ -0,0 +1,682 @@
...
+        <field name="name">dms_file.form</field>
+        <field name="model">dms.file</field>
+        <field name="inherit_id" ref="view_dms_file_form" />
+        <field name="arch" type="xml">
+            <field name="storage_id" position="attributes">
+                <attribute name="options">{}</attribute>
+                <attribute name="groups">dms.group_dms_manager</attribute>
+            </field>
+        </field>
+    </record> -->
```
victoralmau:
```suggestion
```

victoralmau (changes requested):
Other pending issues:

1- Filter only by folders that have parent_id=False (In v15 it works like this).
dms-1

2- The subfolders would be missing here (In v15 it is also wrong, but it would be interesting to solve it if possible.)
dms-2

3- It is not possible to upload a file from the Files smart-button of a folder (In v15 it works like this)
dms-3

Could you please mark as solved the comments that are done in order to clarify what is pending? Thanks

dms/tests/common.py
```diff
@@ -0,0 +1,311 @@
...
+    def create_directory(self, storage=False, directory=False, model_id=False):
+        record = Form(self.directory_model)
+        record.name = uuid.uuid4().hex
+        record.is_root_directory = True
+        if model_id and storage.save_type == "attachment":
+            # set storage_id_save_type to attachment, making model visible
+            record.storage_id = storage
+            record.model_id = model_id
+        if directory:
+            record.is_root_directory = False
```
victoralmau, line 239:
```suggestion
```

dms/tests/common.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def create_storage(self, save_type="database"):
+        return self.storage_model.create(
+            {"name": "Test Storage", "save_type": save_type}
+        )
+
+    def create_directory(self, storage=False, directory=False, model_id=False):
+        record = Form(self.directory_model)
+        record.name = uuid.uuid4().hex
+        record.is_root_directory = True
```
victoralmau, line 233:
```suggestion
        if not directory:
             record.is_root_directory = True
```

victoralmau (changes requested):
These changes will probably resolve tests.

Another possibility would be to use a default in Form() in some cases, something like:

`record = Form(self.directory_model.with_context(default_is_root_directory=True if not directory else False).`

marylla (commented):
Hi @vancouver29,
I tested the dms today, too.

It looks very good so far but I still get the issues mentioned here https://github.com/OCA/dms/pull/262#pullrequestreview-1642405772 by @victoralmau.

Can you have a look at these?

vancouver29:
> Hi @vancouver29, I tested the dms today, too.
> 
> It looks very good so far but I still get the issues mentioned here [#262 (review)](https://github.com/OCA/dms/pull/262#pullrequestreview-1642405772) by @victoralmau.
> 
> Can you have a look at these?

Hi @marylla @victoralmau, Thanks for the feedback. I will adjust the new patch for that issue. A question is still open here. This work is financially supported by a sponsor, @agent-z28, and his company.  Is it legit to put his company as a sponsor of this module and me and another developer as contributors for this module? @pedrobaeza, could you please give us a hint here?

pedrobaeza:
On the file CREDITS.rst, put that the migration to 16.0 has been funded by agent-z18, and on contributors, put your names.

vancouver29:
> Other pending issues:
> 

 ->  1- **(resolved)**  
> 1- Filter only by folders that have parent_id=False (In v15 it works like this).  dms-1
> 

 ->  2- **(unclear)** :  this is unclear to me. Could you please describe the issue in more detail? Or a short video would be helpful. 
> 2- The subfolders would be missing here (In v15 it is also wrong, but it would be interesting to solve it if possible.) dms-2

->   3-  **(resolved)**
> 3- It is not possible to upload a file from the Files smart-button of a folder (In v15 it works like this) dms-3
> 
> Could you please mark as solved the comments that are done in order to clarify what is pending? Thanks

dms/models/dms_file.py
vancouver29: **(resolved)**

dms/__manifest__.py
vancouver29: **(resolved)**

dms/views/dms_file.xml
vancouver29: **(resolved)**

dms/views/dms_file.xml
vancouver29: **(resolved)**

dms/views/dms_file.xml
vancouver29: **(resolved)**

victoralmau:
Thank you very much for the changes, only point 2 is missing.

Explanation point 2:
- Go to DMS > Directories
- Go to the Documents folder
- Click on the Files smart-button
- In the Directories side Searchpanel "Documents" is not auto-selected and the subdirectories are not shown there.

ejemplo-ko-1
ejemplo-ko-2
ejemplo-ko-3

**Tip**: If you click on the "Files" icon of the folder in the kanban view, it works correctly.

ejemplo-ok-1
ejemplo-ok-2

vancouver29:
@victoralmau, thank you so much for the detailed explanation. Point 2 is now resolved. Please test it again. :) @marylla, @agent-z28, @pedrobaeza   FYI.

@len-foss, you fixed the uni tests of this module. Should I put you as the co-contributor for the migration of this module?

victoralmau (approved):
Great job, thank you!

dms/models/dms_file.py
```diff
@@ -601,3 +601,29 @@ def _compute_locked(self):
                 )
             else:
                 record.update({"is_locked": False, "is_lock_editor": False})
+
+    def create_dms_file_from_attachments(self, attachment_ids=None, view_type="list"):
```
pedrobaeza: Why this is called `create_...` if it's not creating anything? What is its purpose? And please don't put empty lines inside a method, that is less readable. If you need to put a "section", just add a comment line and you can give such section "a title".

dms/views/dms_file.xml
pedrobaeza: This commented code is still there.

pedrobaeza (commented):
Are video and audio correctly previewed without the extra dependencies?

dms/models/dms_file.py
```diff
@@ -0,0 +1,629 @@
...
+            for attachment in attachments
+        ):
+            raise UserError(_("Invalid attachments!"))
+
+        for attachment in attachments:
+            return {
+                "name": attachment.name,
+                "datas": attachment.datas,
+                "res_model": attachment.res_model,
+            }
```
len-foss:
```suggestion
        return [{
                "name": attachment.name,
                "datas": attachment.datas,
                "res_model": attachment.res_model,
            } for attachment in attachments]
```

It is a bit weird that it would raise and not simply return an empty list if it's supposed to return a list.

Otherwise, replace the `attachment_ids` argument by `attachment_id` and simply return a dict.

In any case, this implementation always return only one attachment, which is quite wrong and should have been caught by some unit test. (let me know if you need some help).

pedrobaeza:
Please add #264

dms/models/dms_file.py
vancouver29: Yeah. That makes sense. I changed it to `get`

dms/models/dms_file.py
vancouver29: thanks! It is fixed now. The user can upload more than one attachment. If you have time, you can look at the error of the pre-commit :)

vancouver29:
> Please add #264

Did you test it in `dms` v16? 

I tried to reproduce the problem, which is described here https://github.com/OCA/dms/issues/263 with the user `Mark Demo`. But No Error has occurred. So I think we don't need this.

dms/models/dms_file.py
vancouver29: Video and Audio work now :) and now DMS even supports more previewable audio types than Odoo core (mp3)

dms/models/dms_file.py
len-foss: Nice. You can cherry-pick the commit https://github.com/len-foss/dms/commit/dd9994297fc42ef3bb76a26c0dee1bff555515fb and squash it. Note that I rebased on `origin/16.0` before (you get the pre-commit error locally if you rebase).

dms/models/dms_file.py
```diff
@@ -0,0 +1,631 @@
...
+        """Get the dms files from uploaded attachments.
+        :return: An Array of dms files.
+        """
+
+        if attachment_ids is None:
+            attachment_ids = []
+        attachments = self.env["ir.attachment"].browse(attachment_ids)
+
+        if not attachments:
+            raise UserError(_("No attachment was provided"))
```
len-foss:
```suggestion
    def get_dms_files_from_attachments(self, attachment_ids=None):
        """Get the dms files from uploaded attachments.
        :return: An Array of dms files.
        """
        if not attachments:
            raise UserError(_("No attachment was provided"))

        attachments = self.env["ir.attachment"].browse(attachment_ids)
```

A little nitpick that simplifies things a bit; if no attachments are given it will raise anyway. Also the kwarg is somewhat misleading so better to drop it.

len-foss:
```suggestion
    def get_dms_files_from_attachments(self, attachment_ids=None):
        """Get the dms files from uploaded attachments.
        :return: An Array of dms files.
        """
        if not attachments:
            raise UserError(_("No attachment was provided"))

        attachments = self.env["ir.attachment"].browse(attachment_ids)
```

A little nitpick that simplifies things a bit; if no attachments are given it will raise anyway. Also the kwarg is somewhat misleading so better to drop it.

vancouver29: great and thanks! Your suggestion has been adjusted

len-foss:
With regards to the PR mentioned, the fix is in `dms_field`, you cannot apply half of it, and the half you can apply doesn't make sense on its own. Given that the `compute` field on `model` could be added in an override in `dms_field`, it's technically not really right to put that code in `dms`.

pedrobaeza:
Well, I disagree here, as any other chance to have a non sudo access will provoke the same problem, so it's better to fix it in the root, as it also doesn't cause any cons.

len-foss:
Well but the problem is that there is no path for reproduction, as @vancouver29 noted, and no test that can be written in `dms`.
As it stands it's the code from `dms_field` that emits the domain triggering the error.
So if you look at the code, it isn't used anywhere, and the commit message says "Improve search by model". Without going back to the Github PR, there is nothing that indicates why this code should exist.

There should be a comment explaining why it is useful, or at the very least this information should be in the commit message. Something like "while this field is not used, depending modules might need to search on `model`; however the document manager might not have read access to ir.model".

pedrobaeza:
OK, agreed of explaining everything better. The fix and its explanation can be added later in a subsequent PR.

agent-z28 (approved):
For me it looks very good now, i tested it and seems to be a nice module. Thanks for your effort :)

marylla:
@pedrobaeza Can we merge this DMS PR?

vancouver29:
> On the file CREDITS.rst, put that the migration to 16.0 has been funded by agent-z18, and on contributors, put your names.

@pedrobaeza thank you very much for merging this work. But I don't see the changes in `README.rst` anymore (we are contributors and @agent-z28 is the sponsor for this migration from v15 to v16). Is something here missing?  @marylla FYI

pedrobaeza:
You didn't modify `readme/CREDITS.rst` nor `readme/CONTRIBUTORS.rst`, but `README.rst`, so that's why your changes are lost. Read again my comment, in which I mention the file. Do another PR introducing that changes.

len-foss:
@vancouver29 You can check that your changes would be properly applied by running the maintainer tools manually. Just install it with pipx and run `oca-gen-addon-readme --repo-name=dms --branch=16.0 --addon-dir=dms`, this way you won't have any surprise once your PR is merged.

https://github.com/OCA/maintainer-tools#readme-generator

pedrobaeza:
Now only having pre-commit, the readme regeneration is done.

vancouver29:
> You didn't modify `readme/CREDITS.rst` nor `readme/CONTRIBUTORS.rst`, but `README.rst`, so that's why your changes are lost. Read again my comment, in which I mention the file. Do another PR introducing that changes.

@pedrobaeza Sorry, my fault! Thanks for the hint again. I just created a new PR here https://github.com/OCA/dms/pull/268

>@vancouver29 You can check that your changes would be properly applied by running the maintainer tools manually. Just install it with pipx and run oca-gen-addon-readme --repo-name=dms --branch=16.0 --addon-dir=dms, this way you won't have any surprise once your PR is merged.
>
>https://github.com/OCA/maintainer-tools#readme-generator

@len-foss, thank you very much for the hint. I will notice it next time :)

PR #267, opened 2023-11-04
[14.0][IMP] DMS: download button in tree and kanban views

FrancescoBallerini:
Really simple PR: add download button in all views so we can avoid to open record formview just to download file which is annoying in general, expecially when we have to download more than one file.

What do you think?

### DIFF
--- a/dms/views/dms_file.xml
+++ b/dms/views/dms_file.xml
@@ -315,6 +315,12 @@
                 <field name="is_locked" invisible="1" />
                 <field name="is_lock_editor" invisible="1" />
                 <field name="name" />
+                <field
+                    name="content"
+                    filename="name"
+                    widget="binary"
+                    string="Download"
+                />
                 <field name="write_date" />
                 <field name="size" widget="integer" />
                 <field name="mimetype" />


dms/views/dms_file.xml
```diff
@@ -240,7 +240,11 @@
                                         <div
                                             class="o_kanban_record_title o_text_overflow"
                                         >
-                                            <field name="name" />
+                                            <field
+                                                name="content"
+                                                filename="name"
+                                                widget="binary"
+                                            />
```
victoralmau: I think it is not necessary because the action of download already exists.
ejemplo

dms/views/dms_file.xml
```diff
@@ -315,6 +319,12 @@
                 <field name="is_locked" invisible="1" />
                 <field name="is_lock_editor" invisible="1" />
                 <field name="name" />
+                <field
+                    name="content"
+                    filename="name"
+                    widget="binary"
+                    string="Download"
+                />
```
victoralmau, line 323: I think it would be better to add a button with an action to make the download.

victoralmau (changes requested):

dms/views/dms_file.xml
FrancescoBallerini: I will be honest: I did not see that download button, it's still necessary to make one more click but at least you don't have to open/close the form-view everytime.. so yes, it's indeed less necessary than I thought 👍

dms/views/dms_file.xml
FrancescoBallerini: Thanks for your feedbacks!

I could change it to a url action and call something like:

`/web/content?id=14&amp;field=content&amp;model=dms.file&amp;filename_field=name&amp;download=true`

but I have to change it a bit to make it working with URL action, do you think this is right or you suggest a better approach? 

In URL action case I'm pretty sure that I can access record id by context, but I am not sure how pass the other dynamic value, which is "name" for "filename_field", have a suggestion for that?

FrancescoBallerini: @victoralmau Removed `content` field in the kanban, waiting for feedback about this. 

Meanwhile I have made some experiments but I wasn't able to call the download from an URL action defined by xml because I didn't managed reach id and record name dynamically. 

I could add a type object button and return an url action by python, but (just my opinion) it would require much more code to obtain basically the same behaviour than I can obtain with the binary widget. 

Unless you mean to add a contextual action, I can see advantages on that (multiple downloads at same time) but it wasn't the scope of this PR.

What do you think?

victoralmau: What do you think @pedrobaeza ?

pedrobaeza: Not sure what do you want. Reading vertically, it seems you want a direct button to download without doing 2 clicks, so that's something reasonable, but how it's implemented is not so important in this case IMO. It should just work.

victoralmau (approved):

pedrobaeza:
Hey, wait, adding a binary widget means that this view downloads whole files content without requesting the download, which is not sustainable. Am I wrong?

dms/views/dms_file.xml
FrancescoBallerini: @pedrobaeza Yes I basically just wanted to skip a couple of click for the file download, that is correct.

pedrobaeza: Please see my concern on the bottom.

FrancescoBallerini:
> Hey, wait, adding a binary widget means that this view downloads whole files content without requesting the download, which is not sustainable. Am I wrong?

Mmm I'm not sure how the widget works, specifically, but if user has access to the button then yes, it will be able to download it without additional checks, if that's what you mean by "requesting the download". 

I supposed that user should automatically have access to the button if he can see folder content, but now that you let me think about I might have underestimated that part.

pedrobaeza:
I mean that loading the kanban view means to "download" all the files. Please check on your network tab in developer tools if that's happening.

FrancescoBallerini:
> I mean that loading the kanban view means to "download" all the files. Please check on your network tab in developer tools if that's happening.

I have checked the network headers and console: by using the binary widget the content is never downloaded when opening view. It is downloaded only when clicking on the field associated with the widget (which is `name` in this case). 

In the case of a kanban this might lead the user to accidentally clicking on the filename instead of the kanban body, and download the file instead of opening the form-view. But we have removed the additional download button from the kanban-view at the moment, the PR is only adding one in the the tree-view

image

Aniway, in both cases kanban/tree-view, the "content" header is never downloaded when loading the view. File will only be downloaded on request, by clicking on the widget (which correspond to "download" column field in the case of the tree-view in the screenshot).

pedrobaeza (approved):
OK, seen.

/ocabot merge minor

PR #272, opened 2023-11-16
[FIX] dms: remove padding only for directory kanban records

### LINKED ISSUE #271: [16.0] dms 16.0 is removing kanban record padding in all kanban views
## Module
dms 16.0

## Describe the bug
To implement dms directory kanban view, the file dms/static/src/scss/directory_kanban.scss:92 sets padding as 0 for .o_kanban_record > div
Screenshot 2023-11-16 at 12 33 59 PM

As a result all padding is removed from all kanban records as well.
For example CRM kanban view.

Screenshot 2023-11-16 at 12 18 12 PM

It affects all other kanban views as well (for eg: contacts) in the same manner.
 
## To Reproduce
**Affected versions**: 16.0

1. Install dms 16.0 on any odoo 16.0 instance

**Expected behavior**
Installing dms module should not affect other kanban views

**Additional context**
Thanks a lot for the dms module and the much awaited migration to 16.0

lijoantony:
16.0 migration added css to remove padding from dms directory kanban records. Inadvertently, it removed padding from *all* kanban view records in odoo.

This commit fixes it by removing the problematic css. Instead, padding is removed only from directory kanban view records, in a manner similar to what file_kanban.scss does for file kanban view.

Fixes #271

### DIFF
--- a/dms/static/src/scss/directory_kanban.scss
+++ b/dms/static/src/scss/directory_kanban.scss
@@ -8,6 +8,10 @@
 .mk_directory_kanban_view {
     height: 100%;
     .o_kanban_record {
+        > div:nth-child(1) {
+            padding: 0;
+        }
+
         padding: 0;
         max-height: $o-kanban-image-width + 1;
         .mk_directory_kanban_actions {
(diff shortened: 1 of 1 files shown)


len-foss (approved):

pedrobaeza (approved):
Thanks for the patch.

As reference, there's no need to create an issue if you are going to make a PR. Put the issue content in the PR description.

/ocabot merge patch

PR #274, opened 2023-12-13
[15.0][IMP] dms: Show all directories in searchpanel from files

victoralmau:
Show all directories in searchpanel from files

In v13 the `search_panel_select_range()` method (https://github.com/odoo/odoo/blob/13.0/addons/web/models/models.py#L240) returned all directories, now only records that have any files in any directory are returned (this causes that directories without files are not shown).

**v13**:
13 0-dms

**Before**:
15 0-dms-antes

**After**:
15 0-dms-despues

@Tecnativa TT46388

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -295,22 +295,36 @@ class File(models.Model):
 
     @api.model
     def search_panel_select_range(self, field_name, **kwargs):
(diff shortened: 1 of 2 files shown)


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

I think both views should be unified in one in later versions.

PR #275, opened 2023-12-13
[16.0][ADD] dms_storage

len-foss:
Migration of https://github.com/OCA/dms/pull/182 depending on `fs_storage` instead of `storage_backend`.

(The same TODO applies, which is that changing the storage of existing directories is not supported yet)


len-foss:
Tests should work once https://github.com/OCA/storage/pull/314 (or maybe another one if it ends being split) gets merged.

pedrobaeza:
Please rebase

len-foss:
Done :-)

pedrobaeza (approved):
Let's merge this useful one:

/ocabot merge nobump

PR #276, opened 2023-12-14
[15.0][FIX] dms: Set the correct parent_id of directories in searchpanel

victoralmau:
Related to https://github.com/OCA/dms/pull/274

If the parent directory is not in all the records we should not set `parent_id` because the user does not have access to parent.

@Tecnativa TT46388


pedrobaeza (approved):
Fast-tracking for solving the problem:

/ocabot merge patch

PR #277, opened 2023-12-18
[WIP][16.0][FIX] dms: fix directory actions

len-foss:
If the directory is empty, the action would return all files, and the panel show all directories except for the one that triggered the action.

The file count was not consistent with the action if the directory is hidden, since its files would thus also be hidden; this field is a related on the storage, so the value is consistent for the directory and all its files.

### DIFF
--- a/dms/static/src/js/views/search_panel.esm.js
+++ b/dms/static/src/js/views/search_panel.esm.js
@@ -22,6 +22,7 @@ patch(SearchModel.prototype, "dms.SearchPanel", {
                     domain.push([category.fieldName, "=", category.activeValueId]);
                 }
                 if (domain.length === 0) {
+                    // / only correct for the main action!
                     domain.push([category.fieldName, "=", false]);
                 }
             }


dms/models/directory.py
```diff
@@ -446,12 +446,33 @@ def _compute_count_total_directories(self):
...
+            },
+        }
+
     def _compute_count_total_files(self):
         model = self.env["dms.file"]
         for record in self:
             # Prevent error in some NewId cases
             record.count_total_files = (
-                model.search_count([("directory_id", "child_of", record.id)])
+                model.search_count(self._get_directory_files_domain())
```
victoralmau:
```suggestion
                model.search_count(record._get_directory_files_domain())
```

victoralmau (changes requested):

dms/models/directory.py
len-foss: Thank you for catching this!

dms/models/directory.py
```diff
@@ -446,12 +446,33 @@ def _compute_count_total_directories(self):
...
+            ("is_hidden", "=", self.is_hidden),
+        ]
+
+    def action_dms_files_all_directory(self):
+        self.ensure_one()
+        return {
+            "type": "ir.actions.act_window",
+            "name": _("Files"),
+            "res_model": "dms.file",
+            "view_mode": "tree,form",
```
victoralmau:
```suggestion
            "view_mode": "kanban,tree,graph,pivot,form",
```

victoralmau (changes requested):
Can you do something similar with the "_Subdirectories_" smart-button of the directory form view? That way, we will get the searchpanel to have set the folder we have clicked on so we can easily create a directory.

len-foss:
@victoralmau 

>Can you do something similar with the "Subdirectories" smart-button of the directory form view? That way, we will get the searchpanel to have set the folder we have clicked on so we can easily create a directory.

So I checked the 4 cases, over (directory/files) and (child_of/direct).
It turns out only the case directory/direct works fully, which means that the correct domain is displayed, and the search panel is correctly set. In the directory/child_of case, the search panel is wrong.

Currently in both file cases the domain is wrong if there are no files. In the file/direct case, the domain is always wrong, but the search panel is correct.

With my proposed change, the domain is correct in all cases but the search panel is not. So it's slightly better but not a true fix. I'll try to look more in-depth at what happens to see if there's a true solution.

len-foss:
The previous approach was flawed I think, as the problem comes from the JS function `_getCategoryDomain`.

If you look at actions like `action_dms_files_directory`, the `("directory_id", "child_of", active_id)` has been removed from the action domain in comparison with version `15.0`. Why? Because if this domain is there, the search panel filters too much information to be able to display the proper hierarchy. 
However this means it's `_getCategoryDomain` that automatically fills in the domain. 
For files the issue is that it chooses the `child_of` operator even if the action is intended to be an `=`.
The current override is basically a hack to filter only top-level directories when coming from the main action. 
However this hack is not able to distinguish when we come from other actions and that's what is messing the "subdirectories" action for instance.

So given the current code, `_getCategoryDomain` should get the current action as argument to decide how to modify the domain, but that's not the case. So, I think the override should be somewhere else where the info is available and can be passed down. I'm not familiar enough with the JS framework to know where, though.

@vancouver29 if you have any idea, let me know!

victoralmau:
Related to https://github.com/OCA/dms/pull/301

PR #279, opened 2024-01-04
[15.0][ADD] hr_dms_field: New module

victoralmau:
Changes done:
- [x] `dms`:  Add ACLs to `dms.storage` for users with the basic permission.
- [x] `dms_field`: Create dms.field.template to set model, groups and directory structure.
- [x] `dms_field`: Remove the wizard to create a directory from a record.
- [x] `dms_field`: Directory structure is created when record is created (partner for example).
- [x] `dms_field`:  Change the information on the right side.
- [x] `dms_field`: Have directory/file tree open when loading

**Directory**
directory

**File** (preview when clicking on image/icon)
file-1

Please  @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT46804


victoralmau:
Ping @pedrobaeza and @CarlosRoca13

CarlosRoca13 (approved):
LGTM :+1:

dms_field/models/dms_file_template.py
```diff
@@ -0,0 +1,148 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class DmsFileTemplate(models.Model):
+    _name = "dms.file.template"
```
pedrobaeza: Why `dms.file.template` instead of `dms.field.template`? The other name is misleading.

dms_field/views/dms_file_template_views.xml
```diff
@@ -0,0 +1,65 @@
...
+                        >
+                            <field name="dms_directory_ids" mode="dms_tree" />
+                        </page>
+                    </notebook>
+                </sheet>
+            </form>
+        </field>
+    </record>
+    <record id="action_dms_file_template" model="ir.actions.act_window">
+        <field name="name">File Templates</field>
```
pedrobaeza: Let's call it "Embedded DMS templates"

pedrobaeza (commented):
In general it works OK, but it's ugly that each time you add a folder or a file into the embedded widget, the reload unfolds all the hierarchy, having to manually unfold again. I think the default should be unfolded by default, and if there's a possibility of keeping in the JS layer the folding state. But at least the first option should be implemented.

pedrobaeza:
It still keeps folded by default.

victoralmau:
> It still keeps folded by default.

Done.

dms/models/directory.py
```diff
@@ -479,6 +480,11 @@ def _compute_size(self):
             )
             record.size = sum(rec.get("size", 0) for rec in recs)
 
+    @api.depends("size")
+    def _compute_human_size(self):
+        for item in self:
+            item.human_size = human_size(item.size) if item.size else ""
```
pedrobaeza: Never assign `""`. Put False.

dms/security/ir.model.access.csv
```diff
@@ -3,6 +3,7 @@ id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
 access_dms_tag_user,dms_tag_user,model_dms_tag,group_dms_user,1,1,1,1
 access_dms_category_user,dms_category_user,model_dms_category,group_dms_user,1,1,1,1
 
+access_dms_storage_base_user,dms_storage_base_user,model_dms_storage,base.group_user,1,0,0,0
```
pedrobaeza, line 6: Why adding this open permission? This seems too much.

pedrobaeza (commented):
Second level folders are still folded by default.

dms/security/ir.model.access.csv
victoralmau: This change is absolutely necessary IMO.

`dms` adds read permissions for `base.group_user` to all models except `dms.storage` (`base.group_portal` for example has read permission).

The `dms_field` module and `mode="dms_tree"` widget get: storage, directories and files. A basic user without permissions in _Documents: User_ (for example an employee when accessing employee form view), not having access to `dms.storage` would have a permissions error.

victoralmau:
> Second level folders are still folded by default.

Done.

pedrobaeza (approved):
/ocabot merge minor

PR #280, opened 2024-01-05
[16.0][IMP] dms: Show all directories in searchpanel from files

victoralmau:
FWP from 15.0: https://github.com/OCA/dms/pull/274 + https://github.com/OCA/dms/pull/276 + https://github.com/OCA/dms/pull/278

Show all directories in searchpanel from files

In v13 the `search_panel_select_range()` method (https://github.com/odoo/odoo/blob/13.0/addons/web/models/models.py#L240) returned all directories, now only records that have any files in any directory are returned (this causes that directories without files are not shown).

**v13**:
13 0-dms

**Before**:
15 0-dms-antes

**After**:
15 0-dms-despues

@Tecnativa TT46388


chienandalu (commented):
Another difference I saw it's that when you're in a folder, you get all the files from the subfolders, which wasn't the original behavior, where just got the files in that folder.

Not which one I prefer though :smile:

pedrobaeza (approved):
Apart from David's comment.

victoralmau:
> Another difference I saw it's that when you're in a folder, you get all the files from the subfolders, which wasn't the original behavior, where just got the files in that folder.
> 
> Not which one I prefer though 😄

Oops, thanks, that's solved (I thought this was the same as in v15).

chienandalu (approved):
Great! :+1:

pedrobaeza:
/ocabot merge patch

Please check if all other changes of 15.0 are reflected in this version.

PR #284, opened 2024-01-18
[13.0][FIX] dms_field: Opening a directory using the right click open action

victoralmau:
FWP from 12.0: https://github.com/OCA/dms/pull/245

Steps to reproduce:

    Create a new root directory
    Right-click and choose `Open`

Current behavior:

Expected behavior:

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #287, opened 2024-01-23
[15.0][ADD] dms_auto_classification: New module

victoralmau:
New module

ejemplo

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT46805

PR #288, opened 2024-01-23
[15.0][ADD] dms_field_auto_classification: New module

victoralmau:
New module

Locked by:
- [x] `dms_field`: https://github.com/OCA/dms/pull/279
- [x] `dms_auto_classification`: https://github.com/OCA/dms/pull/287

@Tecnativa TT46805


victoralmau:
Ready to review.

dms_field_auto_classification/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Auto classification files to dms_field",
```
pedrobaeza:
```suggestion
    "name": "Auto classify files into embedded DMS",
```

dms_field_auto_classification/models/dms_classification_template.py
```diff
@@ -0,0 +1,44 @@
...
+    detail_ids = fields.One2many(
+        string="Details",
+        comodel_name="dms.classification.template.detail",
+        inverse_name="parent_id",
+    )
+
+    @api.depends("model_id")
+    def _compute_model(self):
+        for item in self:
+            item.model = item.model_id.model
```
pedrobaeza, line 23: I think this will fail with non admin permissions, as models are restricted (and the same for any model access in whole module).

dms_field_auto_classification/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Automatically classify files within a .zip file to the corresponding DMS directory(s) related to a record.
```
pedrobaeza:
```suggestion
Automatically classify files within a .zip file to the corresponding directory(s)
related to an embedded DMS.
```

dms_field_auto_classification/static/description/icon.png
pedrobaeza, line 1: You can use the same icon as the `dms_auto_classification`.

dms_field_auto_classification/models/dms_classification_template.py
victoralmau: This field is compute sudo.
Template must be created by a user with `ir.model` permissions.

I have added to tests a user with permissions in Documents: Manager to confirm that that user can do the whole process (without having permissions to `ir.model`).

pedrobaeza: > Template must be created by a user with `ir.model` permissions.

Only maximum admins have this permission, so this should be avoided (`mis_builder` for example had similar problems).

victoralmau: I understand the problem of high permissions, but what suggestion do you propose? 
Another option is not to select a model (`ir.model`) and write the name (res.partner for example) directly.

pedrobaeza (approved):

chienandalu (commented):
Maybe I'm doing something wrong, but I couldn't make it work in demo:

- I'm using your test zip
- I set the VAT numbers to two customers.
- I try to classify them

But no classification is make or folder linked to the partners...

victoralmau:
> Maybe I'm doing something wrong, but I couldn't make it work in demo:
> 
>     * I'm using your test zip
> 
>     * I set the VAT numbers to two customers.
> 
>     * I try to classify them
> 
> 
> But no classification is make or folder linked to the partners...

Sorry, you need to create an Embedded DMS templates for partners (https://github.com/OCA/dms/pull/288/files#diff-d37451b07c0c0fd8b55c34a39c8ac57c604748a580057a00fc24ccdbb3525ddeR23) and then (with the corresponding button) in the form view of each partner you must create the linked folder (this is how `dms_field` works).

chienandalu (approved):

Issue #189, opened 2022-07-04
14c - DMS uninstall/install error


kimasplund:
So as to my previous error reported here i tried to uninstall and reinstall. 
Uninstall went ok but then installing it back produces the following error:
```
Validation Error
A directory has to have a parent directory.
```
And then installation fails.

log
2022-07-04 09:01:50,264 311315 INFO xxx.fi odoo.addons.mail.models.mail_alias_mixin: Mail alias created for dms.directory Coolsculpting (id 2) 
2022-07-04 09:01:50,267 311315 INFO xxx.fi odoo.addons.mail.models.mail_alias_mixin: Mail alias created for dms.directory Fillers & Botox (id 6) 
2022-07-04 09:01:50,269 311315 INFO xxx.fi odoo.addons.mail.models.mail_alias_mixin: Mail alias created for dms.directory IPL (id 4) 
2022-07-04 09:01:50,271 311315 INFO xxx.fi odoo.addons.mail.models.mail_alias_mixin: Mail alias created for dms.directory Kundformulär (id 1) 
2022-07-04 09:01:50,273 311315 INFO xxx.fi odoo.addons.mail.models.mail_alias_mixin: Mail alias created for dms.directory Kundformulär / Needlepen (id 9) 
2022-07-04 09:01:50,274 311315 INFO xxx.fi odoo.addons.mail.models.mail_alias_mixin: Mail alias created for dms.directory Kundformulär / Vi Peel (id 8) 
2022-07-04 09:01:50,276 311315 INFO xxx.fi odoo.addons.mail.models.mail_alias_mixin: Mail alias created for dms.directory Tandblekning (id 7) 
2022-07-04 09:01:50,279 311315 ERROR xxx.fi odoo.schema: Table 'dms_directory': unable to set NOT NULL on column 'name' 
2022-07-04 09:01:50,280 311315 INFO xxx.fi odoo.models: Storing computed values of dms.directory.storage_id 
2022-07-04 09:01:50,281 311315 INFO xxx.fi odoo.models: Storing computed values of dms.directory.parent_id 
2022-07-04 09:01:50,281 311315 INFO xxx.fi odoo.models: Storing computed values of dms.directory.tag_ids 
2022-07-04 09:01:50,281 311315 INFO xxx.fi odoo.models: Storing computed values of dms.directory.model_id 
[...]

PR #289, opened 2024-01-25
[15.0][FIX] dms: Improve the unlink method in base to avoid errors

victoralmau:
Related to: https://github.com/OCA/dms/issues/189

In the uninstall process of dms, checks were made that prevented the removal of fields and models being deleted, also causing the dms_file and dms_directory tables to continue to exist.

Steps to reproduce the error before:
- Install dms
- Uninstall dms
- The dms.file and dms.directory models still exist.
- The tables dms_file and dms_directory in the database still exist.

@Tecnativa

### DIFF
--- a/dms/models/base.py
+++ b/dms/models/base.py
@@ -1,4 +1,5 @@
 # Copyright 2021 Tecnativa - Jairo Llopis
+# Copyright 2024 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
 from odoo import models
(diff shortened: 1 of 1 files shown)


chienandalu (approved):

pedrobaeza:
I don't get the relation between the problem and this patch.

victoralmau:
With this change when the module is uninstalled and `ir.*` model records are deleted (`ir.model.fields` for example) this code will not be executed, solving the problem.

Before this change and with the steps to reproduce the problem indicated, you can verify that the `dms_file` and `dms_directory` tables were not deleted and the `dms.file` and `dms.directory` models exist.

pedrobaeza (approved):
/ocabot merge patch

PR #292, opened 2024-02-08
[14.0][FIX] dms

adasatorres:
Error when adding file using the  file_ids field in directories view form.

The current_dir was an object of type NewID and there was a problem serializing it with the json.dumps method.

PR #294, opened 2024-02-08
[15.0][FIX]dms: Avoid error when set files from directory form view

victoralmau:
FWP from 14.0: https://github.com/OCA/dms/pull/292

Avoid error when set files from directory form view.

Error when adding file using the file_ids field in directories view form.

The current_dir was an object of type NewID and there was a problem serializing it with the json.dumps method.

@Tecnativa


pedrobaeza (approved):
Let's fast-track it:

/ocabot merge patch

PR #297, opened 2024-02-09
[15.0][IMP] dms_auto_classification: Allow users with Documents: User permissions can use the wizard

victoralmau:
Allow users with Documents: User permissions can use the wizard.

@Tecnativa TT46805


pedrobaeza (commented):
Doesn't the act_window need to change the group?

victoralmau:
> Doesn't the act_window need to change the group?

The act_window does not currently have any groups set https://github.com/OCA/dms/blob/15.0/dms_auto_classification/wizards/wizard_dms_classification_views.xml#L58

pedrobaeza:
I think the menu and act_window should have them.

victoralmau:
Changes done.

pedrobaeza (approved):

victoralmau:
Ping @chienandalu

chienandalu (approved):

PR #302, opened 2024-02-16
[16.0][IMP] dms: Improve tests

victoralmau:
Changes done:
- Change `setup()` to `setUpClass()`
- Add BaseCommon
- Remove `multi_users` decorator (not used)
- Remove `_setup_test_data` and `_load` from common (not used)
- Change DirectoryMailTestCase (extends from StorageDatabaseBaseCase)
- Add `@mute_logger` to skip unnecessary unlink logs

@Tecnativa

### DIFF
--- a/dms/tests/test_benchmark.py
+++ b/dms/tests/test_benchmark.py
@@ -20,7 +20,7 @@ _logger = logging.getLogger(__name__)
 class BenchmarkTestCase(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
-        super(BenchmarkTestCase, cls).setUpClass()
+        super().setUpClass()
         cls._clean_existing_records()
         cls._setup_benchmark_data()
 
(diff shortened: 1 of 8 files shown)


dms/tests/test_benchmark.py
```diff
@@ -2,25 +2,81 @@
...
-from .common import track_function
-
 _logger = logging.getLogger(__name__)
 
+# ----------------------------------------------------------
+# Decorators
+# ----------------------------------------------------------
+
+
+def track_function(
```
pedrobaeza: Why moving this function?

pedrobaeza (commented):
`multi_users` is not used anywhere?

dms/tests/test_benchmark.py
victoralmau: Because it is only used in this test.

victoralmau:
> `multi_users`

No, `@users('user-1', 'user-2')` is used instead.

dms/tests/test_benchmark.py
pedrobaeza: Well, in this case, for reducing the diff, but also if in any moment it's needed by other test, let's keep it where it was. It's not an overhead having it there. It's loaded once the same being in one specific test or in the common one.

pedrobaeza (approved):

chienandalu (approved):
:+1:

PR #305, opened 2024-02-26
[16.0][FIX] dms: Improve the unlink method in base to avoid errors

victoralmau:
FWP from 15.0: https://github.com/OCA/dms/pull/289

Related to: https://github.com/OCA/dms/issues/189

In the uninstall process of dms, checks were made that prevented the removal of fields and models being deleted, also causing the dms_file and dms_directory tables to continue to exist.

Steps to reproduce the error before:
- Install dms
- Uninstall dms
- The dms.file and dms.directory models still exist.
- The tables dms_file and dms_directory in the database still exist.

@Tecnativa


pedrobaeza (approved):
Fast-tracking the fw-port:

/ocabot merge patch

PR #307, opened 2024-03-05
[15.0][IMP] dms: Add move (change directory) action from files

victoralmau:
Add move (change directory) action from files

Changes done:
- [x] Set tracking=True to directory_id field from files to leave log if "_moved_" to another directory.
- [x] Add `root_directory_id` field from files.
- [x] Add optional extra fields to `dms.file` tree view (`directory_id` and `root_directory_id`).
- [x] Add wizard to "move" (change directory) files.

ejemplo

@Tecnativa TT48180


victoralmau:
Ping @chienandalu

pedrobaeza (commented):
Shouldn't this allow to move directories as well?

victoralmau:
> Shouldn't this allow to move directories as well?

I don't think so for different reasons:

- The purpose of this improvement is intended for files only.
- If "_move_" directories means changing their parent directory, that can already be done editing the corresponding directory and changing the "parent directory", that will "move" everything.
- If "_move_" directories is "_move_" all the files it contains, for that there is already the existing option.

chienandalu (approved):
Tested :+1:

pedrobaeza (approved):
/ocabot merge minor

PR #309, opened 2024-03-07
[15.0][FIX] dms: Display the correct values in directory form view

victoralmau:
The total number of files and directories displayed in the smart-buttons of the directory form is currently confusing (searchpanel is already filtered). The total files information is also added (in any subfolder) and the same with directories.

ejemplo

In this way, the data displayed will be the same as in the directory tree view.

@Tecnativa TT48244

### DIFF
--- a/dms/views/directory.xml
+++ b/dms/views/directory.xml
@@ -410,7 +410,7 @@
                         >
                             <field
                                 string="Subdirectories"
-                                name="count_total_directories"
+                                name="count_directories"
                                 widget="statinfo"
                             />
                         </button>
@@ -422,7 +422,7 @@
                         >
                             <field
                                 string="Files"
-                                name="count_total_files"
+                                name="count_files"
                                 widget="statinfo"
                             />
                         </button>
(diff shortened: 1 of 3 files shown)


chienandalu (approved):

dms/views/directory.xml
```diff
@@ -422,7 +422,7 @@
                         >
                             <field
                                 string="Files"
-                                name="count_total_files"
+                                name="count_files"
```
pedrobaeza, line 425: If I'm right, this change will see for example 25 files, although clicking on it will show only the files in the same folder, which can be less. Is it this way? If so, I think it's also confusing, and the change should be to improve the labels on the other stats.

victoralmau: It is just the opposite (now it will be consistent).

**Before**:
antes

**After**:
despues

pedrobaeza:
Well, I see more coherent as it was. If you see the tree in the after GIF, it shows the number 23 on right , so that should be the number to be shown as well in the smart-button.

victoralmau:
> Well, I see more coherent as it was. If you see the tree in the after GIF, it shows the number 23 on right , so that should be the number to be shown as well in the smart-button.

Finally it is left as it is at present to avoid confusion (searchpanel self-selects with the corresponding directory).

PR #310, opened 2024-03-08
[15.0][IMP] dms_field: Add support to drag & drop

CarlosRoca13:
cc @Tecnativa TT48179

ping @victoralmau @chienandalu 

With this changes we add the drag and drop option for dms fields. On next gif you can see an example:
Ejemplo dms_field


dms_field/static/src/js/views/many_drop_target.js
```diff
@@ -0,0 +1,69 @@
+/* global base64js*/
+/* Copyright 2020 Creu Blanca
+ * Copyright 2021 Tecnativa - Alexandre D. Díaz
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+odoo.define("dms_field.DragDrop", function (require) {
```
chienandalu: Is this copied from somewhere else?

CarlosRoca13:
Changed

victoralmau (approved):

pedrobaeza:
Does this need to depend on `web_drop_target`?

CarlosRoca13:
dms is depending yet on that module

legalsylvain (approved):
Hi @CarlosRoca13. Nice work !

tested on runboat. works like a charm.

pedrobaeza:
Isn't possible to preserve previous folding state in the tree when dragging the file? Maybe adding the element by hand with jstree API instead of refreshing whole one.

CarlosRoca13:
What do you mean? When I drag the file I don't see the directories folding...
keepling_state

pedrobaeza:
The problem is reproducible when you drop on a folder with subfolders that are folded:

Peek 11-03-2024 08-36

CarlosRoca13:
@pedrobaeza Solved and improved the code to avoid full recharging when adding new directory or new field using the wizard :smile:

pedrobaeza:
Still incorrect. Now it folds all the subfolders:

Peek 11-03-2024 13-01

CarlosRoca13:
It should be working correctly right now :)

pedrobaeza (approved):
Great, thanks.

/ocabot merge minor

PR #315, opened 2024-03-26
[15.0][FIX] dms_field: Update drag and drop overlay when finish dms_field load

CarlosRoca13:
@Tecnativa TT48179

ping @pedrobaeza


pedrobaeza (approved):
Amending previous patch:

/ocabot merge patch

PR #316, opened 2024-03-27
[15.0][IMP] dms: Add default_order to file and directory tree to prevent confusion

victoralmau:
Add `default_order` to file and directory tree to prevent confusion

When using searchpanel and there is no default_order, records obtained have no defined order (search_read without order).

@Tecnativa TT48180

### DIFF
--- a/dms/views/directory.xml
+++ b/dms/views/directory.xml
@@ -377,7 +377,7 @@
         <field name="name">dms_directory.tree</field>
         <field name="model">dms.directory</field>
         <field name="arch" type="xml">
-            <tree>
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why not putting this order in the models?

victoralmau:
Order is already set in the models (`dms.file` and `dms.directory`), it is a side effect of searchpanel (the same happens elsewhere, employees for example).

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #320, opened 2024-03-27
[15.0][IMP+FIX] dms_field: Add model_ref to access groups to avoid possible errors when creating directories from a template.

victoralmau:
**[FIX]** Avoid auto-generate multiple access groups + Refactoring

Now the process will be as follows:
- The auto-generated access group will be created.
- The user will be defined in explicit_user_ids of the group created in the previous step.

In the directory created (from the corresponding record) the auto-generated group (for the user to have access) + the access groups that the template had will be used.

**[IMP]** Add `model_ref` to access groups to avoid possible errors when creating directories from a template.

Example of use case:
- Create a partner.
- Create a template linked to partners.
- Create a directory through the template.
- An access group linked to the partner is created.
- Delete the partner directory.
- Creates a new directory through the template.
- No error will occur and the previous access group will be re-used.

@Tecnativa TT48180 TT49606


pedrobaeza:
I don't get the use case for this. For what do you need this?

victoralmau:
The example use case and the steps to reproduce it are in the initial description.

**Before**
ejemplo-antes

**After**
ejemplo-despues

pedrobaeza:
But shouldn't it be then 2 fields: the model and the res_id?

victoralmau:
> But shouldn't it be then 2 fields: the model and the res_id?

I had thought about that possibility, but it would involve adding something like this https://github.com/OCA/dms/blob/15.0/dms/models/base.py#L11 (i don't like it)

pedrobaeza:
But then you are returning a group that is not correct when there are more than 1 folder. Why don't you just put a link to the directory on the group with ondelete="cascade"?

victoralmau:
Removing the access group when deleting a directory would be incorrect because the access group can be used in many other directories.

IMO there are different possibilities:
- Option A: Leave everything as it was before, an error message will be displayed and the access group will not be linked to any particular record.
- Option B: Create only the `model_ref` field in the access groups and show data in the views. Do not re-use the group. Show an error if there is any access group linked to the record (e.g. employee) so that it is removed first.

Any other option?

pedrobaeza:
I'm still not convinced of your statement that one group is for multiple folders. If I understand correctly, you link a parent folder "Employees" to the model `hr.employee`. Then, each record creates a subfolder "Employee 1", with a group allowing to access to that employee, "Employee 2" with another group, and so on...

Thus, each of these groups should be link to the directory, not to the model, as if not, we will have as many groups as employees you have. Am I wrong in something?

victoralmau:
Currently `dms_field` will create as many groups (`dms.access.group`) automatically as appropriate when creating an employee's directory and link it to that directory. This functionality is required to add the employee's user as an extra user.

The "auto generated access group" will be linked to the employee's directory but if you delete the employee's directory and try to recreate it, it will try to recreate the auto access group and the error will occur because the name already exists, so you want to add these fields.

victoralmau:
New changes added, please @pedrobaeza and @CarlosRoca13 , can you review it?

dms_field/models/dms_field_template.py
```diff
@@ -103,24 +103,43 @@ def _copy_files_from_directory(self, directory, new_directory):
...
+        """Get the existing auto-generated group or create a new one.
+        The permissions of the auto-generated group should not be changed
+        because there is “nothing” to change.
+        """
+        group_model = self.env["dms.access.group"]
+        groups_ref = group_model.search(
+            [("dms_field_ref", "=", "%s,%s" % (record._name, record.id))]
+        )
+        if groups_ref:
+            return groups_ref
```
pedrobaeza: You need to rewrite here groups + user.

dms_field/models/dms_field_template.py
```diff
@@ -103,24 +103,43 @@ def _copy_files_from_directory(self, directory, new_directory):
...
-        return groups
+    def _prepare_autogenerated_group(self, record):
+        group_name = _("Autogenerated group from %(model)s (%(name)s) #%(id)s") % {
+            "model": record._description,
+            "name": record.display_name,
+            "id": record.id,
+        }
+        vals = {
+            "name": group_name,
+            # We need to set all the permissions so that the user canmanage his
```
pedrobaeza:
```suggestion
            # We need to set all the permissions so that the user can manage their
```

dms_field/models/dms_field_template.py
```diff
@@ -103,24 +103,43 @@ def _copy_files_from_directory(self, directory, new_directory):
...
+            "dms_field_ref": "%s,%s" % (record._name, record.id),
+        }
+        # Apply sudo() because the user may not have permissions to access
+        # ir.model.fields.
+        user_field = self.sudo().user_field_id
+        if user_field:
+            user = record[user_field.name]
+            if user:
+                vals.update({"explicit_user_ids": [(4, user.id)]})
+        return vals
```
pedrobaeza, line 129: The rest of the possible groups are not inherited here. Why?

dms_field/models/dms_field_template.py
victoralmau: This is only for the auto-generated group. It is already changed to be updated.

dms_field/models/dms_field_template.py
victoralmau: If you look at the `_prepare_directory_vals()` method you will see that the groups in the directory to be created are:
- Template folder groups (`directory.group_ids`)
- Auto-generated group (with full permissions) and the corresponding explicit user.

victoralmau:
Further changes have been added to `hr_dms_field` to cover the case of modifying an employee (the auto-generated access group must be updated) and deleting it.

In v16 these changes can be added directly in `dms_field` in the `dms.field.mixin` model.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #318, opened 2024-03-27
[16.0][IMP] dms: Add move (change directory) action from files

victoralmau:
FWP from 15.0: https://github.com/OCA/dms/pull/307

Add move (change directory) action from files

Changes done:
- [x] Set tracking=True to directory_id field from files to leave log if "_moved_" to another directory.
- [x] Add `root_directory_id` field from files.
- [x] Add optional extra fields to `dms.file` tree view (`directory_id` and `root_directory_id`).
- [x] Add wizard to "move" (change directory) files.

ejemplo

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT48180

PR #323, opened 2024-04-11
[17.0][MIG] DMS : Migration to 17.0

### LINKED ISSUE #269: Migration to version 17.0
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

# Modules to migrate

- [x] dms - By @tva-subteno-it - #323
- [x] dms_attachment_link - By @victoralmau - #373
- [x] dms_storage - By @BernatObrador - #362

Missing module? Check https://github.com/OCA/maintainer-tools/wiki/%5BFAQ%5D-Missing-modules-in-migration-issue-list

tva-subteno-it:
# Migration of the DMS module to v17.

Code review is in progress, so changes might still occur.
Closes #269 

Done:
- [x] Reorganized manifest into security, actions, templates, data, views and wizard. Same for the assets, organized by styles, JS and XML
- [x] Removed inexistant assets `mail.assets_messaging`
- [x] Changed the onbaording flow, for since v17, new models `onboarding.onboarding.step` and `onboarding.onboarding` has been added. Corresponding data has been added in the `data` folder. Onboarding actions per model has been added. The `action_close` has been added to the `onboarding.onboarding` model as well as different open actions in `onboarding.onboarding.step`. Also updated `res_company` to reflect those changes
- [x] Refactoring of `portal_my_dms_directory` and `portal_my_dms` controller methods
- [x] Rename of the `directoy` and `access_group` filename to fit the model name
- [x] Fix context parsing in fields, by changing the context from a string to a dictionnary ([cf this line](https://github.com/tva-subteno-it/dms/blob/cb30f227faa36741930cf4ee3676653486add4c7/dms/models/abstract_dms_mixin.py#L34))
- [x] Fixed `unlink` method of `Base` model causing infinite loop. [See the comment](https://github.com/tva-subteno-it/dms/blob/cb30f227faa36741930cf4ee3676653486add4c7/dms/models/base.py#L26-L28) for more info
- [x] Fixed `copy` method of access group, as well as the ones in `directory` and `file` in order to display (copy) or (1) when duplicated
- [x] Fixed `_name_get` being replace by `_compute_display_name`, `attrs="{'invisible...` replaced by `invisible="...`
- [x] In `_compute_permissions` removed the `if self.env.su` since odoo already does it in the `_filter_access_rules` called a few line below
- [x] Changed `security.xml` to better handle what is the user is allowed to see/write/create/unlink by calling `_get_allowed_directory_ids` doing the job of checking what the user can do according to its access groups.
- [x] Created two new icons to fit the new design of Odoo V17
- [x] Removed useless Javascript file (not in the manifest anymore) as well as restructured the files into a more consistant structure
- [x] Changed Javascript files displaying file's preview to use the new approach, with the `useFileViewer` hook
- [x] Fixed the patch of multiple class using the solution in the [Odoo documentation](https://www.odoo.com/documentation/17.0/developer/reference/frontend/patching_code.html#applying-the-same-patch-to-multiple-objects)
- [x] Fix of the kanban view for directory and files to the new architecture. This led to some refactoring in the SCSS, to remove useless class names, and clean up some duplicated code between file/directory styles sheets
- [x] Changed the `res_settings` view according to the [Migration guide v16->v17](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#tasks-to-do-in-the-migration)
- [x] Translations (maybe after the PR is merged)
- [x] Correct code according to the code review
- [x] Add documentation, docstrings, comments
- [x] Refactor code, re-order fields by alphabetical order for more readability, follow the contributing guide for [file naming](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#file-naming). Follow the attribute model order [stated here](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#models). **Done : abstract_dms_mixin, base, dms_access_group, dms_directory**
- [x] Fixing tests


pedrobaeza:
Let's put here further conversation about this migration instead on the general migration issue.

That icon may serve. Do you have it on SVG?

tva-subteno-it:
Yes, they are both under `static/description`.

pedrobaeza:
Great, then other contributors may improve it in the future. For now seems enough.

pedrobaeza:
Please check red CIs.

dms/readme/CREDITS.md
```diff
@@ -0,0 +1,7 @@
+The migration of this module from 15.0 to 16.0 was financially supported
```
pedrobaeza: This paragraph should be removed now in this version.

tva-subteno-it:
Thanks for taking the time of doing a little code review, but I don't have finished the migration yet. Things might change over time (I'm not quite used of contributing to projects, so I don't know if I'm doing things right).
About the CIs, I think I will correct them all at once once the migration is finished, as well as documenting the code, translations etc.

tva-subteno-it:
What should I do about Copyrights ? I'm migrating this module because a customer needs it. So at my job we copyright every file with this : 
```txt
Copyright 2024 Subteno (https://www.subteno.com).
License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
```

which matches the current license used. 
Should I add the `Copyright 2024 Subteno (https://www.subteno.com).` line on every file ? And the whole thing if the file doesn't already have one ?

tva-subteno-it:
I added another icon for the portal. Similarly to the previous one, I edited it in order to fit the others portal's icon.
portal_icon
image

It was a bit hard to find the correct one, and edited it order to stay "minimalist" and follow the color scheme of the portal, but it doesn't look too odd I think.
Let me know your thoughts.

pedrobaeza:
Yes, it seems good. Don't underestimate your design skills :wink: 

About the copyright in headers, you should only add it on the files you touch.

tva-subteno-it:
In the directory model, the `file_ids` field's copy attribute is set to False, but in the `copy` method of the directory, the files are manually duplicated : 
```py
def copy(self, default=None):
    # ...
    for record in self.file_ids:
        record.copy({"directory_id": new.id})
```
Should the files be duplicated with the directory or not ? I think it can be quite dangerous if you want to duplicate a folder, but you forgot there are many files, or big files in it.

pedrobaeza:
About the copy, there are 2 aspects:

- Functional / business logic one: When you copy a folder in a file explorer, the contents are copied, so I think it's OK to copy contents as well when duplicating. If the storage is attachment, no real space duplication will happen.
- Implementation aspect: not sure why it's copy=False and done manually. Maybe when this was done (v12), copy didn't work correctly, or it was a thing of permissions due to the overrides there. As there are no code comments for knowing the reasons, we can only speculate. You can try switching to `copy=True` to see if you find any issue.

tva-subteno-it:
Well, I set `copy=True`, as well as removing the manual copy, and it worked juste fine. There was the same process for child_directories_ids, so I did the same and it worked. 
Maybe indeed it was some leftovers from the V12

tva-subteno-it:
Am I allowed to change the `ecmaVersion` in `.eslintrc.yml` ? I would like to change it to 2020, because when running pre-commit, I'm getting an error on `props.record.data?.path_json`, simply because it doesn't recognize `?.`. (even though it works).

tva-subteno-it:
Same for ruff error : `B023 Function definition does not bind loop variable` inside the lambda of a filtered. I could just ignore this specific error, because these errors are juste in the case of `.filtered`, which is unavoidable.
It would be much more difficult and less readable trying to remove this error, knowing that it's done on purpose.

pedrobaeza:
> Am I allowed to change the ecmaVersion in .eslintrc.yml ?

I would say that the official specification for v17 is still ECMA 2016. @ged-odoo can you please tell us?

pedrobaeza:
> B023 Function definition does not bind loop variable

This linter in fact is correct. We are not doing something totally correct, and should be changed. How it should be done? From the existing code:

Before:

```python
tags = record.tag_ids.filtered(lambda rec: not rec.category_id or rec.category_id == record.category_id)
```

After:

```python
tags = record.tag_ids.filtered(lambda rec, record=record: not rec.category_id or rec.category_id == record.category_id)
```

victoralmau (changes requested):
Can you unify all change commits into 1 (v17 migration) for easier review?

victoralmau:
Please, cherry-pick https://github.com/OCA/dms/pull/327 to commit history before migration v17 commit.

tva-subteno-it:
@victoralmau Everything should be done. I rebased all my commits and cherry-picked the last change. I'm currently working on the tests. Let me know if you have any question/suggestion, or anything else.

.eslintrc.yml
```diff
@@ -4,7 +4,7 @@ env:
 
 # See https://github.com/OCA/odoo-community.org/issues/37#issuecomment-470686449
 parserOptions:
-  ecmaVersion: 2019
```
victoralmau: This should not be changed here.

dms/controllers/portal.py
```diff
@@ -170,6 +138,104 @@ def portal_my_dms_directory(
...
+        allowed_directory_from_access_group = (
+            request.env["dms.access.group"]
+            .search([("users", "in", request.env.uid)])
+            .directory_ids
+        )
+        has_all_access = request.env["res.users"]._has_all_access(request.env.uid)
+        if (
+            dms_directory_id not in allowed_directory_from_access_group.ids
+            and not has_all_access
+        ):
```
victoralmau: I believe that this is not correct in all cases and should look something similar to
`res = self._dms_check_access("dms.directory", dms_directory_id, access_token)`

dms/models/dms_access_groups.py
```diff
@@ -19,6 +20,7 @@ class DmsAccessGroups(models.Model):
     perm_create = fields.Boolean(string="Create Access")
     perm_write = fields.Boolean(string="Write Access")
     perm_unlink = fields.Boolean(string="Unlink Access")
+    perm_read = fields.Boolean(string="Read Access", default=True)
```
victoralmau: This permission was removed several versions ago because it was redundant, i.e. none of these options would make sense:
- Create an access group with other permissions but no read permission.
- Create an access group without any permissions (read, create, write and delete).

Therefore, I think that this field is not necessary and any user who is in an access group, has read access (if you want to create a read-only group, you would not select any permission and it would be enough).

dms/models/dms_category.py
```diff
@@ -117,18 +106,9 @@ def _compute_count_files(self):
...
             record.count_files = len(record.file_ids)
 
-    def name_get(self):
-        if not self.env.context.get("category_short_name", False):
-            return super().name_get()
-        vals = []
-        for record in self:
-            vals.append(tuple([record.id, record.name]))
-        return vals
-
```
victoralmau: I believe this should not be removed.

dms/models/dms_directory.py
```diff
@@ -258,64 +276,15 @@ def check_access_token(self, access_token=False):
...
-    def _get_own_root_directories(self):
-        res = self.env["dms.directory"].search_read(
-            [("is_hidden", "=", False)], ["parent_id"]
-        )
-        all_ids = [value["id"] for value in res]
-        res_ids = []
-        for item in res:
-            if not item["parent_id"] or item["parent_id"][0] not in all_ids:
-                res_ids.append(item["id"])
-        return res_ids
```
victoralmau: IMO this should not be removed.

dms/models/dms_directory.py
```diff
@@ -330,22 +299,13 @@ def _inverse_model_id(self):
...
             record.res_model = record.model_id.model
 
-    def name_get(self):
-        if not self.env.context.get("directory_short_name", False):
-            return super().name_get()
-        vals = []
-        for record in self:
-            vals.append(tuple([record.id, record.name]))
-        return vals
-
```
victoralmau: IMO this should not be removed.

dms/models/dms_directory.py
```diff
@@ -630,12 +589,7 @@ def copy(self, default=None):
         else:
             names = self.sudo().parent_id.child_directory_ids.mapped("name")
         default.update({"name": unique_name(self.name, names)})
-        new = super().copy(default)
-        for record in self.file_ids:
-            record.copy({"directory_id": new.id})
-        for record in self.child_directory_ids:
-            record.copy({"parent_id": new.id})
-        return new
```
victoralmau: Is this no longer necessary?

dms/models/dms_security_mixin.py
```diff
@@ -60,25 +66,16 @@ def _compute_record_ref(self):
...
-        if self.env.su:
-            self.update(
-                {
-                    "permission_create": True,
-                    "permission_read": True,
-                    "permission_unlink": True,
-                    "permission_write": True,
-                }
-            )
-            return
```
victoralmau: IMO this should not be removed.

dms/models/res_users.py
```diff
@@ -0,0 +1,91 @@
...
+        :return: The access group that has all access
+        :rtype: odoo.model.dms_access_group
+        """
+        return self.env["dms.access.group"].search(
+            [
+                ("users", "in", user_id or self.env.uid),
+                ("has_all_access", "=", True),
+                (field_name, "=", True),
+            ]
+        )
```
victoralmau: Although I understand the idea, I do not agree with this change; the initial approach took attachment-type storage into account.

dms/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["whool"]
+build-backend = "whool.buildapi"
```
victoralmau, line 3: I believe this change should not be necessary here.

dms/views/menu.xml
```diff
@@ -2,33 +2,73 @@
...
-    />
-    <menuitem
-        id="cat_menu_dms_config_system"
-        name="System"
-        groups="dms.group_dms_manager"
-        parent="cat_menu_dms_config"
-        sequence="25"
-    />
+    >
+        <menuitem
```
victoralmau: No indentation is necessary here.

victoralmau (changes requested):
Good job, some things:

- It would be missing the pre-commit commit before the v17 migration commit ([IMP] dms: pre-commit auto fixes) https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0 I think that would help to "reduce" the changes in the migration commit.

- Can you indicate all the changes you have done in the migration commit?

- On the other hand, other improvements (adding the `has_all_access` field from `dms.access.group` or `is_first_creation` field in dms.directory) should go in another commit [IMP] after the migration commit OR do it in another PR after merging this (to make it easier to review this PR).

.eslintrc.yml
tva-subteno-it: The problem is that if I leave to 2019, I get a warning with `?.` (like `object?.attribute`) even though it works.

dms/pyproject.toml
tva-subteno-it: I didn't change anything ?  This file is just coming from v16.

dms/models/dms_directory.py
tva-subteno-it: Yes, I think we discussed it above with @pedrobaeza. This code was probably added when Odoo wasn't supporting yet duplication of relational fields. But I tested without it and it's perfectly functional.

dms/models/dms_security_mixin.py
tva-subteno-it: The Odoo standard already checks for `self.env.su` in the method `_filter_access_rule` ([see here](https://github.com/odoo/odoo/blob/3cfe548ee03f96713c7a6d462409be9c825d22e7/odoo/models.py#L4133-L4136)) called just after, so if the user is `su` then Odoo already checks it.

dms/views/menu.xml
tva-subteno-it: Isn't it just more readable to have the list of menuitem indented ?

dms/models/dms_category.py
tva-subteno-it: If you look at [the migration guide v16 -> v17](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0) you can see that `name_get` doesn't exist anymore and has been replaced by `_compute_display_name` which doesn't return anything.

victoralmau:
Please, cherry-pick https://github.com/OCA/dms/pull/318 to commit history (before migration v17 commit).

tva-subteno-it:
> Please, cherry-pick #318 to commit history (before migration v17 commit).

Done. Also, I edited the comment of this PR to list what's still to be done, and what's already done.

victoralmau:
> > Please, cherry-pick #318 to commit history (before migration v17 commit).
> 
> Done. Also, I edited the comment of this PR to list what's still to be done, and what's already done.

Ok, all my comments are pending.

tva-subteno-it:
> > > Please, cherry-pick #318 to commit history (before migration v17 commit).
> > 
> > 
> > Done. Also, I edited the comment of this PR to list what's still to be done, and what's already done.
> 
> Ok, all my comments are pending.

Hi, I've just returned from vacation. I've already replied to some of your comments, but I'm working on the rest.

dms/models/dms_directory.py
tva-subteno-it: There is no more need to use it, since now the ir.rule already handles what the user is able to search/write/create/unlink or not based on it access group.

dms/models/dms_access_groups.py
tva-subteno-it: Yeah I get your point. But I don't think we should remove it, because the `perm_read` field match the `perm_read` of ir.rule. That way, when overriding `_get_domain_by_access_groups`, `_search` or any "checking" method, we can keep everything the same between create, unlink or whatever without having any special behavior for the read. What I did in the meantime, is to set perm_read as reandonly and removing it from the view.

victoralmau:
Please, cherry-pick https://github.com/OCA/dms/pull/332 to commit history (before migration v17 commit).

victoralmau:
Please, cherry-pick https://github.com/OCA/dms/pull/334 to commit history (before migration v17 commit).

tva-subteno-it:
I've corrected the unit tests. I also removed the code I added about the `has_all_access` and `is_first_creation`. I managed to find more *Odoo vanilla* approach.
But I have a question about the translations. When will the module be translated ? After this PR is merged ? For the moment the only translations are the ones coming from the v16.

tva-subteno-it:
For the moment, I edited the fr.po file myself since I'm french. It shouldn't cause any problem with Weblate (I think) because the project is currently not available for this version on Weblate.

pedrobaeza:
OK, no problem, as indeed there won't be conflict. Thanks for the migration work.

/ocabot migration dms

victoralmau (changes requested):
Hello again,

I see that some of my comments from the previous revision are still pending.

I have also seen that `perm_read` field (and other similar fields are added in `dms.access.group`) and it is not necessary https://github.com/OCA/dms/pull/323#discussion_r1579290730

Best regards.

dms/models/dms_directory.py
tva-subteno-it: I already answered, but this conversation might be lost somewhere... If you look at the migration guide [v16 -> v17](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0) you can see that now _name_get is deprecated and no longer used.

dms/models/dms_access_groups.py
tva-subteno-it: Anyway, I found a way of getting of `perm_read`.

tva-subteno-it:
Hi @victoralmau,
I resolved all the comments. I also fixed a bug raised in the issue of this PR. 
About the `perm_read` field, you will see that I finally found a way to remove it.
You left a comment about the indentation in the `menu.xml` file. The problem is that I'm not sure what you meant, so I changed it a bit, but pre-commit edit this file anyway, so I didn't manage to remove all the indents.

Feel free to check again, and tell me if you see anything left to improve/fix.

victoralmau:
Please, cherry-pick https://github.com/OCA/dms/pull/338 to commit history (before migration v17 commit).

tva-subteno-it:
> Please, cherry-pick #338 to commit history (before migration v17 commit).

Done

tva-subteno-it:
How does the process works now? Will someone check the code, or something like that? When do you think this can be merged?

dms/controllers/portal.py
```diff
@@ -24,42 +36,44 @@ def _dms_check_access(self, model, res_id, access_token=None):
     def _prepare_home_portal_values(self, counters):
         values = super()._prepare_home_portal_values(counters)
         if "dms_directory_count" in counters:
-            ids = request.env["dms.directory"]._get_own_root_directories()
-            values["dms_directory_count"] = len(ids)
```
victoralmau: Please don´t remove the `_get_own_root_directories()` method, it is necessary.

Example use case:
- Directory A
-- Subdirectory A1
-- Subdirectory A2

Depending on the user's permissions, he may only have access to the following directories:
- Subdirectory A1
- Subdirectory A2

and those directories should appear as “_roots_”, for example in portal view.

dms/models/dms_security_mixin.py
```diff
@@ -60,25 +67,15 @@ def _compute_record_ref(self):
...
```
victoralmau: Why is this being removed? I think it should be kept.

dms/models/res_users.py
```diff
@@ -0,0 +1,116 @@
...
+        :rtype: list[int]
+        """
+
+        # get the directories based on a record
+        attachment_directories = self.env["dms.directory"].search(
+            [
+                ("res_id", "not in", [False, 0]),
+                ("res_model", "!=", False),
+            ]
+        )
```
victoralmau:
```suggestion
        attachment_directories = self.env["dms.directory"].search(
            [
                ("storage_id.inherit_access_from_parent_record", "=", True),
            ]
        )
```

dms/models/res_users.py
```diff
@@ -0,0 +1,116 @@
...
+            if model is not None:
+                record = model.browse(directory.res_id)
+                # standard way to check access rights
+                invalid = record - record.with_user(
+                    record.env.user
+                )._filter_access_rules_python(operation.replace("perm_", ""))
+                if not invalid.exists():
+                    authorized_directory_ids.append(directory.id)
+
+        return authorized_directory_ids
```
victoralmau: IMO this change is not necessary, it is confusing and incorrect (I see `.sudo()` is applied at the beginning, which will cause all records queried to have the 'wrong permissions'). can we leave it as is in v16?

victoralmau (changes requested):

dms/models/dms_security_mixin.py
tva-subteno-it: It's been removed because the `_filter_access_rules` called below is already checking if `self.env.su` : 

```py
# see odoo/models._filter_access_rules
if self.env.su:
    return self
```

So we were duplicating the same logic.

pedrobaeza:
Please cherry-pick #342

dms/models/res_users.py
tva-subteno-it: In V16 the handle of record related access was not working. You can try it out easily in v16 :
- Install the *Sale* app
- Create a new attachment storage for the `sale.order` model, and check "inherit permissions" as well as "create files form message attachments"
- Create a new order where the user is the admin, and add an attachment
- Give only the "User: Own document only" permission to a second user (not the admin) for sale order
- Log in with the second user. It can see the new directory as well as all the files even though he should not be able to see it (because it is not its order).
- You can even remove all access to sale for the second user
- Despite loosing access to all sales (the module is not even appearing in the menu), the second user can still see all files/directory related to the order

In order to match to your suggestion, I have modified the code accordingly, removing the `sudo()` (it was useless anyway since this method is called in ir_rule, therefore `su` is already True) and using the `storage_id_inherit_access_from_parent_record` field. Once I push, you will see the changes I made.

dms/controllers/portal.py
tva-subteno-it: Okay, there was a misunderstanding on my side : I made it so that if you want to see Subdirectory A1, then I display (emptied) Directory A, leading to Subdirectory A1.

tva-subteno-it:
Hi @victoralmau I pushed changes. I hope it's what you had in mind. There is still a conversation pending, in case you want to respond or need more details about this problem, but you can resolve it if you think it's ok.

tva-subteno-it:
> Please cherry-pick #342

Done

dms/models/dms_security_mixin.py
```diff
@@ -60,25 +66,15 @@ def _compute_record_ref(self):
...
```
victoralmau: Why is this being removed? I think it should be kept.

dms/models/dms_security_mixin.py
```diff
@@ -202,19 +194,30 @@ def _get_permission_domain(self, operator, value, operation):
...
     def _search_permission_create(self, operator, value):
+        if self._name == "dms.directory":
+            all_access_group = self.env["dms.access.group"].search(
+                [
+                    ("users", "in", self.env.uid),
+                    ("perm_create", "=", True),
+                ]
+            )
+            if all_access_group:
+                return []
```
victoralmau: This is not correct, it should be removed.

dms/models/res_users.py
```diff
@@ -0,0 +1,151 @@
...
+                    record.with_env(
+                        self.env(su=self.env.user._is_superuser())
+                    ).check_access_rule(operation.replace("perm_", ""))
+                except AccessError:
+                    authorized = False
+
+                if authorized:
+                    authorized_directory_ids.append(directory.id)
+
+        return authorized_directory_ids
```
victoralmau: IMO I still think this is not necessary (in any case it would be an improvement, not something necessary for migration). This change would be replicating what already exists in the defined `ir.rules`.

dms/pyproject.toml
victoralmau, line 3: I believe this change should not be necessary here.

dms/security/security.xml
```diff
@@ -115,7 +116,9 @@
         <field name="perm_create" eval="1" />
         <field name="perm_write" eval="0" />
         <field name="perm_unlink" eval="0" />
-        <field name="domain_force">[('permission_create', '=', user.id)]</field>
+        <field
+            name="domain_force"
+        >[('id', 'in', user._get_allowed_directory_ids("perm_create"))]</field>
```
victoralmau: IMO this changes (and similar) is not necessary (as discussed in the `res.users` file changes).

dms/template/portal.xml
```diff
@@ -0,0 +1,132 @@
...
+<odoo>
+    <template
+        id="portal_my_home_dms"
+        name="Portal My Home : dms entries"
+        inherit_id="portal.portal_my_home"
+        priority="20"
+    >
+        <xpath expr="//div[@id='portal_client_category']" position="inside">
+            <t t-call="portal.portal_docs_entry">
+                <t t-set="title">Your Documents</t>
```
victoralmau:
```suggestion
                <t t-set="title">Documents</t>
```

victoralmau (changes requested):
The changes to `.copier-answers.yml,` `.editorconfig`, `.pre-commit-config.yaml`, `.ruff.toml`, `.github/workflows` and `pandoc-3.1.12.3-x86_64-macOS.pkg` files are not necessary, right?

Please, before resolving some conversation, reply about it to clarify if it is resolved and why, otherwise it is impossible to know what is really resolved and especially why.

dms/pyproject.toml
tva-subteno-it: I didn't change anything. This is just coming the v16 project.

dms/security/security.xml
tva-subteno-it: As I said in the `res.users` discussion, this is not working in v16, there is no point in comparing a boolean to an id, it will always be True. But I can, if needed make it a FIX commit.

dms/models/res_users.py
tva-subteno-it: Well, what was exsting in ir_rule was not working. The domain was comparing a boolean with the user id, which was of course always True. If you test locally the v16, access groups are not working at all.
If you want, I can leave as is for migration, and add a FIX commit, but in v16 no matter the access group everyone has access to everything.

tva-subteno-it:
> The changes to `.copier-answers.yml,` `.editorconfig`, `.pre-commit-config.yaml`, `.ruff.toml`, `.github/workflows` and `pandoc-3.1.12.3-x86_64-macOS.pkg` files are not necessary, right?
> 
> Please, before resolving some conversation, reply about it to clarify if it is resolved and why, otherwise it is impossible to know what is really resolved and especially why.

About the changes in those files, I haven't manually touched it. I've cloned them directly from v16. If you see some changes compared to v16 then it has to be a bot, like Copier. I remember that at some point I cherry-picked a commit that made changes to those files because after that, pre-commit was no longer passing.

pedrobaeza:
You don't have to cherry-pick, but do a rebase.

dms/models/dms_security_mixin.py
tva-subteno-it: The method `_filter_access_rules` called just below that already starts by checking if `self.env.su` and then returns `self`. The previous code was just duplicating the same behavior.

tva-subteno-it:
> You don't have to cherry-pick, but do a rebase.

Well no I was wrong, I didn't cherry picked anything, but the branch 17.0 was updated on the 1st of May. You can go check the 17.0 branch, and you'll see that it is the same as what is currently on my migration branch. The other files were already there when I clone.

pedrobaeza:
I insist: doing a proper rebase, that modifications won't be here.

dms/models/dms_security_mixin.py
tva-subteno-it: Indeed, the `_get_permission_domain` already handles it. 
I removed it.

dms/template/portal.xml
tva-subteno-it: Okay. I changed.

pedrobaeza:
The rebase is correct now, but you have added the file pandoc... in one of your commits. Please remove it in the same commit.

tva-subteno-it:
> I insist: doing a proper rebase, that modifications won't be here.

Yeah you're right. I don't know what I was thinking, maybe I thought I already did it. Anyway, I rebased and as you said, all those changes are not showing up anymore.

dms/pyproject.toml
tva-subteno-it: Well, in I just had to rebase... Now it's done, and this changes should not appear anymore.

tva-subteno-it:
> The rebase is correct now, but you have added the file pandoc... in one of your commits. Please remove it in the same commit.

I removed it. I don't even know what it is, so when I got it in the first place I just left it there ><

pedrobaeza:
pre-commit seems to download it sometimes for README regeneration. Not sure why.

tva-subteno-it:
Hi @victoralmau, I replied to some of your reviews, can you please tell me what do you think ?

dms/models/res_users.py
victoralmau: This is still pending

dms/models/dms_security_mixin.py
victoralmau: This is still pending

dms/models/res_users.py
victoralmau: This is still pending

dms/models/res_users.py
tva-subteno-it: I know, I was waiting for a response from your side, that's why I didn't resolve it. I wanted to know if you agreed with me (or not) before resolving the conversation. But maybe that's not the way I should do things?

dms/models/res_users.py
tva-subteno-it: I know, I was waiting for a response from your side, that's why I didn't resolve it. I wanted to know if you agreed with me (or not) before resolving the conversation. But maybe that's not the way I should do things?

victoralmau:
Please, cherry-pick https://github.com/OCA/dms/pull/344 to commit history before migration v17 commit.

tva-subteno-it:
Hello @victoralmau, I really would like to know what I should do once you add comments on the code. You raised some issue for which I replied. Later on you said they were still pending, but I was waiting for a reply from you after I explained why I did this.
Then what's the procedure, will you reply and tell me when you think it's good, or once I explain myself I can mark the conversations as resolved immediately?

victoralmau:
Please answer to what is pending in each conversation (as far as I know, what I commented that is still pending is because the need for that change has not been explained).

https://github.com/OCA/dms/pull/323#discussion_r1630700490

https://github.com/OCA/dms/pull/323#discussion_r1630697414

https://github.com/OCA/dms/pull/323#discussion_r1622188272

tva-subteno-it:
Okay, I figured it out... Looks like I have to submit my review, and simply replying is not enough 😆 
Luckily I'm not the only one struggling (see this [dicsussion](https://github.com/orgs/community/discussions/10369)).
My bad, you never saw any of my comments...

dms/models/res_users.py
victoralmau: Sorry, I had missed this.

If something doesn't work in v16 it should be resolved in v16 first.
It is intentional that the access rules compare with `user.id`, the goal is that the `_compute_permissions()` method set everything correctly (https://github.com/OCA/dms/pull/326).

Checking the use case you indicate in v16 in runboat, it works correctly.

Use case:

Marc Demo: Permissions Document > User and Sale > User: Own Documents Only (**no more**)

storage
carpetas

S00007 > Order is assigned to Mitchell Admin
S00016 > Order is assigned to Marc Demo

If we log in with Marc Demo, he can only see the folder **S00016**.
ejemplo-marc-demo

Having cleared all this up, if it works correctly in v16, it should also work in v17 without all these changes (similar to v16).

tva-subteno-it: Indeed, I don't know what I did, but following your instructions everything works. I rolled back all the changes I made about it.

dms/models/res_users.py
tva-subteno-it: I removed it since it works fine with the help of your input on another conversation.

dms/security/security.xml
tva-subteno-it: I rollbacked as what it was.

tva-subteno-it:
Hello, @victoralmau. Thanks for your answer, indeed it worked just fine, I might have forgotten a step in the process. Since everything was fine, I rolled back my changes.

dms/models/res_users.py
victoralmau: Ok, you forgot to remove all the `res.users` code which is now no longer needed.

tva-subteno-it: Yeah, you're right, my bad. I removed it, but when I pushed I had a conflit. 
I removed it (again).

dms/models/dms_directory.py
```diff
@@ -0,0 +1,806 @@
...
+            record.count_total_files = (
+                model.search_count([("directory_id", "child_of", record.id)])
+                if record.id
+                else 0
+            )
+
+    def _compute_count_total_elements(self):
+        for record in self:
+            record.count_total_elements = (
+                record.count_elements + record.count_total_directories
```
victoralmau:
```suggestion
                record.count_total_files + record.count_total_directories
```

victoralmau (changes requested):
Good job, after checking it functionally:

- The image shown of the file in the portal if we access with access_token (share option) does not work.

dms/models/dms_directory.py
tva-subteno-it: Ok I understand. It was because portal user doesn't have read access to the file, therefore can't access the icon field.
I pushed a fix, you can try.

victoralmau:
Using the Share link in a directory shows the files in the portal (ok) but not the directories, can you check it?

tva-subteno-it:
> Using the Share link in a directory shows the files in the portal (ok) but not the directories, can you check it?

Yes, I'm going to have a look at that, but I'm guessing it's also a problem of access right...

tva-subteno-it:
Done, but I have a question: If I share a directory with sub directory in it, is the portal user meant to have access to the sub directories as well (and so on)?

victoralmau:
> Done, but I have a question: If I share a directory with sub directory in it, is the portal user meant to have access to the sub directories as well (and so on)?

Thank you, I will check it out soon.

Exactly, if you have access to a directory with access_token you have access to everything in it (including any level).

tva-subteno-it:
> Thank you, I will check it out soon.
> 
> Exactly, if you have access to a directory with access_token you have access to everything in it (including any level).

Okay, I fixed it.

victoralmau (approved):
Code and functional review OK.

Ping @CarlosRoca13

CarlosRoca13 (approved):
Code and functional review :+1: Thanks :smile:

dms/models/dms_access_groups.py
```diff
@@ -39,74 +49,100 @@ class DmsAccessGroups(models.Model):
         store=True,
         recursive=True,
     )
-
     directory_ids = fields.Many2many(
-        comodel_name="dms.directory",
-        relation="dms_directory_groups_rel",
```
pedrobaeza: Why are you removing these attributes? It's better to be explicit in this for avoiding collisions.

dms/models/dms_directory.py
```diff
@@ -127,77 +126,70 @@ def _default_parent_id(self):
...
     )
-
     file_ids = fields.One2many(
+        string="Files",
         comodel_name="dms.file",
         inverse_name="directory_id",
-        string="Files",
         auto_join=False,
-        copy=False,
+        copy=True,
```
pedrobaeza: This is changing the behavior. AFAIK, the problem here is that we can't assure the security consistency on the copied files.

dms/models/abstract_dms_mixin.py
```diff
@@ -7,10 +8,13 @@ class AbstractDmsMixin(models.AbstractModel):
...
     _description = "Abstract Dms Mixin"
 
+    # ----------------------------------------------------------
+    # Fields
+    # ----------------------------------------------------------
     name = fields.Char(required=True, index="btree")
     # Only defined to prevent error in other fields that related it
     storage_id = fields.Many2one(
-        comodel_name="dms.storage", string="Storage", store=True, copy=True
+        string="Storage", comodel_name="dms.storage", store=True, copy=True
```
pedrobaeza: Although I understand that you are applying the same order for the attributes, first of all, it's not very important, as they are keyword arguments, so any parser won't have problem with any order, and it's only let's say a human commodity (and not for all), but the most important thing is that this is increasing a lot the noise on the migration commit, so please refrain from doing this on modules already existing. If you want to do it, please do it on a separate commit with `[REF]` tag, but I think it's not needed at all.

dms/models/abstract_dms_mixin.py
```diff
@@ -1,4 +1,5 @@
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# Copyright 2024 Subteno - Timothée Vannier (https://www.subteno.com).
```
pedrobaeza: You are not doing real changes on this file, so better to not modify copyright.

dms/models/abstract_dms_mixin.py
```diff
@@ -7,10 +8,13 @@ class AbstractDmsMixin(models.AbstractModel):
     _name = "abstract.dms.mixin"
     _description = "Abstract Dms Mixin"
 
+    # ----------------------------------------------------------
```
pedrobaeza: Although I know you are mimicking it, this comment format is taking 3 lines for only expressing one thing, so it's better to avoid it. The current way of distributing the code doesn't need extra comments IMO, but if you want to put them, just put one line like:
```suggestion
    # FIELDS
```

dms/models/dms_access_groups.py
```diff
@@ -1,17 +1,27 @@
...
 # Copyright 2020 RGB Consulting
+# Copyright 2024 Timothée Vannier - Subteno (https://www.subteno.com).
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
 
 class DmsAccessGroups(models.Model):
+    # ----------------------------------------------------------
```
pedrobaeza: This comment is totally unneeded. They are clear and put at the beginning without empty lines in the middle, so they are very recognizable.

dms/models/dms_access_groups.py
```diff
@@ -39,74 +49,100 @@ class DmsAccessGroups(models.Model):
...
     directory_ids = fields.Many2many(
-        comodel_name="dms.directory",
-        relation="dms_directory_groups_rel",
         string="Directories",
-        column1="gid",
-        column2="aid",
-        auto_join=True,
-        readonly=True,
+        comodel_name="dms.directory",
+        readonly=False,
```
pedrobaeza: Why changing to `False`? And any case, it's the default value, so you can remove it.

dms/models/dms_access_groups.py
```diff
@@ -39,74 +49,100 @@ class DmsAccessGroups(models.Model):
...
+                )
+
+    # ----------------------------------------------------------
+    # Default methods
+    # ----------------------------------------------------------
+    @api.model
+    def default_get(self, fields_list):
+        """
+        When creating a new group, add the current user to the explicit users.
+        """
```
pedrobaeza: Incorrect PEP257 docstring format. It can be one line, but if it doesn't fit, it should be this way:
```suggestion
        """When creating a new group, add the current user to the explicit users.
        """
```

dms/models/dms_access_groups.py
```diff
@@ -121,24 +157,14 @@ def _compute_inclusive_permissions(self):
...
-                    % perm: (
+                    "perm_inclusive_%s" % perm: (
                         one["perm_%s" % perm]
                         or one.parent_group_id["perm_inclusive_%s" % perm]
                     )
                     for perm in ("create", "unlink", "write")
                 }
             )
 
-    @api.model
```
pedrobaeza: Please reduce the noise in the migration commit. Do it if you want on a REF one, but again, I advise you to not do it on code that is not yours, as everyone has a different style, and we should focus on the things that brings value to the community.

dms/models/dms_access_groups.py
```diff
@@ -148,22 +174,21 @@ def default_get(self, fields_list):
...
+    # ----------------------------------------------------------
+    # CRUD/ORM Overrides
+    # ----------------------------------------------------------
+    def copy(self, default=None):
+        default = dict(default or {})
+        default["name"] = _("%s (copy)") % self.name
+        return super().copy(default=default)
+
+    def write(self, vals):
+        self.env.registry.clear_cache()
```
pedrobaeza: Why this? When you do weird things, you must always put a comment saying why. If not, any other review/migration, this will be removed if we don't find the reason why it's put.

dms/models/dms_category.py
```diff
@@ -21,11 +22,9 @@ class Category(models.Model):
     _rec_name = "complete_name"
 
     # ----------------------------------------------------------
-    # Database
+    # Fields
```
pedrobaeza: As this comment format existed previously (which I have already said that seems too many lines for one comment), and you are changing it, I think you can reduce it as recommended or remove it all.

dms/models/dms_file.py
```diff
@@ -34,24 +34,21 @@ class File(models.Model):
...
         help="If a file is set to archived, it is not displayed, but still exists.",
     )
-
     directory_id = fields.Many2one(
-        comodel_name="dms.directory",
         string="Directory",
+        comodel_name="dms.directory",
         domain="[('permission_create', '=', True)]",
-        context="{'dms_directory_show_path': True}",
+        context={"dms_directory_show_path": True},
```
pedrobaeza, line 51: No need to change this.

dms/models/dms_file.py
```diff
@@ -98,48 +90,63 @@ class File(models.Model):
...
     )
-
     migration = fields.Char(
-        compute="_compute_migration",
         string="Migration Status",
+        compute="_compute_migration",
         readonly=True,
         prefetch=False,
         compute_sudo=True,
+        store=True,
```
pedrobaeza: Why doing it stored? I don't think it's needed and it can lead to side effects if the depends is not correct.

dms/models/dms_file.py
```diff
@@ -527,48 +296,322 @@ def _check_name(self):
...
             binary = base64.b64decode(record.content or "")
             values = record._update_content_vals(values, binary)
             updates[tools.frozendict(values)].add(record.id)
-        with self.env.norecompute():
-            for vals, ids in updates.items():
-                self.browse(ids).write(dict(vals))
+        for vals, ids in updates.items():
+            self.browse(ids).write(dict(vals))
+
+    def copy(self, default=None):
```
pedrobaeza: It's impossible to follow the real changes with the reorder you have made on the methods. Please undo it, and leave it to the refactor (or accept it as it is, as working in community sometimes mean this, to resign to certain things that are not totally of our taste on the mood of focusing on the real jobs, which is to migrate this).

I can't continue reviewing this file meanwhile.

pedrobaeza (changes requested):
I have started to review the migration commit, but it's barely impossible due to all the "style changes". Please refrain from doing them on the migration commit. You may heard that squashing commits should be done, but not at this level. The commit should be the unit to ease the review, so you must squash when you do a progressive migration refinement through several commits, not when you are doing a total move of every line of code for placing according a specific order. I agree that is good to have the code sorted according certain guidelines, but the first rule is to minimize the diff on any existing code. Please check the rest of my inline comments where I expand these thoughts.

dms/models/dms_access_groups.py
tva-subteno-it: Does that mean that you want me to re-write every docstring of every method?

dms/models/dms_access_groups.py
tva-subteno-it: That's some old code I might have written. I'm removing it.

dms/models/dms_directory.py
tva-subteno-it: What do you mean? Everything should be copied just fine. The copy method in dms_file.py was not doing much than what Odoo now already does.

dms/models/dms_file.py
tva-subteno-it: To change what? If you compare with like the v16 or v15 I only removed the outer quotes according to the [official doc](search_panel_select_multi_range) for the context param to be a dict and not a string

dms/models/dms_file.py
tva-subteno-it: Got it, I'm currently working on reverting every changes that are not migration related (such as file renaming, comments and reordered fields/methods.

tva-subteno-it:
@pedrobaeza it should be all good now. I think I've reversed all the reordering I did, and you can know more easily do the code review.

dms/models/base.py
```diff
@@ -19,6 +20,14 @@ def unlink(self):
...
             domain = [("res_model", "=", self._name), ("res_id", "in", self.ids)]
-            self.env["dms.file"].sudo().search(domain).unlink()
-            self.env["dms.directory"].sudo().search(domain).unlink()
+            # Has to check if existing before unlinking, because even if the search
+            # returns an empty recordset, it will still call the unlink method on it.
+            # This can result in an infinite loop and a recursion depth error.
+            files = self.env["dms.file"].sudo().search(domain)
+            if files:
+                files.unlink()
+
```
pedrobaeza: Please don't empty lines inside a method.

dms/models/base.py
```diff
@@ -19,6 +20,14 @@ def unlink(self):
             and self._name not in ("dms.file", "dms.directory")
         ):
             domain = [("res_model", "=", self._name), ("res_id", "in", self.ids)]
-            self.env["dms.file"].sudo().search(domain).unlink()
-            self.env["dms.directory"].sudo().search(domain).unlink()
+            # Has to check if existing before unlinking, because even if the search
```
pedrobaeza, line 23: Great explanation to know why to put the `if`!

dms/models/directory.py
```diff
@@ -89,11 +93,8 @@ def _default_parent_id(self):
             return False
 
     group_ids = fields.Many2many(
-        comodel_name="dms.access.group",
```
pedrobaeza: Why removing the explicit relation table and fields? Is it better to fix it to avoid changes in the future if the ORM changes.

dms/models/directory.py
```diff
@@ -248,15 +249,17 @@ def check_access_token(self, access_token=False):
...
                 if item.id == self.id:
                     return True
-                else:
-                    directory_item = self
-                    while directory_item.parent_id:
-                        if directory_item.id == item.id:
-                            return True
-                        directory_item = directory_item.parent_id
-                    # Fix last level
+
```
pedrobaeza: Please don't add empty lines inside a method. It's less readable and requires more scroll while having syntax highlighting is enough. I don't repeat the comment more times, but the same for the entire modifications.

dms/models/dms_file.py
pedrobaeza: The link doesn't work to the documentation you mention doesn't work. Can you correct it?

pedrobaeza (commented):
Thanks for the efforts and your patience. I think we are almost there. Just some final comments, and I don't nickpit more.

dms/models/base.py
tva-subteno-it: Thanks! I thought it would be better to write a comment to prevent another dev from removing it by thinking it's not useful

dms/models/dms_file.py
tva-subteno-it: Oh yeah you're right, sorry. Here is the new [new link](https://www.odoo.com/documentation/17.0/developer/reference/backend/orm.html#odoo.fields.Many2one). If you search for other places where they use the context in fields, you'll see that they don't use quotes round the dict.

pedrobaeza: Thanks for the link. Well, the option to provide strings is also supported, and it's useful when you need to put a dynamic expression depending on a field value for example, but in this case, it's the same. It was only for reducing the diff of the migration.

tva-subteno-it:
@pedrobaeza I fixed what you asked, and sent the right url this time (I hope).

dms/models/dms_file.py
tva-subteno-it: Does that mean that I can leave as is?

pedrobaeza (approved):
OK, there are still some stylistic things (like empty lines inside methods, as I didn't mention each of them), but let's merge it for not disturbing you more. Thanks for the work.

/ocabot merge nobump

PR #324, opened 2024-04-15
[15.0][IMP] dms: Add number items (directories and files) to directory kanban view

victoralmau:
Add number items (directories and files) to directory kanban view.

ejemplo

@Tecnativa TT48648

### DIFF
--- a/dms/static/src/scss/directory_kanban.scss
+++ b/dms/static/src/scss/directory_kanban.scss
@@ -27,10 +27,14 @@
                         border-bottom: solid 1px $gray-400;
                     }
                 }
+                span.total_items {
+                    margin-left: 3px;
+                    margin-top: 3px;
+                }
             }
         }
         .o_kanban_image {
-            width: $o-kanban-image-width + 1;
+            width: $o-kanban-image-width - 1;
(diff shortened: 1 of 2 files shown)


dms/static/src/scss/directory_kanban.scss
```diff
@@ -27,10 +27,14 @@
                         border-bottom: solid 1px $gray-400;
                     }
                 }
+                span.total_items {
```
pedrobaeza, line 30: Why do you need this CSS adjustments?

pedrobaeza (approved):

dms/static/src/scss/directory_kanban.scss
victoralmau: Some reasons:
- It is necessary that the text is not next to the icon (the size of the icon has been reduced).
- The left block has no defined width, therefore, another block (the one with the image, which has a fixed width) must be reduced "a little" to make all the content fit.

CarlosRoca13 (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #325, opened 2024-04-16
[15.0][IMP] dms_field: Add Parent directory option to templates

victoralmau:
A parent directory can be set in templates, if set, the directory that is created linked to a record will be a "child" of that directory.

@Tecnativa TT48650

### DIFF
--- a/dms_field/models/dms_field_template.py
+++ b/dms_field/models/dms_field_template.py
@@ -20,6 +20,11 @@ class DmsFieldTemplate(models.Model):
         domain=[("save_type", "!=", "attachment")],
         string="Storage",
     )
+    parent_directory_id = fields.Many2one(
+        comodel_name="dms.directory",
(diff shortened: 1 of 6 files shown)


dms_field/models/dms_field_template.py
```diff
@@ -75,9 +80,10 @@ def create_dms_directory(self):
         template = self._get_template_from_model(res_model)
         if not template:
             raise ValidationError(_("There is no template linked to this model"))
+        parent_id = self.parent_directory_id.id if self.parent_directory_id else False
```
pedrobaeza:
```suggestion
        parent_id = self.parent_directory_id.id
```

pedrobaeza (approved):

CarlosRoca13 (approved):

victoralmau:
It's ready to merge, right?

PR #326, opened 2024-04-19
[15.0][FIX] dms: security boolean fields

Tardo:
These fields are booleans.


Tardo:
Ok, I see that this is intentional...

PR #329, opened 2024-05-03
[15.0][IMP] dms_auto_classification: Add folder support within the .zip file

victoralmau:
Add folder support within the `.zip` file

Changes done:
- [x] The full path is now shown in the detail line.
- [x] The dms file that will be created will be with the name (not the full file path).
- [x] When scanning the `.zip` files the folders are skipped.
- [x] Add more use cases in tests
- [x] Improve documentation

@Tecnativa TT49047


pedrobaeza:
You have to more specific with your commit messages. What "folder support" means in this case? If I'm not mistaking, what you are doing is to gather all the files, no matter if it's in any subfolder,so that's not strictly folder support to begin with, but as said, you have to be always more specific. "Devil is in the details". As said on our internal task, the ideal is that the pattern supports folders (in this case the denomination is correct). Meanwhile, you have to include this new behavior in the README and put the ROADMAP, although I would prefer to implement the feature at once. If not, migrating all the old classifications will mean more work than the implementation itself. Another thing you can do is a mix:

- Add a check to indicate to "ignore folders", and apply this new algorithm.
- Add the roadmap for adding the folders in the pattern.

This way, all the existing classification templates with the check will preserve its behavior.

victoralmau:
I comment a little about some issues.

1- Allow folders in patterns.
Actually this was possible before and it is possible now, `re.search(pattern, filename)` is used, so you can set the pattern you want to "filter" by folder or whatever you need. Let's not confuse this functionality (to use or not the file example.text with the name that will have the file that will be created).

2- Incorrect use of "folders" as files.
Previously we used `namelist()` that already returned all the files (including folders), and the filename contained the name of the folder (test1/test.txt for example). 
folder (test1/test.txt for example).

All this has led to some fixes:

A- Folders should be skipped, for that reason `infolist()` is now used, which allows to know more easily if it is a folder or not.
B- When trying to create a file, only the file name should be used (therefore everything before / is omitted because it will refer to the folder regardless of the level). 

Having clarified all this, are these fixes necessary? I think so, but if you think not I add it in ROADMAP as pending.

pedrobaeza:
Having read what you put, and if I have understood well, the patch is not correct:

- You should keep complete path in the preview popup.
- ~~As said, all the files shouldn't be considered by default. Add the check for having such behavior only by opt-in.~~ Let's document the behavior in the README

victoralmau:
Changes done and updated PR description.

dms_auto_classification/README.rst
```diff
@@ -45,12 +45,15 @@ Example of filename pattern to use only .txt files: .txt$
 Directory pattern example 1: Documents
 Directory pattern example 1: Documents / Data
```
pedrobaeza, line 46: Is this correct? Shouldn't it be `Documents/Data`?

dms_auto_classification/README.rst
```diff
@@ -45,12 +45,15 @@ Example of filename pattern to use only .txt files: .txt$
...
 
+The pattern will be applied to the full file path (including folders): test/test-1.txt
```
pedrobaeza:
```suggestion
If the pattern doesn't contain path separators (/), the file name will be search across all the subdirectories.
```

dms_auto_classification/README.rst
victoralmau: Yes, Documents / Data is correcct according to `complete_name` directory field

ejemplo

CarlosRoca13 (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #335, opened 2024-05-20
[15.0][IMP] dms_field: Add #id from record to auto-generate access group to prevent error

victoralmau:
Add #id from record to auto-generate access group to prevent error

Example use case:
- Create an employee in company A (auto-generated group is created).
- Archive employee
- Create the same employee in company B (auto-generated group is created).

@Tecnativa TT49176

### DIFF
--- a/dms_field/i18n/dms_field.pot
+++ b/dms_field/i18n/dms_field.pot
@@ -6,6 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 15.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2024-05-20 11:48+0000\n"
+"PO-Revision-Date: 2024-05-20 11:48+0000\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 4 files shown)


CarlosRoca13 (approved):

dms_field/i18n/dms_field.pot
```diff
@@ -164,7 +166,6 @@ msgstr ""
 #: model:ir.model.fields,field_description:dms_field.field_decimal_precision__dms_directory_ids
 #: model:ir.model.fields,field_description:dms_field.field_dms_access_group__dms_directory_ids
 #: model:ir.model.fields,field_description:dms_field.field_dms_category__dms_directory_ids
-#: model:ir.model.fields,field_description:dms_field.field_dms_classification_template__dms_directory_ids
```
pedrobaeza: Please remove all these changes from the files, as you locally don't have all the modules installed. Anyway, this makes me wonder if we should have this field populated for all the models:

https://github.com/OCA/dms/blob/d926f50e5e41a848d591cd7b28382da9cef040a1/dms_field/models/base.py#L11

pedrobaeza (approved):
/ocabot merge patch

dms_field/i18n/dms_field.pot
victoralmau: Changes removed. IMO I think that `dms_directory_ids` field should exist in `dms`, but I think it's something that can be discussed (along with other `dms_field` stuff) in the 16.0 migration.

pedrobaeza: OK, I ping @etobella for his knowledge.

etobella (approved):
LGTM Makes complete sense

pedrobaeza:
Well, @etobella my ping was about the off-topic opened here about having `dms_directory_ids` field populated in every model, causing that ugly effect in the translations, instead of having a mixin.

PR #338, opened 2024-05-29
[16.0][FIX] dms: Remove incorrect character in maximum upload size error message

### LINKED ISSUE #337: The error message is not exactly explained when one try to upload the files exceed the file size limit.
The name of the module that has a bug.
dms

The error message is not exactly explained when one try to upload the files exceed the file size limit.

**Affected versions**:
V16

1. Set the file size limit to 1 as 1 MB
2. Create New file record for file bigger than 1 MB.
3. There is error message as expected : "The maximum upload size is 1 MB)." But there is unnecessary character ")" there.
4. Upload the file by clicking the Upload button
5. There is message error: "A file with the same name already exists."  I think we should have the same error message as point 3.

victoralmau:
Remove incorrect character in maximum upload size error message

Related to #https://github.com/OCA/dms/issues/337

@Tecnativa

### DIFF
--- a/dms/i18n/de.po
+++ b/dms/i18n/de.po
@@ -2122,7 +2122,7 @@ msgstr "Der Name der Datei ist ungültig."
 #. odoo-python
 #: code:addons/dms/models/dms_file.py:0
 #, python-format
-msgid "The maximum upload size is %s MB)."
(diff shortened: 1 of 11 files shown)


pedrobaeza (approved):
/ocabot merge patch

PR #339, opened 2024-05-31
[16.0][MIG] dms_field: Migration to 16.0

CarlosRoca13:
cc @Tecnativa TT48507

- [x] This migration has renamed the dms_tree view to dms_list due to an incompatibility with the tree view.

- [x] Adapt code to owl and use reactivity that it offers

- [x] Use new drag and drop as it is used on dms module

ping @victoralmau @pedrobaeza


CarlosRoca13:
Set as draft because drag and drop is not already implemented, but the module is usable yet

dms_field/views/dms_field_template_views.xml
```diff
@@ -0,0 +1,65 @@
...
+                        />
+                        <field name="group_ids" widget="many2many_tags" />
+                    </group>
+                    <notebook position="inside">
+                        <page
+                            name="dms"
+                            string="Documents"
+                            attrs="{'invisible':['|', ('storage_id', '=', False), ('model_id', '=', False)]}"
+                        >
+                            <field name="dms_directory_ids" mode="dms_list" />
```
victoralmau:
```suggestion
                            <field name="dms_directory_ids" mode="dms_list" attrs="{'invisible': [('id', '=', False)]}" />
```

dms_field/models/dms_field_template.py
```diff
@@ -0,0 +1,159 @@
...
+            "is_root_directory": directory.is_root_directory,
+            "name": directory_name,
+            "group_ids": [(4, group.id) for group in groups],
+        }
+
+    @api.constrains("model_id")
+    def _check_model_id(self):
+        for template in self:
+            if self.env["dms.field.template"].search(
+                [("model_id", "=", template.model_id.id)]
```
victoralmau:
```suggestion
                [("model_id", "=", template.model_id.id), ("id", "!=", template.id)]
```

victoralmau (changes requested):

CarlosRoca13:
@victoralmau Changes done and drag & drop added

victoralmau (approved):

dms_field/tests/test_dms_field.py
```diff
@@ -0,0 +1,169 @@
...
+            }
+        )
+        cls.template.group_ids.group_ids = [(4, cls.group.id)]
+        cls.template.group_ids.explicit_user_ids = [(4, cls.user_b.id)]
+        cls.storage = cls.template.storage_id
+        template_ctx = cls.template.with_context(
+            res_model=cls.template._name, res_id=cls.template.id
+        )
+        template_ctx.create_dms_directory()
+        cls.template.refresh()
```
```suggestion
        cls.template.invalidate_model()
```

victoralmau (changes requested):
The same suggestion for all occurrences.

CarlosRoca13:
@victoralmau Changes done!

Added migration script to change the type of ir.ui.view records that keeps dms_tree type

victoralmau (approved):

dms_field/models/base.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2020 Creu Blanca
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
```
pedrobaeza: I think we should change this to a mixin, and on each subsequent module, inherit from that mixin. This way, we avoid an extra field `dms_directory_ids` on each of the existing models in our DB, and to have an overhead on each of the `create` methods. Do you agree @victoralmau @etobella ?

etobella: I do agree. It would be cleaner

victoralmau: I do agree too.

CarlosRoca13:
Done! :smile:

pedrobaeza (commented):
I don't understand why renaming from `dms_tree` to `dms_list` and causes potential problems.

Another thing is that the README is not explaining how to enable this in any model.

CarlosRoca13:
I have added the explanation for the change from dms_tree to dms_list in the commit message and added in CONFIGURE.rst the way to use the embedded view in any model

pedrobaeza:
I'm testing the module and the eye button (or the button "Open" in the right part) is not working properly: for PDF files, it downloads it instead of previewing it, and for other files without preview feature, it gets a black popup:

Peek 2024-06-19 08-54

CarlosRoca13:
So I'm seeing the pdf ok..
seeing_pdf

For other files it is how it is working now the dms, the preview is delegated to the navigator if you click on the preview image of "other files" you will have the same result
dms preview

pedrobaeza:
Please use Firefox for PDF preview. It's not working well there. About the other, I think it the button can be faded if not available.

CarlosRoca13:
It is something about your navigator... 

Test on firefox:
pdf

pedrobaeza (approved):
OK then...

/ocabot merge nobump

PR #340, opened 2024-06-03
[16.0][MIG] hr_dms_field: Migration to 16.0

victoralmau:
Migration to 16.0

Locked by:
- [x] `dms_field`: https://github.com/OCA/dms/pull/339

@Tecnativa TT48870


victoralmau:
It's ready to review

dms_field/tests/test_dms_field.py
```diff
@@ -155,7 +155,7 @@ def test_creation_process_01(self):
         self.assertEqual(total, 1)
         # Create directory again (access groups are the same)
         template.create_dms_directory()
-        self.partner.refresh()
+        self.partner.invalidate_model()
```
pedrobaeza, line 158: Why do we need to manually refresh? Isn't this got by depends?

hr_dms_field/static/src/js/base/dms_list_controller.esm.js
```diff
@@ -0,0 +1,13 @@
+/** @odoo-module **/
+
+import {DMSListControllerObject} from "@dms_field/views/dms_list/dms_list_controller.esm";
+import {patch} from "@web/core/utils/patch";
+
+patch(DMSListControllerObject, "hr_dms_field.DMSListControllerObject", {
```
pedrobaeza, line 6: Can you clarify with a code comment why is this needed?

hr_dms_field/tests/test_hr_dms_field.py
```diff
@@ -45,7 +37,7 @@ def test_employee_exists(self):
 
     def test_employee_create_01(self):
         employee = self.employee_model.create({"name": "Test employee"})
-        employee.refresh()
+        employee.invalidate_model()
```
pedrobaeza, line 42: The same about the need of invalidating.

hr_dms_field/static/src/js/base/dms_list_controller.esm.js
victoralmau: Comment added.

dms_field/tests/test_dms_field.py
victoralmau: Sorry, which depends? I don't understand what you want to do here, this is necessary, it's only a test, it's like this in v15 and nothing has been commented on the migration of `dms_field` to v16.....until now.

pedrobaeza: I know that it's the same as in v15, but you need to do an invalidation because data is not refreshed, and it's usually not refreshed because a computed field doesn't have proper dependencies, and thus, you need to manually trigger the recomputation. For what is this needed?

victoralmau: There is no compute field (and therefore no depends) https://github.com/OCA/dms/blob/16.0/dms_field/models/dms_field_mixin.py#L12

pedrobaeza: So there's no need of `invalidate_model`?

victoralmau: Ok, deleted. According to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0 you should change the `refresh()` method to `invalidate_model()` (a warning log also indicates this).

pedrobaeza: Yeah, but seeing the change, I wonder why it's needed. Let's see if this affects in something in the tests. If so, then we would document at least why it's needed, or maybe fix the part why this is forced.

pedrobaeza: It seems that the one2many that is fed into the create method is not accessible if you don't "refresh". That seems very weird. Is it happening the same on UI? As said, at least we can put a comment saying the reason for refreshing.

victoralmau: In UX it is refreshed when the directory is created, https://github.com/OCA/dms/blob/16.0/dms_field/static/src/views/dms_list/dms_list_renderer.esm.js#L450.

Comments added in tests, is it correct now?

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #342, opened 2024-06-04
[IMP] dms: Capture correct error to be shown at notification

CarlosRoca13:
Until now, when there was an exception when adding a file with drag and drop, the same message was displayed regardless of the exception thrown.

With this changes we get the message of the exception to show the correct message.

Steps to reproduce the problem:
1. Set system parameter dms.binary_max_size to 1
2. Drag a file larger than 1 MB into any folder

You will see the error "A file with the same name already exists" instead of "The maximum upload size is 1 MB."

cc @Tecnativa TT49411

ping @chienandalu @victoralmau


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #345, opened 2024-06-18
[15.0][FIX] dms: Allow share (directories and files) without 'Contact creation' permission

### LINKED ISSUE #343: Can't see the share button in file or directory
## Module
The name of the module that has a bug.

dms

## Describe the bug
The Share button is not visible in the file or directory for DMS version 16, eventhough I have add the portal user in the DMS user group

## To Reproduce
**Affected versions**:
V16.

1. Click the Directory Documents
2. Add Portal User in the groups
3. Refresh the browser
4.  Cick the related files inside the Documents Directory
5.  I can't see the Share button, only see the Lock Button.

**Expected behavior**
The Share button is suppposed to be shown in the Files or Directory Form

**Additional context**
Test in local, because the runboat is not run well.
The share button is shown in v14, but not in V16. 
I haven't tried this on V15 yet.

Thanks.

victoralmau:
Allow share (directories and files) without 'Contact creation' permission

Extra changes: Replace Share buttons with actions

Steps to reproduce:
 - Go to Settings > Users & Companies > Users and create a user only with Documents > Manager permission. (User must not have the Extra Rights > Contact Creation permission).
 - Go to Documents and go into a directory form view and click on "Share" button.
 - An access error is displayed

Fixes https://github.com/OCA/dms/issues/343

@Tecnativa

### DIFF
--- a/dms/__manifest__.py
+++ b/dms/__manifest__.py
@@ -31,6 +31,7 @@
         "views/res_config_settings.xml",
         "views/dms_portal_templates.xml",
         "wizards/wizard_dms_file_move_views.xml",
+        "wizards/wizard_dms_share_views.xml",
     ],
     "assets": {
         "web.assets_backend": [
--- a/dms/security/ir.model.access.csv
+++ b/dms/security/ir.model.access.csv
@@ -24,3 +24,4 @@ access_security_access_groups_user,access_security_access_groups_user,model_dms_
 access_security_access_groups_dms_user,access_security_access_groups_dms_user,model_dms_access_group,group_dms_user,1,1,1,1
 
 access_wizard_dms_file_move,access_wizard_dms_file_move,model_wizard_dms_file_move,group_dms_user,1,1,1,1
+access_wizard_dms_share,access_wizard_dms_share,model_wizard_dms_share,group_dms_manager,1,1,1,0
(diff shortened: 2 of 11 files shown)


pedrobaeza:
Does the standard brings record rules for this wizard? If not, you are restricting them here.

victoralmau:
> Does the standard brings record rules for this wizard? If not, you are restricting them here.

Sorry, you are right, standard does not add rules to the wizard, I have removed them.

pedrobaeza:
I don't feel this as right, as now you are giving too many permissions to share any document. I think we should simply accept that the "Contact creation" permission is needed for this. We can hide the buttons with such group though, and discuss with Odoo for changing such permission upstream.

victoralmau:
We did something similar in https://github.com/odoo/odoo/pull/162199 but finally odoo in v16 has made it "_different_" (creating a specific _project.share.wizard_), it is an approach we could do here too.

pedrobaeza:
OK, that sounds OK, inheriting the generic one and converting it into a specific one for this model.

victoralmau:
Changes done.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #346, opened 2024-06-19
[15.0][FIX] hr_dms_field: Apply correct condition + apply .sudo() to prevent ACL error from dms.access.group

victoralmau:
Apply correct condition + apply .`sudo()` to prevent ACL error from `dms.access.group`

@Tecnativa TT49683


pedrobaeza (approved):
Fast-tracking these obvious patches generated in the previous PR:

/ocabot merge patch

PR #349, opened 2024-07-02
[ADD] dms_partner_attachment_link: add module

dmitriypaulov:
This module adds "DMS Attachments" smart button to the partner's form leading to a set of related, stored as attachments DMS files


pedrobaeza:
Which is the difference with https://github.com/OCA/dms/tree/15.0/dms_attachment_link ?

pedrobaeza:
And why not using `dms_field` for this?

dmitriypaulov:
> Which is the difference with https://github.com/OCA/dms/tree/15.0/dms_attachment_link ?

The difference is that dms_attachment_link adds a wizard for files quick creation from records of specific models. This module just adds a shortcut to see partner related files. I hope that is clear, please let me know if I'm wrong

pedrobaeza:
Well, not really clear the usefulness of this, but it's OK, don't worry.

dms_partner_attachment_link/models/res_partner.py
```diff
@@ -0,0 +1,48 @@
...
+        :rtype: dict
+        """
+        self.ensure_one()
+        action = self.env.ref("dms.action_dms_file").read()[0]
+        action.update(
+            {
+                "domain": [
+                    ("res_id", "=", self.id),
+                    ("res_model", "=", "res.partner"),
+                    ("storage_id.save_type", "=", "attachment"),
```
ivs-cetmix: What is this `storage_id` domain condition for?
Shouldn't all the files be shown no matter where they are stored?

ivs-cetmix (changes requested):
Rather a question yet...

dms_partner_attachment_link/models/res_partner.py
dmitriypaulov: From the task description: "Keep in mind that we can find related partners only for files that are stored in the Attachment Storages"

dmitriypaulov: @ivs-cetmix see also readme/DESCRIPTION.md file

victoralmau (commented):
Although I understand the need for this module, doing so will always make a new one necessary for almost any model and I think that is not a good approach.

My approach to this would be more transversal (maybe add it even in the `dms` module), add an action automatically in the form view of a record (only if there is some document: folder or file) called "View Documens" (or similar) and that when clicked we go to the "Files" action (kanban view) with the folder of the record selected. what do you think?

PR #350, opened 2024-07-22
[15.0][IMP+FIX] dms: Improve performance + Avoid non-existing record access error

victoralmau:
Improve performance + Avoid non-existing record access error

Changes done:
- [x] Improve `_dms_operations()` method to do nothing if the model is not one of those used in `dms.storage`.
- [x] Avoid access error with `.browse()` if we access a non-existing record (for example, deleted by database).

@Tecnativa TT50231

### DIFF
--- a/dms/models/dms_security_mixin.py
+++ b/dms/models/dms_security_mixin.py
@@ -132,8 +132,11 @@ class DmsSecurityMixin(models.AbstractModel):
                 continue
             domains.append([("res_model", "=", model._name), ("res_id", "=", False)])
             # Check record access in batch too
-            group_ids = [i for i in group["res_id"] if i]  # Hack to remove None res_id
-            related_ok = model.browse(group_ids)._filter_access_rules_python(operation)
+            res_ids = [i for i in group["res_id"] if i]  # Hack to remove None res_id
+            # Apply exists to skip records that do not exist. (e.g. a res.partner deleted
+            # by database).
+            model_records = model.browse(res_ids).exists()
+            related_ok = model_records._filter_access_rules_python(operation)
             if not related_ok:
                 continue
             domains.append(
--- a/dms/models/ir_attachment.py
+++ b/dms/models/ir_attachment.py
@@ -1,6 +1,7 @@
 # Copyright 2021 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 from odoo import api, models
+from odoo.tools import ormcache
 
 
 class IrAttachment(models.Model):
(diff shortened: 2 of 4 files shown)


dms/models/dms_security_mixin.py
```diff
@@ -133,7 +133,12 @@ def _get_domain_by_inheritance(self, operation):
             domains.append([("res_model", "=", model._name), ("res_id", "=", False)])
             # Check record access in batch too
             group_ids = [i for i in group["res_id"] if i]  # Hack to remove None res_id
-            related_ok = model.browse(group_ids)._filter_access_rules_python(operation)
+            # Do a search instead of browse() in case a record does not exist (e.g. a
```
pedrobaeza: If the problem is not existing groups, then you can do `model.browse(group_ids).exists()`, but are you sure this is the problem? The problem was on attachments with an unexisting `res_model` or `res_id` target.

victoralmau: The problem is that in some cases the linked directories or files may not have been deleted (e.g. by deleting with SQL in migration scripts), something that is covered in https://github.com/OCA/dms/blob/15.0/dms/models/base.py#L11.

pedrobaeza: But I insist, this one is not preventing that case.

victoralmau: This change prevents an error from being displayed when trying to access a record that does not exist (the initial use case), and the test verifies it. If you have any other suggestion, it is welcome.

pedrobaeza: OK, part of my non-understanding is the bad variable name `group_ids`. Please call it `referred_ids` or similar. After that, my initial suggestion `model.browse(referred_ids).exists()` is more compact and may preserve caches.

victoralmau: Changes done.

dms/models/ir_attachment.py
```diff
@@ -36,8 +36,21 @@ def _dms_directories_create(self):
             )
 
     def _dms_operations(self):
+        """Perform the operation only if there is a storage with linked models.
+        The directory (dms.directory) linked to the record (if it does not exist)
+        and the file (dms.file) with the linked attachment would be created.
+        """
+        # Apply sudo to prevent ir.rule from being applied.
+        storages = self.env["dms.storage"].sudo().search([("model_ids", "!=", False)])
+        dms_models = storages.mapped("model_ids.model")
```
pedrobaeza: If we continue this way, this must be cached and invalidate when changes. But I don't ge why `res_model` and `res_id`, which is a core mechanism, is filled for the DMS, and we may need to avoid it and use other fields.

pedrobaeza (approved):
/ocabot merge patch

PR #353, opened 2024-07-26
[16.0][FIX] dms_field: Incompatibility with web_responsive

CarlosRoca13:
When we edit the name of a file or directory while positioned in a field, like the one added by hr_dms_field with web_responsive installed, the input text appears white, making it impossible to edit.
white

With these changes, the text appears black and can be edited normally.
black

cc @Tecnativa TT50329

ping @victoralmau @pedrobaeza


victoralmau (approved):
Functional review ok

pedrobaeza (approved):
Not very friend of `!important`, but it's a specific case that can't be solved other way.

/ocabot merge patch

PR #363, opened 2024-09-30
[16.0][IMP] dms_auto_classification + dms_field_auto_classification: Remove warning by several names with the same label

victoralmau:
Remove warning by several names with the same label

`WARNING devel odoo.addons.base.models.ir_model: Two fields (file_id, data_file) of wizard.dms.classification.detail() have the same label: File. [Modules: dms_auto_classification and dms_auto_classification]`

`WARNING devel odoo.addons.base.models.ir_model: Two fields (model, model_id) of dms.classification.template() have the same label: Model. [Modules: dms_field_auto_classification and dms_field_auto_classification]`

@Tecnativa TT51051


pedrobaeza (approved):
Fast-tracking it as obvious:

/ocabot merge patch

PR #365, opened 2024-10-01
feat(dms): add file content search to community version [Silvertouch …

dakshesh-sttl:
…Technologies]


pedrobaeza:
Please don't open the same proposal in all the versions until one is approved, and more if we have already said that this is not the way to go. Closing thus.

PR #366, opened 2024-10-01
feat(dms): add file content search to community version [Silvertouch …

dakshesh-sttl:
…Technologies]

### DIFF
--- a/dms/__manifest__.py
+++ b/dms/__manifest__.py
@@ -1,3 +1,4 @@
+# Copyright 2024 Silvertouch technologies
 # Copyright 2017-2019 MuK IT GmbH
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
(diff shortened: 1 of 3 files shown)


victoralmau (changes requested):
The name of the PR is wrong, remember the [guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst), it should be: _[16.0][IMP] dms: Add indexed content to files_ (or similar)

Considering `base` does with `ir.attachment` (https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_attachment.py#L355) and the `attachment_indexation` module I think the following should be done (if possible):

- The `index_content` field should be a compute store and set the value in a “`similar`” way as it is done in base.
- Create a module called `dms_indexation` that depends on `attachment_indexation` and `dms` and “extend” the dms method to contemplate more formats.

What do you think @pedrobaeza ?

I think that creating PRs for all branches is not a good technique, because if you have to change something you will have to change it in all of them, the right thing to do is to create one and when it is merged take it to higher versions (one by one).

pedrobaeza:
Yes, please put this on a separate module for being optional.

PR #373, opened 2024-10-09
[17.0][MIG] dms_attachment_link: Migration to 17.0

victoralmau:
Migration to 17.0

Please @CarlosRoca13 and @pilarvargas-tecnativa can you review it?

@Tecnativa TT50055


dms_attachment_link/static/src/xml/chatter.xml
```diff
@@ -1,16 +1,13 @@
...
-            <attribute name="class" add="d-block text-center" separator=" " />
-        </xpath>
-        <xpath expr="//button[hasclass('o_AttachmentBox_buttonAdd')]" position="after">
+    <t t-inherit="mail.Chatter" t-inherit-mode="extension">
+        <xpath expr="//FileUploader//button" position="after">
             <button
                 class="o_add_dms_file_button btn btn-link"
                 type="button"
-                t-on-click.stop.prevent="attachmentBoxView._onAddDmsFile"
+                t-on-click="_onAddDmsFile"
```
chienandalu: Why not using the `stop.prevent` API here. It's aimed to avoid boiler plate code in the js logic

dms/views/dms_file.xml
```diff
@@ -281,7 +281,7 @@
         <field name="arch" type="xml">
             <kanban position="attributes">
                 <attribute name="banner_route" />
-                <attribute name="class">o_kanban_mobile</attribute>
+                <attribute name="class">o_kanban_mobile mk_file_kanban_view</attribute>
```
pilarvargas-tecnativa, line 284: Shouldn't it go in a separate commit if necessary for migration?

pilarvargas-tecnativa:
Following the instructions in the readme, the behaviour is not the same as in previous versions. For the ‘Add DMS files’ option to appear, a file must be uploaded. In previous versions, clicking on the chatter clip already showed this option.
dms

victoralmau:
Ok, I understand what you are saying, but in v17 that is the default behavior (https://github.com/odoo/odoo/blob/17.0/addons/mail/static/src/core/web/chatter.js). Should we change that behavior in this module? I think not, it should be done in another one. 

Another module (`document_url`) does something similar https://github.com/OCA/knowledge/blob/17.0/document_url/static/src/js/url.esm.js#L38 (although I don't like the replace that is defined.)

What do you think @pedrobaeza

pedrobaeza:
We need to possibility of adding a DMS file although no current attachment, so yes, we should add that option in this module, similar to `document_url`.

pilarvargas-tecnativa (approved):
LGTM thanks!

pedrobaeza (approved):
/ocabot merge patch

PR #374, opened 2024-10-10
[17.0][IMP] dms: Change portal_my_home_dms to use its own div

victoralmau:
Change portal_my_home_dms to use its own div (https://github.com/odoo/odoo/blob/17.0/addons/sale/views/sale_portal_templates.xml)

Please @chienandalu can you review it?

@Tecnativa

### DIFF
--- a/dms/template/portal.xml
+++ b/dms/template/portal.xml
@@ -10,17 +10,19 @@
         inherit_id="portal.portal_my_home"
         priority="20"
     >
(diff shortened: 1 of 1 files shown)


flotho:
any chance to have a screen before/after ?
regards

victoralmau:
In this case there is no difference between before and now, just change the view definition to a “_correct_” way (at least IMO).

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #375, opened 2024-10-16
[16.0][ADD] sale_dms_field

agent-z28:


agent-z28:
@victoralmau 
@bizzappdev

victoralmau (changes requested):
Code and functional review OK.

Some comments:
- According to the [Guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst) change the PR title to `[16.0][ADD] sale_dms_field: New module` and the commit message to: `[ADD] sale_dms_field: New module`.
- If in sales we set sequences with dates and we use for example `S/%(year)s `we will have an error similar to `account_dms_field`. I think that the most appropriate (in another PR) is to move https://github.com/OCA/dms/blob/16.0/account_dms_field/models/dms_field_template.py#L11 to `dms_field` (it is only necessary to replace **/** to **-** and then it can be removed from `account_dms_field` and it will work for this module (`sale_dms_field`) or any other.

agent-z28:
@victoralmau i did the naming changes.
We can fix the mentioned issue in another  PR ?

victoralmau:
The suggested change can be done in a new PR (it will be merged before this) or in this PR, but if it is done in this PR it should be the 1st commit (with a different message) and finally  `sale_dms_field`

agent-z28:
> The suggested change can be done in a new PR (it will be merged before this) or in this PR, but if it is done in this PR it should be the 1st commit (with a different message) and finally `sale_dms_field`

@victoralmau okay we will take care next week

agent-z28:
@victoralmau https://github.com/OCA/dms/pull/446

agent-z28:
@victoralmau and then this https://github.com/OCA/dms/pull/445

PR #378, opened 2024-11-19
[16.0][IMP] dms_field: Add partner embedded DMS template

### LINKED ISSUE #377: no conten at contact >> dms tab
No content at contact >> dms tab

## Module
dms version 16

## Describe the bug
if dms modules are installed - no content is visible at contact >> dms tab
see: https://prnt.sc/Xe6irbDQe5oR
what is the idea?

## To Reproduce
testet at version 16 runboat http://oca-dms-16-0-pr375-255a821dce34.runboat.odoo-community.org

1. menu contact
2. tab DMS
3. no content see: https://prnt.sc/Xe6irbDQe5oR

**Expected behavior**
what is the idea here ?

victoralmau:
Add partner embedded DMS template

Fixes #377

@Tecnativa

### DIFF
--- a/dms_field/tests/test_dms_field.py
+++ b/dms_field/tests/test_dms_field.py
@@ -27,23 +27,10 @@ class TestDmsField(TransactionCase):
             {"name": "Test group", "users": [(4, cls.user_a.id)]}
         )
         cls.user_b = new_test_user(cls.env, login="test-user-b")
-        cls.template = cls.env["dms.field.template"].create(
-            {
-                "name": "Partner",
-                "storage_id": cls.env.ref("dms.storage_demo").id,
-                "model_id": cls.env.ref("base.model_res_partner").id,
-                "group_ids": [(4, cls.env.ref("dms.access_group_01_demo").id)],
-                "directory_format_name": "{{object.display_name}}",
-            }
-        )
+        cls.template = cls.env.ref("dms_field.field_template_partner")
         cls.template.group_ids.group_ids = [(4, cls.group.id)]
         cls.template.group_ids.explicit_user_ids = [(4, cls.user_b.id)]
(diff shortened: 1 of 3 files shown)


pedrobaeza:
We should avoid using demo data on tests.

victoralmau:
> We should avoid using demo data on tests.

I don't totally agree, but ok, it's already changed.

CarlosRoca13 (approved):

victoralmau:
Ping @pedrobaeza

dms_field/tests/test_dms_field.py
```diff
@@ -27,6 +27,7 @@ def setUpClass(cls):
             {"name": "Test group", "users": [(4, cls.user_a.id)]}
         )
         cls.user_b = new_test_user(cls.env, login="test-user-b")
+        cls.env.ref("dms_field.field_template_partner").unlink()
```
pedrobaeza: Why do you need to remove it? And in any case, do it safely passing `False` as second argument, because they may not exist on the current DB.

pedrobaeza (changes requested):
> > We should avoid using demo data on tests.
> 
> I don't totally agree, but ok, it's already changed.

About this, the reasons are:

- Being self-contained: any change on data across versions/updates that affect the tests are not contained on the test file itself, and they are more difficult to be traced.
- Being reliable: if you depend on demo data, that demo data may be manually altered on the current DB, making the test to fail.

Thus, we must avoid it, the same as finally Odoo has done in its own tests.

In this module, I also see more dependencies to demo data (like `dms.access_group_01_demo`, `dms.storage_demo`, etc), so they must be removed.

pedrobaeza (approved):
/ocabot merge patch

PR #379, opened 2024-11-29
[16.0][FIX] dms_field: Test compatibility

victoralmau:
Test compatibility

We need to avoid applying a template except when testing functionality with `dms_field*` modules to avoid the error that a directory with the same name already exists (example: create partner).

Related to https://github.com/OCA/dms/pull/378

@Tecnativa


CarlosRoca13 (approved):

pedrobaeza:
Shouldn't the dependent modules inherit from a base test in dms_field?

victoralmau:
I have thought about it, but the base test of dms_field would have only the context key set, I think the cost/benefit of having it is not significant.

pedrobaeza (approved):
/ocabot merge patch

PR #389, opened 2025-03-24
[17.0][IMP] dms: Implement index_content and create dms_indexation mo…

dakshesh-sttl:
…dule


victoralmau (changes requested):
Same comments as indicated in https://github.com/OCA/dms/pull/366#pullrequestreview-2339403212

victoralmau (changes requested):
Some comments:

- Check pre-commit errors.
- Modify the PR title, e.g.: _[17.0][ADD] dms_indexation: New module_
- Unify all changes in a single commit (_[17.0][ADD] dms_indexation: New module for example_).
- Modify the module icon with the existing dms icon + attachment indexation combination (or if you prefer, you can use the one for default).
- You can add files to the readme folder to describe the module and how to use it.
- You are adding more stuff than necessary (the `attachment_indexation` module already takes care of that), I think the method should be very similar to https://github.com/odoo/odoo/blob/583e785fc6e2acababa5865a6e0e8e71f3df8a49/odoo/addons/base/models/ir_attachment.py#L384.

PR #395, opened 2025-03-26
[16.0][ADD] dms_account_avatax_exemption

kobros-tech:
This module connects between account avatax exemption dms modules.


kobros-tech:
Screenshot from 2025-03-27 00-01-59
Screenshot from 2025-03-27 00-03-46
Screenshot from 2025-03-27 00-02-18

kobros-tech:
depends on:
https://github.com/OCA/account-fiscal-rule/pull/512

kobros-tech:
### Adding an exemption file will create an exemption line for the related partner:

Screenshot from 2025-03-28 09-05-18

Screenshot from 2025-03-28 09-05-44

kobros-tech:
@pedrobaeza

pedrobaeza:
I think you can use https://github.com/OCA/dms/tree/16.0/dms_attachment_link as a general wildcard to select attachments from DMS.

kobros-tech:
@pedrobaeza 

I reviewed the module functionality only not code that you told me about.

but this one will create one attachment with two dms.file records for for res.partner model and one for their exemption model.


In fact I am thinking about more further improvement by asking the partner on which type this file for like sign doc, exemption doc, offer, etc?

this will help us better to put the file in its best place.

it can be a JS wizard to select from combo box or check box, 

but this case will need me to create a new module for the type of attachment we upload, what do you think?

pedrobaeza:
If this module continues as is, it should be hosted not here, but where `account_avatax*` modules lives (I think OCA/l10n-usa).

kobros-tech:
transeferred to https://github.com/OCA/l10n-usa/pull/156

Issue #390, opened 2025-03-25
[16.0] dms: Error when using the upload file button


adasatorres:
When uploading a file using the upload button from a folder that has no files, I get the following error: "You must select a directory first."

**Affected versions**:

-  16.0

1. Enter any directory that has no files from the smart button in the kanban or form view.
2. I tried to upload a file using the upload button next to the new button.
3. When you upload the file, it will start loading, but a popup will appear in the top right corner saying that you must select the directory first.

If you repeat the same steps in a folder where files exist, the file will be uploaded successfully.

Based on my observations during testing, the error may be related to the fact that when there are no files, the sidebar information does not exist.

I have used the following instance of the runboat to perform the tests.
http://oca-dms-16-0-1e313944c756.runboat.odoo-community.org/web

I attached a recording of the reproduction of the problem.

https://github.com/user-attachments/assets/b7e7bdb4-e598-43f9-9aca-e83c5179e45f

fariageenBH:
A little help form a friend of mine 😉: "Examine How the Directory is Determined: Inspect the block of code where the error is raised. The issue likely arises because when the folder is empty, the sidebar (or DOM element) that normally holds the directory information is missing, so selectedDirectory ends up being null or undefined."

PR #396, opened 2025-03-28
[16.0][FIX]dms: The method for uploading a file did not work when the folder was empty.

adasatorres:
Good morning, here is a possible solution for issue #390 

The solution I propose is to use the `default_directory_id` from the context of the current controller, as the directory is currently specified within a domain. If there is no domain, a file cannot be uploaded, but if there are no files, there is no domain either.

### DIFF
--- a/dms/static/src/js/views/dms_file_upload.esm.js
+++ b/dms/static/src/js/views/dms_file_upload.esm.js
@@ -102,16 +102,18 @@ export const FileUpload = {
     async onUpload(attachments) {
         const self = this;
         const attachmentIds = attachments.map((a) => a.id);
-        const ctx = this.props.context;
+        const ctx = Object.assign(
+            {},
+            this.actionService.currentController.props.context,
+            this.props.context
+        );
         const controllerID = this.actionService.currentController.jsId;
 
         if (!attachmentIds.length) {
             this.notification.add(_t("An error occurred during the upload"));
             return;
         }
-
         // Search the correct directory_id value according to the domain
-        ctx.default_directory_id = false;
         if (this.props.domain) {
             for (const domain_item of this.props.domain) {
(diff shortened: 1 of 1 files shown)


AlvaroRM11 (approved):
LGTM

rrebollo (approved):
Code Review. LGTM! Amazing. I think this happens in v15 too.

dms/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Document Management System",
     "summary": """Document Management System for Odoo""",
-    "version": "16.0.1.8.3",
+    "version": "16.0.1.8.4",
```
pedrobaeza: Don't change manually module version unless you include a migration script. It will be bumped by the bot.

dms/static/src/js/views/dms_file_upload.esm.js
```diff
@@ -102,16 +102,18 @@ export const FileUpload = {
...
+        );
         const controllerID = this.actionService.currentController.jsId;
 
         if (!attachmentIds.length) {
             this.notification.add(_t("An error occurred during the upload"));
             return;
         }
-
         // Search the correct directory_id value according to the domain
-        ctx.default_directory_id = false;
```
pedrobaeza, line 114: Isn't this causing side effects in any place?

adasatorres: Good evening @pedrobaeza , I don't think that line is necessary, since if the default_directory_id attribute doesn't exist, it should return undefined, and since that's a falsy value, it shouldn't cause any issues.

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to 17

Issue #381, opened 2024-12-09
Locked attachments can still be deleted from the chatter of an object


lorenzomorandini:
## Module
DMS

## Describe the bug
Locking an attachment of an object (example an attachment of an invoice) thru DMS still allows for the attachment to be deleted from the chatter of the invoice.

**Expected behavior**
Deleting / modifying the attachment from the invoice should be blocked as well.

Tested on Odoo 16.

kobros-tech:
could you review this PR #385 to see if the issue is solved or not?

victoralmau:
According to the defined rule https://github.com/OCA/dms/blob/16.0/dms/security/security.xml#L63 users will be able to delete files if they have locked them themselves (regardless of how they try to delete them).

pedrobaeza:
Closing this as no issue then.

lorenzomorandini:
Sorry, my bad for not being clear enough. The problem is that:
- User A blocks a file
- User B can not delete it thru DMS but can delete it thru attachment

User A is an administrator, user B is a normal DMS user.

pedrobaeza:
OK, reopening it.

PR #397, opened 2025-03-28
[16.0][FIX] dms: Delete the file if the attachment is deleted

victoralmau:
Delete the file if the attachment is deleted

When deleting an attachment, an attempt should be made to delete the linked file, causing an error if it is locked. Do not show the Delete/Archive/Unarchive option on files if it is locked and you cannot unlock it.

Fixes https://github.com/OCA/dms/issues/381

@Tecnativa

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -202,6 +202,7 @@ class File(models.Model):
         prefetch=False,
         invisible=True,
         ondelete="cascade",
+        index=True,
     )
 
     def get_human_size(self):
@@ -618,6 +619,13 @@ class File(models.Model):
                 vals = self._create_model_attachment(vals)
             new_vals_list.append(vals)
         return super(File, self).create(new_vals_list)
+
+    def unlink(self):
+        attachments = self.mapped("attachment_id")
+        res = super().unlink()
+        if not self.env.context.get("dms_file"):
+            attachments.with_context(dms_file=True).unlink()
+        return res
 
     # ----------------------------------------------------------
     # Locking fields and functions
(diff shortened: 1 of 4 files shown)


victoralmau:
Ping @lorenzomorandini Could you review it?

pedrobaeza (commented):
Why calling `_dms_operations`?

victoralmau:
> Why calling `_dms_operations`?

It is to avoid creating a new method because this condition https://github.com/OCA/dms/blob/0aa7e6b7f7246278a4fd122e0c38fc4301f429a6/dms/models/ir_attachment.py#L51 must be done and I don't want to repeat code. Abstracting that part in another method is also not necessary and would make it more difficult to understand the change and a bigger diff.

pedrobaeza:
I don't think you need to do that. Just do `self.env["dms.file"].search([("attachment_id", "=", attachment.id)]).unlink()` and it won't find anything if the model is not linked.

victoralmau:
> I don't think you need to do that. Just do `self.env["dms.file"].search([("attachment_id", "=", attachment.id)]).unlink()` and it won't find anything if the model is not linked.

I think it will have performance consequences (do that search always), that's why it was done https://github.com/OCA/dms/blob/0aa7e6b7f7246278a4fd122e0c38fc4301f429a6/dms/models/ir_attachment.py#L40 but ok, I modify it, no problem.

pedrobaeza:
If the m2o field is indexed, it should be better IMO.

lorenzomorandini:
Just tested it, seems to be working, thanks!
I just noticed one thing, kind of related:
- if I delete a file from the chatter it gets deleted from DMS too
- if I delete a file from DMS it does NOT get delete from the chatter

Is this by design?

pedrobaeza (approved):

victoralmau:
> Just tested it, seems to be working, thanks! I just noticed one thing, kind of related:
> 
>     * if I delete a file from the chatter it gets deleted from DMS too
> 
>     * if I delete a file from DMS it does NOT get delete from the chatter
> 
> 
> Is this by design?

The change to remove the linked attachment when deleting a dms file has also been added.

PR #398, opened 2025-03-31
[16.0][IMP] hr_dms_field: Allow to hide the documents by a group (activated by default)

etobella:
This way, we have a better granularity

@victoralmau

### DIFF
--- a/hr_dms_field/security/security.xml
+++ b/hr_dms_field/security/security.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <data>
+        <record id="group_hr_public_dms" model="res.groups">
+            <field name="name">DMS Is accessible publicly on employee profile</field>
+            <field name="category_id" ref="base.module_category_hidden" />
+        </record>
+    </data>
+    <data noupdate="1">
+        <record id="base.group_user" model="res.groups">
+            <field
+                name="implied_ids"
+                eval="[(4, ref('hr_dms_field.group_hr_public_dms'))]"
+            />
+        </record>
+    </data>
+</odoo>
--- a/hr_dms_field/views/hr_employee_view.xml
+++ b/hr_dms_field/views/hr_employee_view.xml
@@ -22,7 +22,11 @@
         <field name="inherit_id" ref="hr.hr_employee_public_view_form" />
         <field name="arch" type="xml">
             <field name="user_id" position="after">
-                <field name="is_logged" invisible="1" />
+                <field
+                    name="is_logged"
+                    invisible="1"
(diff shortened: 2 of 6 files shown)


hr_dms_field/views/hr_employee_view.xml
```diff
@@ -22,14 +22,19 @@
         <field name="inherit_id" ref="hr.hr_employee_public_view_form" />
         <field name="arch" type="xml">
             <field name="user_id" position="after">
-                <field name="is_logged" invisible="1" />
+                <field
+                    name="is_logged"
+                    invisible="1"
+                    groups="hr_dms_field.group_hr_public_dms"
```
victoralmau, line 28: Although it is consistent, it may not be necessary to add `groups` in this field (if for example another module modifies the groups of `page name=“dms”`, it would also need to change the `groups` of this field, which could be confusing).

etobella: I did it this way in order to avoid the computation. I can remove the logic if you prefer

victoralmau: Ok, no problem, you're right, this seems more coherent.

victoralmau (approved):

hr_dms_field/security/security.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <data>
+        <record id="group_hr_public_dms" model="res.groups">
+            <field name="name">DMS Is accessible publicly on employee profile</field>
+            <field name="category_id" ref="base.module_category_hidden" />
+        </record>
+    </data>
+    <data noupdate="1">
+        <record id="base.group_user" model="res.groups">
```
pedrobaeza, line 10: Shouldn't it be HR managers only?

etobella: This group allows to show your dms documents on the hr.employee.public view (only your documents, obviously)

This model is not visible for hr roles (replaced by hr.employee).

However, sometimes, it makes sense to disable this

pedrobaeza: OK, right

pedrobaeza:
Uhm, I'm still not convinced, as for people not wanting to show it, you have to remove first the inheritance on internal user, and then select the field. I think it's better to put a migration script adding the new group to all the existing user + default one, but to not put the inheritance, don't you think?

etobella:
It is modified as a configuration of res.config.settings. It has been activated by default because that was the behavior before

pedrobaeza (approved):
OK, I don't like that way of fine-graining, as it adds everyone or remove all of them checking/unchecking, but it's how it's done in most places, so not blocking more.

/ocabot merge minor

PR #400, opened 2025-04-10
[17.0][FW][FIX] dms: The method for uploading a file did not work when the folde…

yoadria:
…r was empty.
This is a forward-port of PR https://github.com/OCA/dms/pull/396
@BinhexTeam


christian-ramos-tecnativa:
It seems like this is not needed on this version, can you check it?

PR #399, opened 2025-04-02
[18.0][MIG] dms: Migration to 18.0

victoralmau:
Supersedes https://github.com/OCA/dms/pull/385

Changes done:
- [x] Purge translations.
- [x] Squash administrative commits.
- [x] Change the `_check_recursion()` method to `_has_cycle()`
- [x] Change methods `check_access_rights()` + `check_access_rule()` to `check_access()` + adaptations.
- [x] Change `_filter_access_rules()` +`_filter_access_rules_python()` methods to `_filtered_access()` + adaptations.
- [x] Remove `/** @odoo-module **/` from `.js` files.
- [x] Change `tree` to `list`
- [x] Change `t-name=“kanban-box”` to `t-name=“card”`.
- [x] Change `<div class=“oe_chatter”>` to `<chatter />`
- [x] Improvements in `copy()` method
- [x] https://github.com/OCA/dms/pull/385#discussion_r2011498933
- [x] https://github.com/OCA/dms/pull/385#issuecomment-2760582342
- [x] https://github.com/OCA/dms/pull/385#pullrequestreview-2724759632
- [ ] https://github.com/OCA/dms/pull/385#pullrequestreview-2724828958
- [x] https://github.com/OCA/dms/pull/385#issuecomment-2760739152

Migration to 18.0

@Tecnativa TT55505

PR #401, opened 2025-04-11
[IMP] dms: Add functionality to assign linked attachments record ID its value

kobros-tech:
Similar to dms project we can select the model we want and then via a wizard,  we can select the record id/ids of the same kind of the chosen model.

In the followers section I add the record of selected id there `if the model is Partners`

Based on the issue I raised in #399 that the field is readonly and is required on the same time. This commit is solving the problem and making it no problem at all in the UI.

### DIFF
--- a/dms/__manifest__.py
+++ b/dms/__manifest__.py
@@ -40,6 +40,7 @@
         # Wizard
         "wizards/wizard_dms_file_move_views.xml",
         "wizards/wizard_dms_share_views.xml",
+        "wizards/wizard_dms_dir_record_view.xml",
     ],
     "assets": {
         "web.assets_backend": [
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -540,10 +540,6 @@ class DmsDirectory(models.Model):
                 raise ValidationError(
                     _("A directory has to have model in attachment storage.")
                 )
-            if not record.is_root_directory and not record.res_id:
-                raise ValidationError(
-                    _("This directory needs to be associated to a record.")
-                )
 
     @api.constrains("is_root_directory", "storage_id")
     def _check_directory_storage(self):
@@ -767,3 +763,16 @@ class DmsDirectory(models.Model):
             searchpanel_default_directory_id=self.id,
         )
         return action
+
+    def button_add_res_id(self):
+        self.ensure_one()
+        return {
+            # context since 17.0 will be dropped in views
+            # unless we suffix it's key with _view_ref
+            "context": {"directory_id_view_ref": self.id},
+            "name": _("Linked attachments record ID Wizard"),
+            "view_mode": "form",
+            "res_model": "wizard.directory.record",
+            "type": "ir.actions.act_window",
+            "target": "new",
+        }
--- a/dms/security/ir.model.access.csv
+++ b/dms/security/ir.model.access.csv
@@ -25,3 +25,5 @@ access_security_access_groups_dms_user,access_security_access_groups_dms_user,mo
 
 access_wizard_dms_file_move,access_wizard_dms_file_move,model_wizard_dms_file_move,group_dms_user,1,1,1,1
 access_wizard_dms_share,access_wizard_dms_share,model_wizard_dms_share,group_dms_manager,1,1,1,0
+access_wizard_directory_record,access_wizard_directory_record,model_wizard_directory_record,group_dms_user,1,1,1,1
+access_record_placeholder,access_record_placeholder,model_record_placeholder,group_dms_user,1,1,1,1
--- a/dms/views/dms_directory.xml
+++ b/dms/views/dms_directory.xml
@@ -506,10 +506,20 @@
                             required="storage_id_save_type == 'attachment'"
                         />
                         <field name="res_model" invisible="1" force_save="1" />
+                        <button
+                            name="button_add_res_id"
+                            type="object"
+                            string="Load Linked attachments record ID"
+                            class="btn btn-link p-0"
+                            invisible="not model_id or
+                            storage_id_save_type != 'attachment' or
+                            (model_id and count_total_files > 0)"
+                        />
                         <field
                             name="res_id"
                             readonly="True"
                             invisible="storage_id_save_type != 'attachment'"
+                            required="model_id and not is_root_directory"
                         />
                         <field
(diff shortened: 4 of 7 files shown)


kobros-tech:
Screenshot from 2025-04-12 16-55-00

Screenshot from 2025-04-12 16-55-26

Screenshot from 2025-04-12 16-55-37

kobros-tech:
@pedrobaeza 
@victoralmau

kobros-tech:
@rvalyi
@xaviedoanhduy

victoralmau (changes requested):
IMO this is not necessary, it adds code to do something that should not be necessary because it is already done automatically: https://github.com/OCA/dms/issues/404#issuecomment-2798786179

kobros-tech:
> IMO this is not necessary, it adds code to do something that should not be necessary because it is already done automatically: [#404 (comment)](https://github.com/OCA/dms/issues/404#issuecomment-2798786179)

excuse me, how can this be done automatically?
by sending an email for example? or what?

victoralmau:
Just create an attachment linked to a record (`res.partner` for example), the file (`dms.file`) will be created in the corresponding partner directory (the directory will be created automatically if it does not exist).

kobros-tech:
> Just create an attachment linked to a record (`res.partner` for example), the file (`dms.file`) will be created in the corresponding partner directory (the directory will be created automatically if it does not exist).

If the field in the view of dms.file is readonly:

https://github.com/OCA/dms/blob/18.0/dms/views/dms_file.xml#L444

and is defined in the backend as related WITHOUT being editable there:
https://github.com/OCA/dms/blob/18.0/dms/models/dms_file.py#L193

from the code I see it is impossible to assign the res_id a value from the dms.file form view.

kobros-tech:
@victoralmau 
instead of getting encoutagement to have this improvement and become a solution in similar cases, and even better to create a much easier code using OWL to select a record from the model we want, I am getting depressed.

I exerted some effort to understand the security of dms, and still need to understand it more so when I introduce my update it becomes smooth.

but for this PR, I see it very clear that a field is readonly and no way to create it in the GUI.

Shall I close this PR and finish discussion or we could have this new feature here and maybe in similar cases somewhere else?

victoralmau:
I think first of all, is this change necessary? The `res_id` field is readonly in the form view intentionally, according to the explained behavior, the directory and file will create itself. 

What is the use case why you need to create the directory "_manually_" before? That would really be the initial question.

The security issue (there is no problem about it) is “_simple_” for files/directories that are linked to an attachment type storage with the “_Inherit permissions from related record_” checkbox checked, any user (portal or not) will be able to see the files/directories of a record, if they can see that record (`res.partner` or `sale.order` for example).

kobros-tech:
@victoralmau 

usecase is just each customer should have access to exemption documents they sign or upload or send by email or even I add them to their related directories.

here what I like that each partner can have their own directory, but I don't understand how you create a folder for each partner at least this is not explained in the readme.

you and pedro told me once I upload attachment it will be in partner dir., but how you create the dir. without a required field and readonly at the same time.

for the access security, it is fine but I like to add access rules it will minimize the need to create a single group for each single user.


Please write to me steps from scratch to create a folder for Joel portal or any demo partner or record a video and I will continue working on.

I need to have a smooth update and minimize code and make it efficient.

in dms.dir I am blocked and in dms.file and even in the attachment dashboard due to res_id, maybe you have something I did not understand yet.

victoralmau:
I think an example use case with sales orders will help you to understand it.

Storage:
ejemplo-1

Creating several sales orders to different customers (each order has a linked attachment when sent by email).
ejemplo

The customerA portal user will only see your sales orders.
test-customer-a

The customerB portal user will only see your sales orders.
test-customer-b

Does this clarify how it works? I guess with this example you will understand that the directories are created automatically and it is NOT necessary to add access groups or specific rules, if the user can access the linked record (sales order) he will be able to see the directory and related files.

kobros-tech:
@victoralmau 

thank you very much, I guess I will update the readme with a use case for very new users.

I will work on that if I don't need updating or not.

hitrosol:
Thanks for the explanation of using the dms_attachment_link. I just know how to use it :)
But I believe there will be some space for improvement, and having the capability to manage the sub folder below Attachment Storage will be great 👍

victoralmau:
> Thanks for the explanation of using the dms_attachment_link. I just know how to use it :) But I believe there will be some space for improvement, and having the capability to manage the sub folder below Attachment Storage will be great 👍

I think it is important to clarify that the above explanation is **ONLY** of the `dms` module. 

The `dms_attachment_link` module is an extra module that has a “_different_” functionality: Through a wizard it allows you to create an attachment linked to the corresponding element using the selected file as a “template”. An example of this would be to create an attachment in a sales order “installation manual” through that wizard.

kobros-tech:
@victoralmau 

is dms able to include sign requests (signed) docs, or I can add this feature similar to sales orders you showed me?

victoralmau:
The sign request can be linked to any record through the `record_ref` field, but if you consider it necessary, you can create a module similar to [maintenance_sign_oca](https://github.com/OCA/sign/tree/17.0/maintenance_sign_oca) for `dms`.

kobros-tech:
@victoralmau 

I tried to follow but I didn't have the same result, not directories created for sales orders.
I did this work locally without real mailing!

Screenshot from 2025-05-13 17-18-49

Screenshot from 2025-05-13 17-18-42

Screenshot from 2025-05-13 17-18-37

kobros-tech:
whether for Partners dir once I upoaded attachment for admin, I sow it.

This means the dms did not create a root dir for sales orders and did not create subfolders as a result.

In contrast for already created root dirs like Partners subfolders are created as expexted.

Please make sure the function is as expected and introduced so I can find a base to start from.

kobros-tech:
We may write about this in readme so that users can use it easily.

victoralmau:
You have forgotten an important step, it is necessary to create a root folder linked to this storage and to the sale.order model.

ejemplo-1
ejemplo-2

kobros-tech:
@victoralmau 

do you mean I should create the root directory for the specific model like sales orders and for each new attachment in a record another new related dms dir record will be created?

let me try then.

PR #405, opened 2025-04-14
[17.0][FIX] dms: Delete the file if the attachment is deleted

### LINKED ISSUE #381: Locked attachments can still be deleted from the chatter of an object

victoralmau:
FWP from 16.0: https://github.com/OCA/dms/pull/397

Delete the file if the attachment is deleted

When deleting an attachment, an attempt should be made to delete the linked file, causing an error if it is locked. Do not show the Delete/Archive/Unarchive option on files if it is locked and you cannot unlock it.

Fixes https://github.com/OCA/dms/issues/381

@Tecnativa

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -197,6 +197,7 @@ class DMSFile(models.Model):
         string="Attachment File",
         prefetch=False,
         ondelete="cascade",
+        index=True,
     )
 
     def get_human_size(self):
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):
/ocabot merge patch

PR #411, opened 2025-04-21
[17.0][FIX] dms: Show short name of directories in searchpanel

### LINKED ISSUE #408: Parent Directory is repeatedly shown in subdirectory tree view
First of all, thank you for your continued efforts in maintaining and improving the DMS module. Your work is greatly appreciated by the community.

I’ve recently tried the version 17.0 and noticed a change in the directory tree view where the parent directory is now shown within the child directory. In version 16.0, the tree structure was more intuitive and cleaner, making it easier to navigate documents efficiently.

V16.0
Image

V17.0
Image

Would it be possible if tree view in v17.0 can function like it did in v16.0. I believe this could enhance usability for many users who rely on a clean and hierarchical folder structure.

Thanks.

victoralmau:
Show short name of directories in searchpanel

**Before**
antes

**After**
despues

Fixes #408 

@Tecnativa

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -698,6 +698,14 @@ class DmsDirectory(models.Model):
             res = super().write(vals)
         return res
 
+    @api.depends_context("directory_short_name")
+    def _compute_display_name(self):
+        if self.env.context.get("directory_short_name"):
+            for item in self:
+                item.display_name = item.name
+        else:
+            return super()._compute_display_name()
+
     def unlink(self):
         """Custom cascade unlink.
 


dms/models/directory.py
```diff
@@ -698,6 +698,14 @@ def write(self, vals):
             res = super().write(vals)
         return res
 
+    @api.depends_context("directory_short_name")
+    def _compute_display_name(self):
+        res = super()._compute_display_name()
```
pedrobaeza: Don't call super if the context is there

pedrobaeza (approved):

pedrobaeza (approved):

PR #415, opened 2025-04-28
[17.0][FIX] dms: searchpanel changes apply only to dms

giarve:
Fwport https://github.com/OCA/dms/pull/414

PR #418, opened 2025-04-30
[18.0][FIX] dms: searchpanel changes apply only to dms

victoralmau:
FWP from 17.0: https://github.com/OCA/dms/pull/415

searchpanel changes apply only to dms

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #419, opened 2025-04-30
[17.0][IMP] dms: avoid unnecessary DMS queries in unlink

kaynnan:
cc @marcelsavegnago @WesleyOliveira98 

This prevents redundant operations and guards against potential recursion or unnecessary database hits.

PR #420, opened 2025-05-07
[18.0][IMP] dms: avoid unnecessary DMS queries in unlink

victoralmau:
FWP from 17.0: https://github.com/OCA/dms/pull/419

This prevents redundant operations and guards against potential recursion or unnecessary database hits.

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #424, opened 2025-05-12
[18.0][MIG] hr_dms_field: Migration to 18.0

victoralmau:
Migration to 18.0

Please @CarlosRoca13 can you review it?

@Tecnativa TT55524

PR #427, opened 2025-05-12
[18.0] [MIG] hr_dms_field: Migration to 18.0

luisDIXMIT:
Standard migration


victoralmau:
The migration of this module is already available at https://github.com/OCA/dms/pull/424.

luisDIXMIT:
> The migration of this module is already available at #424.

Thanks mate. I will close this then.

Issue #434, opened 2025-06-09
[18.0] dms: Broken access rules


LorenzoC0:
## Module
dms

## Describe the bug
The write, unlink and create access permissions defined in the dms groups are not working as intended. Every operation is allowed regardless of the permissions the user has at that time.

## To Reproduce
**Affected versions**:
- [x] 18.0
- [ ] 16.0 - Not affected 

1. Create a new access group with "Create access" and "Unlink access" active. Leave "Write access" unchecked.
2. Make sure the user you're on is included in the access group
3. Setup a storage with a root folder.
4. Assign the newly created group to the root folder just created.
5. Upload a file in the folder.
6. Open Form view of the file.
7. Create a new Tag and add it to the file
8.  Save.

**Expected behavior**
The expected behavior after these steps is that the user should not be able to save the file since it does not have "Write access" as a permission.

LorenzoC0:
https://github.com/user-attachments/assets/21a3477b-1498-44db-b549-62c12124e218

Here's a video of the issue on the latest Runboat

eLBati:
@LorenzoC0 trying to fix with https://github.com/OCA/dms/pull/435

eLBati:
@kobros-tech @victoralmau any opinion about this? Thanks!

kobros-tech:
> [@kobros-tech](https://github.com/kobros-tech) [@victoralmau](https://github.com/victoralmau) any opinion about this? Thanks!

Can you try something else rather than tag or category?

like update the name of the root folder.

LorenzoC0:
https://github.com/user-attachments/assets/d844beeb-4761-4413-865e-891558e27e3a

@kobros-tech Yes, here's a video editing the root folder, I hope it's not compressed too hard.

As you can see I cannot edit the name of the folder, but the error indicates a missing permission with the dms access group records. Then I tried to directly delete the folder on the kanban view and it successfully does so without Unlink access in the group of the user

victoralmau:
IMO this has already been fixed https://github.com/OCA/dms/pull/436#issuecomment-2969488112

PR #435, opened 2025-06-09
[18.0] [FIX] dms: The write, unlink and create access permissions defined in the dms groups are not working as intended

eLBati:
See https://github.com/OCA/dms/issues/434

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -151,8 +151,12 @@ class DMSFile(models.Model):
                 one.image_1920 = one.content
 
     def check_access(self, operation):
-        self.mapped("directory_id").check_access(operation)
-        return super().check_access(operation)
+        self.mapped("directory_id").with_context(dms_operation=operation).check_access(
+            operation
+        )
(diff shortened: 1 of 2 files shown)


pedrobaeza:
AFAIK, they are working as intended. Please check previous discussions about it.

kobros-tech:
> AFAIK, they are working as intended. Please check previous discussions about it.

@pedrobaeza what if we end up this issue from arising any more by writing more usage description for all usage cases?

pedrobaeza:
Yes, any improvement about documentation is always welcome.

eLBati:
Hello @pedrobaeza @kobros-tech so, if user does not have unlink permission on folder, they should be able to delete folder? See  https://github.com/OCA/dms/issues/434#issuecomment-2956301628
Thanks

eLBati:
Also, this apparently wrong behavior only appears on v18

pedrobaeza:
I haven't dug enough, so I let @victoralmau's to say. At least, the CI in this PR should be green.

pedrobaeza:
Replaced by #436

PR #437, opened 2025-06-16
[16.0][FIX]dms: Error with o-text-overflow on the o_kanban_details_wrapper div.

adasatorres:
Good morning, it seems that sometimes the directory or file name causes a line break and pushes the text outside of the template. I haven't been able to reproduce the issue in other environments, not even in the same one, where there are folders with longer names that work fine. I’ve tested it on different screens, and the result varies. Forcing display: block; seems to fix the issue.

Captura desde 2025-06-16 15-39-12


edescalona (approved):
LGTM

rrebollo (approved):
Code Review. LGTM!

PR #438, opened 2025-07-02
[IMP] dms_field: Make change of record work

etobella:
I detected that it failed when we use the field inside a record and we change the record using the next or previous buttons.

The record was not refreshed properly.

This should fix it.

@CarlosRoca13

### DIFF
--- a/dms_field/static/src/views/dms_list/dms_list_renderer.esm.js
+++ b/dms_field/static/src/views/dms_list/dms_list_renderer.esm.js
@@ -65,9 +65,18 @@ export class DmsListRenderer extends Component {
         });
         onMounted(() => {
             this.$tree = this.$(this.js_tree.el);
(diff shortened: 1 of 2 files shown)


dms_field/static/src/views/dms_list/dms_list_renderer.esm.js
```diff
@@ -357,8 +366,11 @@ export class DmsListRenderer extends Component {
             await this.props.rendererActions.onDMSLoad(node);
         result.then((data) => {
             callback.call(this, data);
+            console.log("DmsListRenderer: loadData", data);
```
pedrobaeza: Remove this.

CarlosRoca13 (approved):
Thanks!

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #439, opened 2025-07-11
[18.0][FIX]dms: Added a check that prevents DB curruption

Wvven:
When an access group is deleted without removing the related directories, the user will be stucked with errors when trying to create new directories. It happens when you have only one Access Group in the dms module

### DIFF
--- a/dms/models/access_groups.py
+++ b/dms/models/access_groups.py
@@ -175,3 +175,16 @@ class DmsAccessGroups(models.Model):
                         "current": one.display_name,
                     }
                 )
+
+    # Prevent deletion of groups that are assigned to directories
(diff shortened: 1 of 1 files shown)


victoralmau (commented):
IMO I don't think this is totally necessary and/or safe, i.e. deleting a group that is assigned to directories, should not be a problem if you are aware of what is involved.

Examples:
- A “specific” group that is assigned to a many of directories is deleted. There will be no problem because there will be other groups defined that still give access to files/directories to some user.
- A “generalist” group that is assigned to a many of directories is deleted. There will be no problem if there are more groups assigned to those directories that indirectly still give access to some user.

PR #440, opened 2025-07-16
[ADD]dms_zip_download: Added module

LorenzoC0:
Added a new module that allows users to download multiple selected files as a zip archive.


dms_zip_download/wizards/dms_file_download_wizard.py
```diff
@@ -0,0 +1,71 @@
...
+                    max_size=max_size_mb,
+                )
+                raise UserError(error_msg)
+
+        zip_buffer = io.BytesIO()
+
+        with zipfile.ZipFile(zip_buffer, "w", zipfile.ZIP_DEFLATED) as zip_file:
+            for file in self.file_ids:
+                file_content = file.with_context(bin_size=False).content
+                if file_content:
```
victoralmau, line 46: I don't understand this condition, the file will always have content, right? Therefore, you can remove this condition and one less indentation.

dms_zip_download/wizards/dms_file_download_wizard.py
```diff
@@ -0,0 +1,71 @@
...
+                    original_name = file_name
+                    while file_name in zip_file.namelist():
+                        name_parts = original_name.rsplit(".", 1)
+                        if len(name_parts) > 1:
+                            file_name = (
+                                f"{name_parts[0]}_{file_counter}.{name_parts[1]}"
+                            )
+                        else:
+                            file_name = f"{original_name}_{file_counter}"
+                        file_counter += 1
```
victoralmau, line 60: IMO this should not be necessary, because the .zip file should contain the entire directory/subdirectory structure (at any level) and the files in the appropriate directory. Also, with this other approach, there will never be files that have the same name https://github.com/OCA/dms/blob/f1ed453f2ff997f7bea8c8ce31028f47c7a46705/dms/models/dms_file.py#L512.

Considering this suggestion, I think it would also be interesting to add the possibility to “download a zip” of one/several directories, what do you think?

victoralmau (commented):
I think this module will be interesting, but is it really necessary to have it separate? 
I think it would be more appropriate to have it inside dms, what do you think @pedrobaeza and @etobella ?

etobella:
If it has no external dependencies, it could be included in the main module😁

PR #441, opened 2025-08-18
[16.0][IMP] dms_field: Propagate groups from field template

SirPyTech:
With this PR, the groups declared in child directories of the field template are propagated to the directories created in the target record.

I have also included a couple of small refactorings, in their own commit.

Let me know what you think!


dms_field/tests/test_dms_field.py
```diff
@@ -244,3 +244,40 @@ def test_parents(self):
...
+        )
+        self.assertRecordValues(
+            new_child_dir,
+            [
+                {
+                    "inherit_group_ids": child_dir.inherit_group_ids,
+                    "group_ids": child_dir.group_ids.ids,
+                }
+            ],
+        )
```
```suggestion
        subdirectory_1 = partner.dms_directory_ids.child_directory_ids.filtered(
            lambda x: x.name == self.subdirectory_1.name
        )
        self.assertFalse(subdirectory_1.inherit_group_ids)
        self.assertEqual(subdirectory_1.group_ids, access_group)
        subdirectory_2 = partner.dms_directory_ids.child_directory_ids.filtered(
            lambda x: x.name == self.subdirectory_2.name
        )
        self.assertTrue(subdirectory_2.inherit_group_ids)
        self.assertIn(self.group, subdirectory_2.group_ids)
```
IMO this is more readable and checks that both subdirectories have the right data.

victoralmau (changes requested):
Code review OK

By default, the value `inherit_group_ids` is True, so until now, it was always set to True in child directories. Changing this IMO has no side effects and allows for more flexibility in more complex cases (where a subdirectory of a template has different permissions than others).

victoralmau:
I think all these changes can be combined into a single commit: _[IMP] dms_field: Propagate groups from field template_

The title of the PR should be changed to something like: _[16.0][IMP] dms_field: Propagate groups from field template_

dms_field/tests/test_dms_field.py
SirPyTech: Thanks for the quick review!
This needed some tweaking, but I accepted your suggestion overall, please check

SirPyTech:
> I think all these changes can be combined into a single commit: _[IMP] dms_field: Propagate groups from field template_

I prefer to highlight which commits do not change the behavior (`[REF]` ones) from the other ones, and to keep each commit focused on their change, so I'd like to keep them all.

> The title of the PR should be changed to something like: _[16.0][IMP] dms_field: Propagate groups from field template_

Right, added more details in the PR title, I'm not a fan of including the version in the PR title though.
If you need that for filters, note that PRs can be filtered by version `16.0` using `base:16.0` filter (see https://github.com/OCA/dms/pulls?q=is%3Apr+is%3Aopen+base%3A16.0).

pedrobaeza:
@SirPyTech the version has to be in the PR title, not in the commit message.

SirPyTech:
> > The title of the PR should be changed to something like: _[16.0][IMP] dms_field: Propagate groups from field template_
> 
> Right, added more details in the PR title, I'm not a fan of including the version in the PR title though.
If you need that for filters, note that PRs can be filtered by version `16.0` using `base:16.0` filter (see https://github.com/OCA/dms/pulls?q=is%3Apr+is%3Aopen+base%3A16.0).

> @SirPyTech the version has to be in the PR title, not in the commit message.

Why does it "have to be" there?
If that's used for filtering the PRs, the dedicated filter `base` already exists as I wrote above

pedrobaeza:
Current convention is for seeing in a glance which is the target branch from the pull request list or the received notification mails, without having to navigate to the PR itself to see it. Please make reviewers' life easier.

SirPyTech:
> Current convention is for seeing in a glance which is the target branch from the pull request list or the received notification mails, without having to navigate to the PR itself to see it. Please make reviewers' life easier.

I am a reviewer too and I have never needed the version in the title because I'm using the `base` filter, and I think that should be the way for searching PRs of a specific version; in the rare occasions I needed more, I used GitHub's API.

Could you please propose this convention to the Guidelines in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst?
The contributors already have so many guidelines to follow, it would be nice if they were at least all written somewhere.

When this new convention is in the guidelines, maybe the OCA bot could help enforcing it :robot: and you wouldn't need to ask all the contributors to edit their PR's title.

pedrobaeza:
There's no section in guidelines for that, and I'm not going to lose more time in that bureaucracy given the result in other attempts, but if someone asks you for that, and that would ease the review, please do it as I do. If not, then we will be reluctant to review your PRs. It's that simple.

victoralmau:
My comment is still pending https://github.com/OCA/dms/pull/441#issuecomment-3196129887

_I think all these changes can be combined into a single commit: [IMP] dms_field: Propagate groups from field template_

IMO, all changes can be combined in a single commit; having several in this case adds nothing.

SirPyTech:
> My comment is still pending [#441 (comment)](https://github.com/OCA/dms/pull/441#issuecomment-3196129887)
> 
> _I think all these changes can be combined into a single commit: [IMP] dms_field: Propagate groups from field template_
> 
> IMO, all changes can be combined in a single commit; having several in this case adds nothing.

Why is this case special?
By the way, I already answered that one in https://github.com/OCA/dms/pull/441#issuecomment-3196190899:
> I prefer to highlight which commits do not change the behavior (`[REF]` ones) from the other ones, and to keep each commit focused on their change, so I'd like to keep them all.

It happens to have different opinions :shrug: I think having different commits makes more sense and is more aligned with the guidelines:
> Only make a single commit per logical change set.

(from https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message)

Rewriting a method to allow override is a logical change set, and that deserves a commit on its own.
Changing the behavior of a method is another logical change set too, and that deserves a commit on its own.

Would you agree with a compromise solution of having one `[REF]` commits that includes the existing `[REF]` commits, and one `[IMP]` commit?

HekkiMelody (approved):
Code and functional review, LGTM, we've successfully built an implementation based on this.

Regarding the more formal disputes: the separate [REF] commits seem cleaner to me, and also make reviewing and future forward porting easier.

On the other hand, if there's strong maintainer preference for a single commit, I think in this specific case the changes are small enough that I wouldn't want to block the PR over this disagreement.

victoralmau (approved):

SirPyTech:
@OCA/document-maintainers can someone have a look?
This is a Beta module and the PR has 2 approvals from 4 months ago, so it should be good for merge.
Thanks!

dms_field/models/dms_access_group.py
```diff
@@ -38,18 +40,28 @@ def _compute_company_id(self):
...
         for item in self.filtered("dms_field_ref"):
-            dms_field_ref = "%s,%s" % (item.dms_field_ref._name, item.dms_field_ref.id)
-            if self.search(
-                [("dms_field_ref", "=", dms_field_ref), ("id", "!=", item.id)]
-            ):
+            domain = expression.AND(
+                [
+                    item._get_domain_for_item_from_dms_field_ref(item.dms_field_ref),
+                    [
+                        ("id", "!=", item.id),
```
pedrobaeza: Remove the trailing comma for having these 3 lines into 1.

pedrobaeza (approved):
/ocabot merge patch

dms_field/models/dms_access_group.py
SirPyTech: Thanks for having a look!
I usually have every item in the domain in their own line so that changes to each item do not highlight the whole domain as changed in diffs, but in this domain that only has one item it probably makes little sense :sweat_smile:

P.S.: Thanks for merging too!

pedrobaeza: Yeah, but at the end, diff viewers like the one in GitHub are already highlighting the specific part of the line that is changed, and the probability that this will be changed is very low.

PR #444, opened 2025-09-08
Add buttons to the document kanban card.

marcus-gpi:
The action buttons are rather hidden away in the kanban context view.

This PR brings them right onto the kanban card.

### DIFF
--- a/dms/views/dms_file.xml
+++ b/dms/views/dms_file.xml
@@ -247,6 +247,18 @@
                                     </div>
                                     <div class="o_kanban_record_bottom">
                                         <div class="oe_kanban_bottom_right">
+                                            <a type="edit">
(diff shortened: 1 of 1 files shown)


victoralmau (changes requested):
Excuse me, why do we need these icons?
dms-0

The actions already exist (in addition to previewing when you click on the thumbnail).
dms-1

marcus-gpi:
No need to accept them if you don't want them. However, this is major usability improvement for my users, and I thought I'd suggest them if you wanted them.

pedrobaeza:
I would put that clicking on the icon download the file. About the edition, clicking on the rest of the kanban card should go to the form view where the edition takes place. Finally, I wouldn't put deletion on one click for avoiding errors. Are the first 2 things covered already. If not, I think that can be a good improvement for all.

PR #448, opened 2025-10-03
[18.0][IMP] dms_field: Use ir.assets for xml to add the sequence and avoid race condition with hr_holidays accrual plan

etobella:
We have detected that the dms_field way of defining the field is providing issues with hr_holidays on the view accural. If hr_holidays and dms_field are both installed and we try to access accruals, we receive the following error:

image

this comes from the definition of accrual plan widget in 18 (fixed in 19). The way I found to solve it is to move the DMSList in this case too.

https://github.com/odoo/odoo/blob/18.0/addons/hr_holidays/static/src/xml/x2many_field.xml

It affects 17.0 too, but not 16.0


pedrobaeza:
Can't you trick the position of loading here? https://github.com/OCA/dms/blob/f7bacd4f6ee0e64fc6798f5544a04fdd943189b9/dms_field/__manifest__.py#L27

etobella:
Not directly, but by using ir.assets I am able to do that. WDYT?

pedrobaeza:
@CarlosRoca13 I think you did the asset thing. Can you check if this is correct?

CarlosRoca13 (approved):
Thanks!

pedrobaeza (approved):
/ocabot merge patch

PR #458, opened 2026-01-13
[18.0][IMP] dms_field: Avoid to recharge model on each load

CarlosRoca13:
cc @Tecnativa TT60256

ping @pedrobaeza @victoralmau


pedrobaeza (approved):
Working as expected. The question is why that was introduced. Maybe that was preventing some glitch?

victoralmau (approved):

PR #459, opened 2026-01-20
[18.0][FIX] dms: Fix the inheritance of the search_panel so that the domain is applied correctly.

CarlosRoca13:
Before these changes, the domain was never applied to search by the parent, so all subfolders were being displayed instead of only the direct children of the selected one.

After making the changes, only a single level of folders is shown, without displaying all descendant folders.

cc @Tecnativa TT59933

ping @victoralmau @pedrobaeza


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #461, opened 2026-02-05
[18.0][FIX] dms_auto_classification: Filter by the appropriate directories

victoralmau:
Filter by the appropriate directories

It is important not to apply `sudo()` to filter by the directories to which the user has access and to filter only by those of the appropriate company.

@Tecnativa TT60719


dms_field_auto_classification/wizards/wizard_dms_classification.py
```diff
@@ -46,9 +46,11 @@ def _compute_record_ref(self):
     def _compute_directory_id(self):
         """Overwrite to redefine the directory if the template has a linked model."""
         self_with_model = self.filtered(lambda x: x.template_id.model_id)
-        directory_model = self.env["dms.directory"].sudo()
+        company = self.parent_id.template_id.company_id or self.env.company
+        directory_model = self.env["dms.directory"]
         for item in self_with_model:
             domain = [
+                ("company_id", "=", company.id),
```
pedrobaeza: Shouldn't this be only if the target model has `company_id` in its fields (maybe it doesn't)?

victoralmau: These are two different things. It is one thing if the target model has the `company_id` field, and another if the directories we are searching to classify are from the right company (`company_id` field).

Extra: Although it is not something related to these modules, a directory linked to company A should not have a linked record from company B (that is not the problem in this case).

pedrobaeza (approved):
OK then. I don't know in deep the structure of this.

CarlosRoca13 (approved):

PR #462, opened 2026-02-06
[18.0][FIX] dms_field: Avoiding dms.directory access errors in some cases

victoralmau:
Avoiding `dms.directory` access errors in some cases

Use case:
- Modify the user Marc Demo and define some permissions in Employees
- Go to the employee Abigail Peterson and create her folder
- Modify the Access group > Everyone for Employee DMS and remove the corresponding group to leave only Mitchell Admin defined as an explicit user
- Log in with Marc Demo and try to modify some data for the employee Abigail Peterson (define a tag, for example).

**Before**
antes

**After**
despues

@Tecnativa TT60793

### DIFF
--- a/dms_field/models/dms_field_mixin.py
+++ b/dms_field/models/dms_field_mixin.py
@@ -21,6 +21,24 @@ class DMSFieldMixin(models.AbstractModel):
         auto_join=True,
     )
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

CarlosRoca13 (approved):
Thanks!

PR #466, opened 2026-03-11
[18.0][IMP] dms_field: Updated domain check for dms directory creation

anusriNPS:
Added domain related to `storage_id` which allows to create
directories for same records across multiple storages. It restricts to look for available directories within the storage related to the template while  raising ValidationError: `There is already a linked directory created.`


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #467, opened 2026-03-19
[17.0][FIX] dms: Change save_type when migrate to another storage

peluko00:
When migrating files from one storage to another, their save type needs to change to match the storage's **save_type**. After that, these files will no longer require migration.

cc https://github.com/APSL 38655
@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall @BernatObrador @ppyczko please review

### DIFF
--- a/dms/models/dms_file.py
+++ b/dms/models/dms_file.py
@@ -117,6 +117,8 @@ class DMSFile(models.Model):
         compute="_compute_save_type",
         string="Current Save Type",
         prefetch=False,
+        readonly=False,
+        store=True,
     )
 
     migration = fields.Char(
@@ -263,6 +265,7 @@ class DMSFile(models.Model):
                 {
                     "content": dms_file.with_context(**{}).content,
                     "storage_id": dms_file.directory_id.storage_id.id,
+                    "save_type": dms_file.directory_id.storage_id.save_type,
                 }
             )
 


victoralmau (changes requested):
IMP, this change isn't appropriate. Currently, the `save_type` field is set to `compute` (not `store`); perhaps it should be `store`, but in any case, the value shouldn't be defined this way. The proper approach is to set the appropriate value in the compute (`_compute_save_type`) based on different scenarios and/or add a `depends` clause if necessary.

To clarify, I have never used this migration action.

peluko00:
> IMP, this change isn't appropriate. Currently, the `save_type` field is set to `compute` (not `store`); perhaps it should be `store`, but in any case, the value shouldn't be defined this way. The proper approach is to set the appropriate value in the compute (`_compute_save_type`) based on different scenarios and/or add a `depends` clause if necessary.
> 
> To clarify, I have never used this migration action.

When executes action_migrate they need change the `save_type` because if you change from filestore to Amazon s3 if you don't change this attribute the migration will be failed.

victoralmau:
Although I've never used the `action_migrate()` method, I do understand its purpose: to change the storage type from database to file or vice versa. That said, I believe the current approach is incorrect (and the changes in this PR are as well).

The`https://github.com/OCA/dms/blob/5bcd966d9a9007952a646d9612263e426511f31a/dms/models/dms_file.py#L247 method  should check whether all the storage_id of the selected records have the same save_type; if not, it shouldn’t be allowed. In any case, I don’t think this approach is the right one either.

In my opinion, the correct change should be:
- Allow changing the `save_type` in `dms.storage` (except from or to attachment)
- Define an inverse in the storage’s save_type field to perform the migration.
- The https://github.com/OCA/dms/blob/5bcd966d9a9007952a646d9612263e426511f31a/dms/actions/file.xml#L7 action would not be necessary.
- The `migration` and `require_migration` fields in `dms.file` would not be necessary.
- The change https://github.com/OCA/dms/blob/5bcd966d9a9007952a646d9612263e426511f31a/dms/models/dms_file.py#L265 would not be necessary.
- We should change `content_binary` or `content_file` in https://github.com/OCA/dms/blob/5bcd966d9a9007952a646d9612263e426511f31a/dms/models/dms_file.py#L264

What do you think, @etobella ?

peluko00:
> Although I've never used the `action_migrate()` method, I do understand its purpose: to change the storage type from database to file or vice versa. That said, I believe the current approach is incorrect (and the changes in this PR are as well).
> 
> The`
> 
> https://github.com/OCA/dms/blob/5bcd966d9a9007952a646d9612263e426511f31a/dms/models/dms_file.py#L247
> 
> method should check whether all the storage_id of the selected records have the same save_type; if not, it shouldn’t be allowed. In any case, I don’t think this approach is the right one either.
> In my opinion, the correct change should be:
> 
> * Allow changing the `save_type` in `dms.storage` (except from or to attachment)
> * Define an inverse in the storage’s save_type field to perform the migration.
> * The https://github.com/OCA/dms/blob/5bcd966d9a9007952a646d9612263e426511f31a/dms/actions/file.xml#L7
>    action would not be necessary.
> * The `migration` and `require_migration` fields in `dms.file` would not be necessary.
> * The change https://github.com/OCA/dms/blob/5bcd966d9a9007952a646d9612263e426511f31a/dms/models/dms_file.py#L265
>    would not be necessary.
> * We should change `content_binary` or `content_file` in https://github.com/OCA/dms/blob/5bcd966d9a9007952a646d9612263e426511f31a/dms/models/dms_file.py#L264
> 
> What do you think, @etobella ?

The reason for this change is precisely to support that kind of use case: allowing files currently stored in the filestore to be migrated to another storage backend, such as S3, by changing the storage configuration and moving the existing file contents accordingly. So the goal is not only database ↔ file migration, but making storage backend transitions possible without requiring a separate manual action on dms.file.

PR #470, opened 2026-03-27
[18.0][IMP] dms: Clean up attachments after creation of file

### LINKED ISSUE #468: [18.0] dms: attachment created during uploading of a new file is new garbage collected
## Module
dms

## To Reproduce (scenario 1)
**Affected versions**: 18.0

1. Go to any directory click **Upload**;

Image

2. a `dms.file` is created and an `ir.attachment`;

Image
 
4. run scheduled action "Base: Auto-vacuum internal data", and the attachment is not deleted;
5. delete the `dms.file`;
6. run the scheduled action again;
7. the `ir.attachment` is still not deleted.

## To Reproduce (scenario 2)
1. Don't select a directory;
2. upload a file, and you get the error "You must select a directory first";
3. the `ir.attachment` is created though.

**Expected behavior**
I would expect the `ir.attachment` created during uploading new files to be garbage collected, or deleted directly.

**Additional context**
I'm using filestore storage.

Image

tarteo:
Fixes: https://github.com/OCA/dms/issues/468

### DIFF
--- a/dms/static/src/js/views/dms_file_upload.esm.js
+++ b/dms/static/src/js/views/dms_file_upload.esm.js
@@ -97,7 +97,15 @@ export function createFileUploadExtension() {
                 throw new Error(attachments.error);
             }
 
-            await this.onUpload(attachments);
+            await this.onUpload(attachments).finally(() => {
+                // Delete the attachments once we're done with them
+                const attachmentIds = attachments.map((a) => a.id);
+                // If the upload failed, don't waste an unlink call
+                if (!attachmentIds.length) {
+                    return;
+                }
+                this.orm.call("ir.attachment", "unlink", [attachmentIds]);
+            });
         },
 
         async onUpload(attachments) {


victoralmau (approved):

pedrobaeza (commented):
Why an attachment is created in first place?

tarteo:
@pedrobaeza It reuses `/web/binary/upload_attachment` that creates the attachment, but functionally I don't think there's a reason.

pedrobaeza:
Then isn't better to do the upload other way for not generating that garbage?

tarteo:
@pedrobaeza I don't know if it's used for something, I assume there is / was a reason to use the endpoint.

pedrobaeza:
I don't think so apart from reducing the code in this part, but having that side effects, this is to be questioned. @victoralmau can you confirm?

victoralmau:
> I don't think so apart from reducing the code in this part, but having that side effects, this is to be questioned. @victoralmau can you confirm?

I see that this was added in v16 (https://github.com/OCA/dms/pull/262), since there is already a method in the `web` version with similar functionality (https://github.com/odoo/odoo/blob/af32885ec5f07d492f3b8e8fff1785996a739f72/addons/web/controllers/binary.py#L217), I infer that the intention was to reduce the code a bit and use it; therefore, the change added in this PR is necessary.

pedrobaeza:
Yes, but as said, with this high side effect, I would switch again to have the complete upload handle here in this module instead of using the other, but generating the garbage. Maybe a part of the other code can be used without generating the attachment.

victoralmau:
> Yes, but as said, with this high side effect, I would switch again to have the complete upload handle here in this module instead of using the other, but generating the garbage. Maybe a part of the other code can be used without generating the attachment.

Another approach: https://github.com/OCA/dms/pull/472

tarteo:
Yes I agree if there's no reason but reducing code we should create a handler directly in DMS that doesn't create the attachment!

PR #472, opened 2026-04-29
[18.0][IMP] dms + dms_file: Upload file and create dms.file directly

victoralmau:
Upload file and create `dms.file` directly

Related to https://github.com/OCA/dms/pull/470#issuecomment-4212192789

Please @CarlosRoca13 can you review it?

@Tecnativa

### DIFF
--- a/dms/controllers/main.py
+++ b/dms/controllers/main.py
@@ -1,7 +1,14 @@
 # Copyright 2017-2019 MuK IT GmbH
+# Copyright 2026 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
-from odoo import http
+import json
+import unicodedata
+
+from odoo import _, http
+from odoo.exceptions import AccessError
 from odoo.http import request
+
+from odoo.addons.web.controllers.binary import clean
 
 
 class OnboardingController(http.Controller):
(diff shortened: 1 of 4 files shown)


tarteo (approved):
LGTM like what we discussed in my PR. Can't test it functionality at the moment

pedrobaeza:
@victoralmau still draft?

victoralmau:
Sorry, it's ready to review

Ping @CarlosRoca13

pedrobaeza:
Testing this in runboat, the file uploads, but the thumbnail for images is not created. Can you please check?

victoralmau:
> Testing this in runboat, the file uploads, but the thumbnail for images is not created. Can you please check?

Fixed, can you review again?

dms/static/src/js/views/dms_file_upload.esm.js
```diff
@@ -133,37 +104,22 @@ export function createFileUploadExtension() {
...
-                        type: "danger",
-                    });
-                    self.actionService.restore(controllerID);
-                });
+            const fileData = await this.http.post(
+                "/web/binary/upload_dms_file",
+                params,
+                "text"
+            );
+            const attachments = JSON.parse(fileData);
```
pedrobaeza: Why calling this `attachments`?

dms/controllers/main.py
```diff
@@ -1,9 +1,18 @@
...
+import json
+import unicodedata
+
+from odoo import _, http
+from odoo.exceptions import AccessError
 from odoo.http import request
 
 
+def clean(name):
+    return name.replace("\x3c", "")
```
pedrobaeza: Can you document why this cleaning is needed? And why not doing it inline in the first file name extraction, which is used in the rest of places?

victoralmau: Keep in mind that this entire approach to uploading a file already exists on the `web` (https://github.com/odoo/odoo/blob/4b9ab0e3a5dbfa571377570016b9582320f87f20/addons/web/controllers/binary.py#L217), and we use the same method; that method is the same as on the web https://github.com/odoo/odoo/blob/4b9ab0e3a5dbfa571377570016b9582320f87f20/addons/web/controllers/binary.py#L47

pedrobaeza: Then mention exactly this as comment of this method and the other one.

dms/controllers/main.py
```diff
@@ -1,9 +1,19 @@
...
+import json
+import unicodedata
+
+from odoo import _, http
+from odoo.exceptions import AccessError
 from odoo.http import request
 
 
+def clean(name):
+    """Similar method https://github.com/odoo/odoo/blob/4b9ab0e3a5dbfa571377570016b9582320f87f20/addons/web/controllers/binary.py#L47 used in upload_dms_file()."""  # noqa: E501
```
pedrobaeza: Then no need to verbatim copy this. Do:

```
from odoo.addons.web.controllers.binary import clean
```

and then use it.

pedrobaeza (approved):
/ocabot merge minor

PR #479, opened 2026-06-03
[18.0][FIX] dms_field: scope autocomputed directories by model

windweaver828:
Autocomputed DMS directories for embedded record views were filtered only by `res_id`.

Because record IDs are not globally unique across Odoo models, this could make a record such as `hr.employee,1` show directories belonging to another model with the same ID, for example `dms.field.template,1`.

This change scopes the autocomputed directory domain by both `res_id` and `res_model`.

Tested:
- `node --check dms_field/static/src/views/dms_list/dms_list_controller.esm.js`
- `git diff --check`

### DIFF
--- a/dms_field/__manifest__.py
+++ b/dms_field/__manifest__.py
@@ -5,7 +5,7 @@
     "name": "DMS Field",
     "summary": """
         Create DMS View and allow to use them inside a record""",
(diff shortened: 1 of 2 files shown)


victoralmau:
This has already been merged into https://github.com/OCA/dms/pull/504

PR #481, opened 2026-06-04
[18.0][ADD] dms_file_description: add descriptions to DMS files

windweaver828:
Adds two DMS modules:

- `dms_file_description`: adds editable descriptions to `dms.file` and displays them in DMS form, list, kanban, and search views.
- `dms_field_file_description_preview`: auto-install bridge that shows file descriptions in the embedded DMS preview pane from `dms_field`.

The preview integration is split from the base module so `dms_file_description` only depends on `dms`.

Tested on odoo-demo with:

```bash
./run_odoo.sh -i dms_field_file_description_preview,dms_ui_enhancements_file_description_preview -u dms_file_description --test-enable --test-tags /dms_file_description --log-level=test

### DIFF
--- a/dms_field_file_description_preview/README.md
+++ b/dms_field_file_description_preview/README.md
@@ -1,0 +1,12 @@
+# DMS Field File Description Preview
+
+Shows DMS file descriptions in the embedded DMS preview pane provided by `dms_field`.
+
+## Usage
+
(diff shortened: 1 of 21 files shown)


pedrobaeza:
I think this can be integrated into the main module. What others think?

PR #484, opened 2026-06-19
[19.0][ADD] dms_libreoffice_preview: server-side LibreOffice→PDF office preview

dnplkndll:
Adds in-browser preview of office files (doc/docx/odt/xls/xlsx/ods/ppt/pptx/odp/rtf) in DMS by registering an office-mimetype handler (score 10) into the dms_preview_pane preview registry: the server converts the file to PDF via LibreOffice (cached as an ir.attachment, keyed on the content checksum, 60s soffice cap). Leaves score 20 free for a future dms_onlyoffice.

Depends on OCA/dms#483 (dms_preview_pane), pinned in `test-requirements.txt` until it lands — the `Detect unreleased dependencies` red is that pin. The `dms` pin is gone now that OCA/dms#475 has merged, and the branch is rebased on `19.0`.

Requires the libreoffice deb (`external_dependencies`). The conversion tests skip themselves when `soffice` is absent, which it is on the OCA runners; they were run locally against a real LibreOffice to confirm they pass rather than merely skip.

AI-assisted (Claude Code); every change reviewed, tested, and owned by the author.


pedrobaeza:
This is covered in a more generic way in https://github.com/OCA/mail/pull/258

PR #492, opened 2026-06-23
[FIX] dms: use _slugify for subject in directory message_new

### LINKED ISSUE #491: dms: incoming email to directory alias crashes in message_new (_slug misused on subject string)
**Module**: `dms` (18.0.1.1.0)
**Odoo**: 18.0

## Description

When a `dms.directory` exposes an email alias and is configured with **"Unpack Emails as: Subdirectory"** (`alias_process = 'directory'`, the default), processing an **incoming email** raises a `ValueError` in `message_new`, so the email cannot be ingested into the directory.

## Steps to reproduce

1. Create a `mail.alias.domain` (Settings) and set an **Email Alias** on a root `dms.directory`.
2. Leave **Unpack Emails as = Subdirectory** (`alias_process = 'directory'`).
3. Route an email (with an attachment) to that alias, e.g.:
   `env['mail.thread'].message_process(None, raw_mime)`.

## Result (traceback)

```
File ".../dms/models/directory.py", line 629, in message_new
    subject = slug(msg_dict.get("subject", _("Alias-Mail-Extraction")))
File ".../http_routing/models/ir_http.py", line 60, in _slug
    identifier, name = value
    ^^^^^^^^^^^^^^^^
ValueError: too many values to unpack (expected 2)
```

## Root cause

In `DmsDirectory.message_new` (`dms/models/directory.py`):

```python
slug = self.env["ir.http"]._slug
subject = slug(msg_dict.get("subject", _("Alias-Mail-Extraction")))
```

`ir.http._slug()` expects a **recordset or a `(id, name)` tuple** (it does `identifier, name = value.id, value.display_name`, falling back to `identifier, name = value`). Here it is given the **subject string**, so the unpacking fails.

The intent is to build a slug from the subject string, which is exactly what `ir.http._slugify()` does.

## Proposed fix

```python
slugify = self.env["ir.http"]._slugify
subject = slugify(msg_dict.get("subject", _("Alias-Mail-Extraction")))
```

Verified locally: with `_slugify`, an incoming email to the directory alias creates the expected subdirectory (named after the slugified subject) with the attachment as a `dms.file`. A PR with the fix follows.

mmircoli-nexapp:
Fixes #491.

## Problem

Incoming emails routed to a `dms.directory` alias configured with **Unpack Emails as: Subdirectory** (`alias_process = 'directory'`, the default) crash in `message_new`:

```
ValueError: too many values to unpack (expected 2)
```

`message_new` calls `ir.http._slug()` on the **subject string**, but `_slug()` expects a recordset or a `(id, name)` tuple, not a plain string.

## Fix

Use `ir.http._slugify()` — the helper meant to slugify a plain string — to build the subdirectory name from the email subject.

## Test

Verified manually: routing an email (with an attachment) to the directory alias now creates the expected subdirectory, named after the slugified subject, with the attachment stored as a `dms.file`. Happy to add an automated regression test under `dms/tests/` if reviewers prefer.

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -625,8 +625,8 @@ class DmsDirectory(models.Model):
             parent_directory._process_message(msg_dict)
             return parent_directory
         names = parent_directory.child_directory_ids.mapped("name")
-        slug = self.env["ir.http"]._slug
-        subject = slug(msg_dict.get("subject", _("Alias-Mail-Extraction")))
+        slugify = self.env["ir.http"]._slugify
+        subject = slugify(msg_dict.get("subject", _("Alias-Mail-Extraction")))
         defaults = dict(
             {"name": unique_name(subject, names, escape_suffix=True)}, **custom_values
         )


pedrobaeza:
Next time, put directly the PR with the issue explanation, without the GitHub issue.

pedrobaeza (approved):

victoralmau (approved):

Issue #500, opened 2026-08-06
[18.0][dms_field] dms.field.template root directory is created but not displayed and second click raises duplicated directory error


DealTechSrl:
Steps to reproduce

1. Install `dms_field`.
2. Create a `dms.field.template` manually from the UI.
3. Set:
   - a DMS storage
   - a model
   - one or more DMS access groups
4. Open the `Documents` tab of the template.
5. Click the storage initialization button/icon.

## Current behavior

The root `dms.directory` for the `dms.field.template` is created in the database, for example:

```text
name = My embedded2
res_model = dms.field.template
res_id = <template_id>
is_root_directory = True
root_directory_id = <same directory id>
```

However, the newly created directory is not displayed in the DMS widget.

If the user clicks the storage initialization button again, `create_dms_directory()` tries to create the same root directory again and raises:

```text
A directory with the same name already exists.
```

In some cases the created directory may also miss the template access groups, making it not readable by a regular user due to DMS record rules

## Suggested fix

Make the `dms.field.template` branch of `create_dms_directory()` idempotent by returning the existing linked root directory when present, and ensure missing template groups are applied.

Additionally, in the JS controller, include a fallback domain based on:

```text
res_model = dms.field.template
res_id = current template id
```

so the widget can display the newly created root directory even before the form one2many cache is refreshed.

Could you please confirm whether this is a duplicate of an existing issue or an expected behavior?

If this is confirmed as a bug, we have ready a pull request with the proposed backend and frontend adjustments.

DealTechSrl:
@victoralmau  message me if needed, I will pass you my PR

pedrobaeza:
You can do the PR directly if you prefer.

DealTechSrl:
@pedrobaeza @victoralmau PR is made

PR #501, opened 2026-08-06
[18.0][FIX] dms_field: make template root directory creation idempotent

DealTechSrl:
Issue: https://github.com/OCA/dms/issues/500
Changed files:
- dms_field/models/dms_field_template.py
- dms_field/static/src/views/dms_list/dms_list_controller.esm.js

Root cause:
When initializing the Documents tab of a dms.field.template, the backend created the template root directory, but repeated calls to create_dms_directory() could try to create the same root again, causing:

"A directory with the same name already exists."

The created directory could also be missing template access groups, making it invisible to regular users. On the frontend, the DMS widget relied on cached dms_directory_ids, so a newly created template root directory could exist in the database but not appear immediately in the tree.

Solution:
Make dms.field.template root directory creation idempotent, keep access groups aligned with the template, and adjust the widget domain so the created root can be displayed after refresh.

### DIFF
--- a/dms_field/models/dms_field_template.py
+++ b/dms_field/models/dms_field_template.py
@@ -79,6 +79,27 @@ class DmsFieldTemplate(models.Model):
         record = self.env[res_model].browse(res_id)
         directory_model = self.env["dms.directory"].sudo()
         if res_model == "dms.field.template":
+            existing_directory = directory_model.search(
+                [
+                    ("storage_id", "=", record.storage_id.id),
+                    ("res_id", "=", record.id),
+                    ("res_model", "=", record._name),
+                    ("is_root_directory", "=", True),
+                ],
+                limit=1,
+            )
+            if existing_directory:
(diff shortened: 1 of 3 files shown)


victoralmau:
In my opinion, the best solution would be not to display the button to create the directory if it already exists.

DealTechSrl:
The PR fix problems due to recursive call of create_dms_directory() made by py and js.

Yes, hide the button when directory already exists can be an addition festure to add.

victoralmau:
> The PR fix problems due to recursive call of create_dms_directory() made by py and js.
> 
> Yes, hide the button when directory already exists can be an addition festure to add.

I still think this change just masks the problem; that is, why is the method being called multiple times? 
Calling the method shouldn't be allowed if a directory already exists (for example, hide the button that call that action). 
Is there a specific use case where this happens? That's exactly what needs to be fixed.

PR #502, opened 2026-08-06
[18.0][FIX] dms: Always b64encode contents from messages

houzefa-abba:
Previous implementation was failing to convert strings to base64, which is the case when we receive plain HTML encoded as base64; as shown in mail02.eml.

This in turn lead to inconsistencies when saving/reading these contents afterwards, as the rest of the code always assumes dms.file::content contains base64 data.

This was falling through cracks because although the resulting data is invalid base64, the b64decode impl in python 3.12/3.13 is lax enough it would ignore failures in this case because our test string is "base64-ish" enough.

However, when running Odoo on Debian with the patch https://sources.debian.org/patches/python3.13/3.13.5-2+deb13u3/CVE-2026-3446.patch/ applied, this test on mail02.eml produced errors in dms.file::_inverse_content when trying to b64decode:

> <class 'binascii.Error'> Invalid base64-encoded string: number of data characters (429) cannot be 1 more than a multiple of 4

### DIFF
--- a/dms/models/directory.py
+++ b/dms/models/directory.py
@@ -646,10 +646,9 @@ class DmsDirectory(models.Model):
                 "directory_id": self.id,
                 "name": uname,
             }
-            try:
-                vals["content"] = base64.b64encode(attachment.content)
-            except Exception:
-                vals["content"] = attachment.content
+            if isinstance(contents_raw := attachment.content, str):
+                contents_raw = contents_raw.encode()
+            vals["content"] = base64.b64encode(contents_raw)
             self.env["dms.file"].sudo().create(vals)
             names.append(uname)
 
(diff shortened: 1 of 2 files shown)


victoralmau (approved):
IMO, the change makes sense.

pedrobaeza (changes requested):
Please remove the change in the test, as it's arbitrary and the test is not adding value.

houzefa-abba:
> Please remove the change in the test, as it's arbitrary and the test is not adding value.

Hmm I added that test bit to show what this PR fixes: the test ensures b64 contents got saved in DB, not raw contents.

I agree the hard-coded 876/657 are ugly, I just pushed a simpler check that tests `base64.b64decode(db_contents) != db_contents`.

pedrobaeza (approved):
/ocabot merge patch

PR #503, opened 2026-08-10
[18.0][FIX] dms: Remove public read access to trigger redirect

christian-ramos-tecnativa:
In order for the access-token to be processed and redirect to the proper url my/dms/directory... the user access must fail first here https://github.com/OCA/OCB/blob/18.0/addons/portal/controllers/mail.py#L155 otherwise you will be redirected to the login page even with the access token available, to test this:

1. Create a folder
2. Press share action
3. Try to enter the url without login
4. You will get redirected to the login, with this patch the folder content its properly displayed
@Tecnativa TT63966
@pedrobaeza @victoralmau

### DIFF
--- a/dms/security/ir.model.access.csv
+++ b/dms/security/ir.model.access.csv
@@ -8,12 +8,10 @@ access_dms_storage_portal,dms_storage_portal,model_dms_storage,base.group_portal
 access_dms_storage_user,dms_storage_user,model_dms_storage,group_dms_user,1,0,0,0
 access_dms_storage_manager,dms_storage_manager,model_dms_storage,group_dms_manager,1,1,1,1
 
-access_dms_directory_public,dms_directory_public,model_dms_directory,base.group_public,1,0,0,0
 access_dms_directory_portal,dms_directory_portal,model_dms_directory,base.group_portal,1,0,0,0
 access_dms_directory_base_user,dms_directory_base_user,model_dms_directory,base.group_user,1,0,0,0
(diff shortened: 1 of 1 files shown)


dms/security/ir.model.access.csv
pedrobaeza: You have to do the same here, isn't it?

pedrobaeza:
This was changed in 4595cc837c950ef223cb2b83e7ead84ea0c8e4e4, so @victoralmau can tell us if there's a reason for that.

dms/security/ir.model.access.csv
christian-ramos-tecnativa: It looks like the file is downloaded directly from the link without any issue

pedrobaeza: Then why the difference in the treatment? I think they should be homogenized. This permissions seems also excessive, but I think this is handled later for restricting more.

christian-ramos-tecnativa: It seems lo be relaetd to the directory needing to render a portal view. I tested it and the file download works even without the access right and this way remains more homogenized as you said

pedrobaeza (approved):

victoralmau (approved):
The change seems coherent.

For folders and files, there are already ACLs named `base.group_portal` and `base.group_user`.

The important thing is that the URL for accessing a folder works correctly and that a logged-in user can view the appropriate content from /my.

Functionality has been tested.

PR #496, opened 2026-07-20
[18.0][FIX] dms_field: Define the appropriate domain by filtering by res_model

victoralmau:
Define the appropriate domain by filtering by res_model

Related to https://github.com/OCA/dms/issues/494 (Bug 2)

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa

PR #504, opened 2026-08-11
[18.0][FIX] dms_field: Call the sanitizeDMSModel() method to obtain the appropriate model

victoralmau:
Call the `sanitizeDMSModel()` method to obtain the appropriate model

Use case example: `hr_dms_field`

Related to https://github.com/OCA/dms/pull/496

@Tecnativa TT64085

### DIFF
--- a/dms_field/static/src/views/dms_list/dms_list_controller.esm.js
+++ b/dms_field/static/src/views/dms_list/dms_list_controller.esm.js
@@ -442,7 +442,7 @@ export function getDMSListControllerObject() {
                     result,
                     new Domain([
                         ["res_id", "=", this.model.root.resId],
-                        ["res_model", "=", this.resModel],
+                        ["res_model", "=", this.sanitizeDMSModel(this.resModel)],
                     ]),
                 ]);
             } else {


pedrobaeza:
What `sanitizeDMSModel` does? Can you explain the problem? Which value returns before and after, for example.

victoralmau:
> What `sanitizeDMSModel` does? Can you explain the problem? Which value returns before and after, for example.

Returns the “correct” model, which is used in `hr_dms_field` https://github.com/OCA/dms/blob/6da958550a511dfa382c1f4208d6024adb4a5d95/hr_dms_field/static/src/js/base/ dms_list_controller.esm.js#L12 to return `hr.employee` when a user with lower permissions is accessing their own employee (`hr.employee.public`).

The `sanitizeDMSModel()` method is already used elsewhere in the code.

CarlosRoca13 (approved):
It makes sense
 👍 Thanks

pedrobaeza (approved):
Thanks for the explanation. Please put that info in the commit and PR comment next time from the beginning for being able to judge correctly and for future references.

For this one, let's not block it.

/ocabot merge patch

